PR #8, opened 2016-01-25
Changes necessary to fix signed builds

madeleinefletcher:
Made some changes that allow BinScope and Policheck to be run.
Fixed problem that caused perf collector NuGet package to be unsigned.


Src/PerformanceCollector/NuGet/PerformanceCollector.NuGet.csproj
```diff
@@ -2,6 +2,7 @@
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildProjectDirectory), 'Global.props'))\Global.props" />
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildProjectDirectory), 'Signing.props'))\Signing.props" />
```
SergeyKanzhelev: curious why do we need it for NuGet project. It only produces nupkg that is not signed

SergeyKanzhelev:
:shipit:

Src/PerformanceCollector/NuGet/PerformanceCollector.NuGet.csproj
madeleinefletcher: I tested it without and it worked, so I removed that.

PR #10, opened 2016-01-25
Adding gitlink integration so we don't have to publish to symweb

madeleinefletcher:
If you're not familiar with gitlink:

https://github.com/GitTools/GitLink
https://oren.codes/2015/09/23/enabling-source-code-debugging-for-your-nuget-packages-with-gitlink/

### DIFF
--- a/Src/DependencyCollector/Net45.Tests/packages.config
+++ b/Src/DependencyCollector/Net45.Tests/packages.config
@@ -3,4 +3,5 @@
   <package id="Microsoft.ApplicationInsights" version="2.0.0-beta4" targetFramework="net451" />
   <package id="StyleCop.MSBuild" version="4.7.49.1" targetFramework="net451" developmentDependency="true" />
   <package id="xunit" version="1.9.2" targetFramework="net451" />
+  <package id="gitlink" version="2.2.0" targetFramework="net45" />
 </packages>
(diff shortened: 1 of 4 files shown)


Src/DependencyCollector/Net45.Tests/packages.config
```diff
@@ -3,4 +3,5 @@
   <package id="Microsoft.ApplicationInsights" version="2.0.0-beta4" targetFramework="net451" />
```
SergeyKanzhelev: Not critical, but I wonder why other packages are targeted to `net451`. Any ideas?

madeleinefletcher: Gitlink only has target of net45. We're not referencing it in code anyway, just using the .exe from the command line.

SergeyKanzhelev:
I don't quite understand how it works. It says gitlnk will update pbd files with the source information. Does it mean we need to carry pdb files with SDK now?

SergeyKanzhelev:
If not, maybe we should keep pushing symbols to symbolsource.org?

madeleinefletcher:
I've updated this PR so it works with the build definition and uploads symbols to //symweb

abaranch:
:shipit:

PR #19, opened 2016-02-03
Fixes for intermittent test failures

madeleinefletcher:


madeleinefletcher:
@abaranch @SergeyKanzhelev do you think these are acceptable changes? 

I caught the exception in IISExpress because sometimes it was throwing access denied exception when we try to kill the process. The box doesn't have any running IIS Express processes on it, so I'm not sure it justifies an actual failure when this happens. 

The other change was because one of the tests was intermittently failing with "Access time should be above 0" message. I think this might be because we switched to using SQLExpress and sometimes it takes less than 1 millisecond to connect.

abaranch:
:shipit:

Dmitry-Matveev:
Small comment: IISExpress processes might've been killed automatically when the parent process (test itself) exited at the end. So, the exception you saw might justify the actual failure during the test execution.

PR #20, opened 2016-02-03
Correct fix for time should be above 0 error

madeleinefletcher:
There is an oddity in the behavior of TimeSpan.FromMilliseconds, where if you pass a value less than 1, it creates a timespan of 0 length. I had to convert the value to ticks manually so we can use TimeSpan.FromTicks()


madeleinefletcher:
@abaranch

abaranch:
:shipit:

PR #21, opened 2016-02-05
Changed default transfer request filter to only filter requests

madeleinefletcher:
Fix for this bug: https://github.com/Microsoft/ApplicationInsights-server-dotnet/issues/16


Src/Web/Web.Shared.Net/Implementation/HttpContextExtensions.cs
```diff
@@ -37,5 +38,24 @@ public static HttpResponse GetResponse(this HttpContext context)
...
         }
+
+        public static RequestNotification TryGetCurrentNotification(this HttpContext context)
+        {
+            RequestNotification notification = default(RequestNotification);
+            try
+            {
+                notification = context.CurrentNotification;
+            }
+            catch (NullReferenceException exp)
```
abaranch: Why do we need this?

Src/Web/Web.Shared.Net/Implementation/HttpContextExtensions.cs
```diff
@@ -37,5 +38,24 @@ public static HttpResponse GetResponse(this HttpContext context)
...
+            {
+                WebEventSource.Log.HttpRequestNotAvailable(exp.Message, exp.StackTrace);
+            }
+            catch (PlatformNotSupportedException exp)
+            {
+                WebEventSource.Log.HttpRequestNotAvailable(exp.Message, exp.StackTrace);
```
abaranch: Do we really need stack in case of unsupported exception?..

Src/Web/Web.Shared.Net/FilterRequest.cs
```diff
@@ -1,14 +1,14 @@
 ﻿namespace Microsoft.ApplicationInsights.Web
```
abaranch: Public API classes are in Web (like Processors, TelemetryInitializers and Module), everything else is under Implementation. I suspect this class needs to be moved. Could you please also check other classes. E.g. RequestNotificationFilter

Src/Web/Web.Shared.Net/Implementation/HttpContextExtensions.cs
madeleinefletcher: Because the getter for CurrentNotification throws if NotificationContext is null, but it's an internal property so I can't check for it.

abaranch: Let's add a comment. It is not obvious because this exception is not documented and looks like we try to catch our own exception.

Src/Web/Web.Shared.Net.Tests/HandlerTelemetryProcessorTests.cs
```diff
@@ -103,5 +103,31 @@ public void ProcessReturnsIfSuccessfulRequestHandlerIsFiltered()
 
             Assert.AreEqual(0, spy.ReceivedCalls);
         }
+
+        [TestMethod]
```
abaranch: Looks like this class needs cleanup method that would set HttpContext.Current to null. And in general in other places we try to encapsulate HttpContext.Current (see ResolvePlatformContext) so that 1. in tests we do not need to care about singletons 2. if in future we want to pass context from somewhere else we do not need to search for it in all places...

madeleinefletcher: What about the processors? Where should I put the ResolvePlatformContext method?

SergeyKanzhelev:
I was thinking more about this problem yesterday and it feels more and more to me that this filtering should only apply to the request telemetry we are collecting. Thus it should be part of request collection module as it was before.

Current implementation using telemetry processors makes all this checks for every single telemetry item that may slow down execution when, for instance, we have many dependencies. 

Also this implementation will filter out all telemetry reported from customer's END callback. It also will filter out all telemetry we will decide to send from the processors. For instance if we want to send "aggregated" event that counts all requests - this aggregated event will be subject to the same handler validation and potentially will be filtered out.

What do you think?

madeleinefletcher:
I didn't think of putting it back in there, but that might be better actually at least for this case. Are you proposing putting all the handler checking there, or just for Transfer Request Handler? We should probably still allow removing the transfer handler from filters through configuration though, right?

SergeyKanzhelev:
yes, the list of handlers was initially implemented as a property of `WebRequestTrackingModule` and those requests were never even reported to `Track` method. It may be a good alternative to the issues we see using TelemetryProcessor for this

madeleinefletcher:
Ok, I've put the handlers back in web request tracking module

Src/Web/Web.Shared.Net/Implementation/RequestNotificationFilter.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// </summary>
+        public RequestNotificationFilter()
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RequestNotificationFilter"/> class.
+        /// </summary>
+        /// <param name="value">The <see cref="System.Web.RequestNotification"/> value to filter on.</param>
+        public RequestNotificationFilter(string value)
```
SergeyKanzhelev: Is it used anywhere?

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -23,7 +23,7 @@
     public class RequestTrackingTelemetryModule : ITelemetryModule, IDisposable
     {
         private readonly EventListener listener;
-
+        private readonly IList<FilterRequest> handlersToFilter = new List<FilterRequest>();
```
SergeyKanzhelev: why `FilterRequest`, not string?

SergeyKanzhelev:
:shipit: with two comments that would be nice to address.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
madeleinefletcher: I couldn't get it to work with the XML config. Do you know how to use a list of strings for XML config?

Src/Web/Web.Shared.Net/Implementation/RequestNotificationFilter.cs
madeleinefletcher: Nope, I forgot to delete it.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
SergeyKanzhelev: As far as I remember you just write `<Add>name</Add>`

SergeyKanzhelev: I had it there: https://mseng.visualstudio.com/DefaultCollection/AppInsights/_git/AppDataCollection/pullrequest/59962?view=compare&path=%2FWeb%2FWeb%2FWeb.Shared.Net%2FRequestTrackingTelemetryModule.cs and I tested it for sure - let me find out

SergeyKanzhelev: removed all test apps. I'm pretty sure Add would make a magic

madeleinefletcher: Changed to list string. Thanks for the tip!

PR #34, opened 2016-02-23
Update eventsources to not use 4.6 specific signatures

abaranch:


SergeyKanzhelev:
Nice. Can we also wrap `Initialize` in try catch or it will be separately?

abaranch:
Try/catch is in a different repo: https://github.com/Microsoft/ApplicationInsights-dotnet/pull/197

SergeyKanzhelev:
:shipit:

Issue #32, opened 2016-02-23
Implement QuickPulse feature that enables real-time monitoring of critical application data.


tokaplan:
QuickPulse (aka ICU) feature enables real-time monitoring of critical application data. SDK will ping the QuickPulse service until it responds with a request to collect, at which point real-time data starts being collected and sent out to the service every second. As soon as the data is not being viewed any more, the service will signal SDK to stop collection and return to being idle.

PR #35, opened 2016-02-23
Implementation of QuickPulse (aka ICU) functionality.

tokaplan:
This is in relation to the issue #32.


Src/PerformanceCollector/Net40/Perf.Net40.csproj
```diff
@@ -44,12 +44,16 @@
...
     <Reference Include="System.Net" />
+    <Reference Include="System.Net.Http" />
     <Reference Include="System.Runtime">
       <HintPath>..\..\..\..\packages\Microsoft.Bcl.1.1.8\lib\net40\System.Runtime.dll</HintPath>
     </Reference>
+    <Reference Include="System.Runtime.Serialization" />
     <Reference Include="System.Threading.Tasks">
       <HintPath>..\..\..\..\packages\Microsoft.Bcl.1.1.8\lib\net40\System.Threading.Tasks.dll</HintPath>
     </Reference>
+    <Reference Include="System.Web" />
```
SergeyKanzhelev: this is not very good to reference Web from performance counters collection. It's heavy and we want to allow to use performance counters collector for worker roles without any `System.Web` references

Src/Microsoft.ApplicationInsights.Web.sln.DotSettings
```diff
@@ -0,0 +1,2 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
```
SergeyKanzhelev: Don’t check in DotSettings. Let's discuss if you really need it

Src/PerformanceCollector/Shared/Implementation/QuickPulse/IQuickPulseServiceClient.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// </summary>
+        Uri ServiceUri { get; }
+
+        /// <summary>
+        /// Pings QPS to check if it expects data right now.
+        /// </summary>
+        /// <param name="instrumentationKey">InstrumentationKey for which to submit data samples.</param>
+        /// <param name="timestamp">Timestamp to pass to the server.</param>
+        /// <returns><b>true</b> if data is expected, otherwise <b>false</b>.</returns>
+        bool? Ping(string instrumentationKey, DateTime timestamp);
```
SergeyKanzhelev: Use `DateTimeOffset`

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
```diff
@@ -0,0 +1,215 @@
...
+
+        private readonly Action onStartCollection;
+
+        private readonly Action onStopCollection;
+
+        private readonly Func<IList<QuickPulseDataSample>> onSubmitSamples;
+
+        private readonly Action<IList<QuickPulseDataSample>> onReturnFailedSamples;
+
+        private DateTime lastSuccessfulPing;
```
SergeyKanzhelev: again `DateTimeOffset` works better here

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseTimeProvider.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.Implementation.QuickPulse
+{
+    using System;
+
+    internal class QuickPulseTimeProvider
```
SergeyKanzhelev: Can you pls call it IClock for consistency with https://github.com/Microsoft/ApplicationInsights-dotnet/blob/271edd5fd8b9c7d84a9bd72448ec14f9b2145259/src/Core/Managed/Shared/Extensibility/Implementation/IClock.cs#L5

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
```diff
@@ -0,0 +1,468 @@
...
+
+        private readonly TimeSpan catastrophicFailureTimeout = TimeSpan.FromSeconds(5);
+
+        private readonly TimeSpan initialDelay = TimeSpan.Zero;
+
+        private readonly object lockObject = new object();
+
+        private readonly object collectedSamplesLock = new object();
+        
+        private readonly Uri serviceUriDefault = new Uri("https://qps.com/api");
```
SergeyKanzhelev: Don't use third-party names for temporary constants

Src/PerformanceCollector/Shared/Implementation/PerformanceCounterUtility.cs
```diff
@@ -97,7 +97,7 @@ public static string FormatPerformanceCounter(string categoryName, string counte
                         CultureInfo.InvariantCulture,
                         @"Invalid performance counter name format: {0}. Expected formats are \category(instance)\counter or \category\counter",
                         performanceCounter),
-                    "performanceCounter");
+                    nameof(performanceCounter));
```
SergeyKanzhelev: seriously, this can go to the separate PR. It's 49 files in PR already

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
```diff
@@ -0,0 +1,468 @@
...
+            this.dataAccumulatorManager.CompleteCurrentDataAccumulator();
+            this.telemetryProcessor.StartCollection(this.dataAccumulatorManager);
+
+            this.ScheduleNextCollection();
+        }
+
+        private void ScheduleNextCollection()
+        {
+            DateTime nextTick = this.collectionTimeSlotManager.GetNextCollectionTimeSlot(this.timeProvider.UtcNow);
+            this.collectionTimer.ScheduleNextTick((nextTick - this.timeProvider.UtcNow).Duration());
```
SergeyKanzhelev: this may throw if `GetNextCollectionTimeSlot` was called at 499 milliseconds and this line will be executed at 500 milliseconds.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
```diff
@@ -0,0 +1,215 @@
...
+                        // the service wants us to keep collecting
+                        break;
+
+                    case false:
+                        // the service wants us to stop collection
+                        this.onStopCollection();
+                        break;
+                }
+
+                this.lastSuccessfulSubmit = keepCollecting.HasValue ? this.timeProvider.UtcNow : this.lastSuccessfulSubmit;
```
SergeyKanzhelev: You are taking `this.timeProvider.UtcNow` too often. I think good design would be to take it only twice - once before the call and once after the call. And use those values everywhere for calculation of intervals, passing to the server and so forth. This may eliminate some errors related to precision and accuracy of `DateTime`

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
```diff
@@ -0,0 +1,228 @@
...
+        }
+
+        public Uri ServiceUri { get; }
+
+        public bool? Ping(string instrumentationKey, DateTime timestamp)
+        {
+            MemoryStream bodyStream = this.WritePingData(timestamp);
+
+            var path = string.Format(CultureInfo.InvariantCulture, "ping?ikey={0}", instrumentationKey);
+            HttpWebResponse response = this.SendRequest(WebRequestMethods.Http.Post, path, bodyStream, true);
```
SergeyKanzhelev: I wonder why it is not a `GET` with timestamp in queryString? Why do you need timestamp at all?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
```diff
@@ -0,0 +1,228 @@
...
+            serializer.WriteObject(bodyStream, monitoringPoints.ToArray());
+
+            bodyStream.Position = 0;
+
+            return bodyStream;
+        }
+        
+        private HttpWebResponse SendRequest(string httpVerb, string path, MemoryStream body, bool enableRetry)
+        {
+            var requestUri = string.Format(CultureInfo.InvariantCulture, "{0}/{1}", this.ServiceUri.AbsoluteUri.TrimEnd('/'), path.TrimStart('/'));
```
SergeyKanzhelev: You may consider using `UriBuilder`. At least have AbsoluteUri trimmed when property is set, not on every send

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
```diff
@@ -0,0 +1,228 @@
...
+
+                    var response = request.GetResponse() as HttpWebResponse;
+                    if (response != null)
+                    {
+                        return response;
+                    }
+                }
+                catch (Exception e)
+                {
+                    QuickPulseEventSource.Log.ServiceCommunicationFailedEvent(e.ToString());
```
SergeyKanzhelev: Use extension method `e.ToInveriantString()`. I believe it's public in Core

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
```diff
@@ -0,0 +1,228 @@
...
+                                        InstrumentationKey = instrumentationKey,
+                                        Instance = this.instanceName,
+                                        Timestamp = sample.EndTimestamp,
+                                        Metrics = metricPoints.ToArray()
+                                    };
+
+                monitoringPoints.Add(dataPoint);
+            }
+
+            var serializer = new DataContractJsonSerializer(typeof(MonitoringDataPoint[]));
```
SergeyKanzhelev: will it be faster to construct it only once? I wonder whether it caches fields that needs to be serialized

SergeyKanzhelev: Also check for security requirements. XmlSerializers often needs temp directory permissions and may not work under `Network Service` account when `%temp%="C:\Windows\Temp"`. Maybe it worth researching code generation here to write JSON manually

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
```diff
@@ -0,0 +1,228 @@
...
+            serializer.WriteObject(bodyStream, monitoringPoints.ToArray());
```
SergeyKanzhelev: If you want to optimize for success scenario you need to write directly to the output stream of http request. This will eliminate a need to allocate in-memory buffer and instead will write directly to http client. With retries you'll serialize three times, but as I said - it worth optimizing for success scenarios as you plan to do http requests very often

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
```diff
@@ -0,0 +1,228 @@
...
+            {
+                try
+                {
+                    var request = WebRequest.Create(requestUri) as HttpWebRequest;
+                    request.Method = httpVerb;
+                    request.Timeout = (int)this.timeout.TotalMilliseconds;
+
+                    if (body != null)
+                    {
+                        body.Position = 0;
```
SergeyKanzhelev: you doing it twice. Decide where you need to do it and where you don't

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
```diff
@@ -0,0 +1,228 @@
...
+
+                        var requestStream = request.GetRequestStream();
+                        body.CopyTo(requestStream);
+                    }
+
+                    var response = request.GetResponse() as HttpWebResponse;
```
SergeyKanzhelev: Are you using own `Thread` or keeping task from thread pool for the duration of http call? If tasks - consider using `async`.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Service contract/MetricPoint.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Microsoft.ManagementServices.RealTimeDataProcessing.QuickPulseService
+{
+    using System.Runtime.Serialization;
+
+    [DataContract]
+    internal struct MetricPoint
```
SergeyKanzhelev: Consider using MetricTelemetry instead. It will ensure consistence of how you collect telemetry

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
```diff
@@ -0,0 +1,228 @@
...
+            };
+
+            var serializer = new DataContractJsonSerializer(typeof(MonitoringDataPoint));
+            serializer.WriteObject(bodyStream, dataPoint);
+            bodyStream.Position = 0;
+
+            return bodyStream;
+        }
+
+        private MemoryStream WriteSamples(IEnumerable<QuickPulseDataSample> samples, string instrumentationKey)
```
SergeyKanzhelev: I wonder why you taking `instrumentationKey` as an argument and `instanceName` only populated once. Shouldn't you take if from telemetry items all the time?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
```diff
@@ -0,0 +1,228 @@
...
+        }
+
+        private MemoryStream WritePingData(DateTime timestamp)
+        {
+            var bodyStream = new MemoryStream();
+
+            var dataPoint = new MonitoringDataPoint
+            {
+                Version = this.version,
+                //InstrumentationKey = instrumentationKey,
```
SergeyKanzhelev: Why not setting it here?

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
```diff
@@ -0,0 +1,468 @@
...
+                        string.Format(
+                            CultureInfo.InvariantCulture,
+                            "Error initializing QuickPulse module. QPS endpoint is not a correct URI: '{0}'",
+                            this.QuickPulseServiceEndpoint),
+                        e);
+                }
+            }
+
+            // create the default production implementation of the service client with the best service endpoint we could get
+            this.serviceClient = new QuickPulseServiceClient(serviceEndpointUri, GetInstanceName(configuration), SdkVersionUtils.GetAssemblyVersion());
```
SergeyKanzhelev: It may be too early to read instance name as customer may set it's own initializer for role instance name population

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseEventSource.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        public string ApplicationName { [NonEvent]get; [NonEvent]private set; }
+
+        #region Infra init - success
+
+        [Event(1, Level = EventLevel.Informational, Message = @"QuickPulse infrastructure is being initialized. {0}")]
+        public void ModuleIsBeingInitializedEvent(
+            string message,
+            string dummy = "dummy",
+            string applicationName = "dummy")
```
abaranch: You need only 1 dummy parameter. Could you please also add eventsource Microsoft.Diagnostics.Tracing.EventRegister package. It adds target that validates EventSource.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseEventSource.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        [Event(3, Level = EventLevel.Informational, Message = @"Performance counter {0} has been successfully registered with QuickPulse performance collector.")]
+        public void CounterRegisteredEvent(string counter, string applicationName = "dummy")
+        {
+            this.WriteEvent(3, counter, this.ApplicationName);
+        }
+        #endregion
+
+        #region Infra init - failure
+        [Event(5, Keywords = Keywords.UserActionable, Level = EventLevel.Warning, Message = @"Performance counter {1} has failed to register with QuickPulse performance collector. This might happen whenever an application is running on a platform that doesn't provide access to performance counters. Technical details: {0}")]
```
abaranch: Warning will not be sent to a customer. What's the point making it user actionable?

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
```diff
@@ -0,0 +1,468 @@
...
+            this.serviceClient = serviceClient;
+            this.performanceCollector = performanceCollector;
+            this.timings = timings;
+        }
+
+        /// <summary>
+        /// Gets the QuickPulse service endpoint to send to.
+        /// </summary>
+        /// <remarks>Loaded from configuration.</remarks>
+        public string QuickPulseServiceEndpoint { get; private set; }
```
SergeyKanzhelev: So changing of endpoint in runtime wouldn't affect where we send telemetry to?
Why is it `private set`? Don't you want to configure it from `ApplicationInsights.config` file?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulsePerfCounterList.cs
```diff
@@ -0,0 +1,40 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.Implementation.QuickPulse
+{
+    using System;
+
+    internal enum QuickPulsePerfCounters
```
abaranch: Could you please move it in a separate file?

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
```diff
@@ -0,0 +1,468 @@
...
+
+        private QuickPulseCollectionTimeSlotManager collectionTimeSlotManager = null;
+
+        private IQuickPulseDataAccumulatorManager dataAccumulatorManager = null;
+
+        private IQuickPulseTelemetryProcessor telemetryProcessor = null;
+
+        private QuickPulseCollectionStateManager stateManager = null;
+
+        private IPerformanceCollector performanceCollector = null;
```
SergeyKanzhelev: Allow to override the list of counters to collect in order for customer to remove asp.net-specific customers from worker role to minimize the number of exceptions thrown in her application

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
abaranch: Could you please cleanup unused code?

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
```diff
@@ -0,0 +1,468 @@
...
+        private void ValidateConfiguration(TelemetryConfiguration configuration)
+        {
+            if (configuration == null)
+            {
+                throw new ArgumentNullException(nameof(configuration));
+            }
+
+            if (configuration.TelemetryInitializers == null)
+            {
+                throw new ArgumentNullException(nameof(configuration.TelemetryInitializers));
```
SergeyKanzhelev: why?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
```diff
@@ -0,0 +1,228 @@
...
+            var bodyStream = new MemoryStream();
+            var monitoringPoints = new List<MonitoringDataPoint>();
+
+            foreach (var sample in samples)
+            {
+                var metricPoints = new List<MetricPoint>
+                                       {
+                                           new MetricPoint
+                                               {
+                                                   Name = @"\ApplicationInsights\Requests/Sec",
```
abaranch: Do we have these names only in one place? Does it make sence to move it to some consts file? (And same for QuickPulsePerfCounterList)

Src/PerformanceCollector/Shared/PerformanceCollectorModule.cs
```diff
@@ -447,12 +448,13 @@ private void ProcessCustomCounters()
...
             int i = 0;
             foreach (PerformanceCounterCollectionRequest req in this.Counters)
-                    {
-                        // TODO remove workaround
-                        req.ReportAs = string.IsNullOrWhiteSpace(req.ReportAs)
-                                           ? req.PerformanceCounter.Trim('\\').Replace(@"\", @" - ")
-                                           : req.ReportAs;
-                        req.ReportAs = this.SanitizeReportAs(req.ReportAs, req.PerformanceCounter, ref i);
+            {
+                // TODO remove workaround
```
abaranch: ?

Src/PerformanceCollector/Xdt.Tests/Xdt.Tests.csproj
```diff
@@ -95,11 +98,11 @@
...
   <Target Name="EnsureNuGetPackageBuildImports" BeforeTargets="PrepareForBuild">
     <PropertyGroup>
-      <ErrorText>This project references NuGet package(s) that are missing on this computer. Enable NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is {0}.</ErrorText>
+      <ErrorText>This project references NuGet package(s) that are missing on this computer. Use NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is {0}.</ErrorText>
     </PropertyGroup>
-    <Error Condition="!Exists('..\..\..\..\packages\StyleCop.MSBuild.4.7.49.0\build\StyleCop.MSBuild.Targets')" Text="$([System.String]::Format('$(ErrorText)', '..\..\..\..\packages\StyleCop.MSBuild.4.7.49.0\build\StyleCop.MSBuild.Targets'))" />
+    <Error Condition="!Exists('..\..\..\..\packages\StyleCop.MSBuild.4.7.49.1\build\StyleCop.MSBuild.Targets')" Text="$([System.String]::Format('$(ErrorText)', '..\..\..\..\packages\StyleCop.MSBuild.4.7.49.1\build\StyleCop.MSBuild.Targets'))" />
```
abaranch: FYI Latest one is 50

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
```diff
@@ -0,0 +1,468 @@
...
+        {
+            this.collectionTimer = new Timer(this.CollectionTimerCallback);
+
+            this.stateTimer = new Timer(this.StateTimerCallback);
+            this.stateTimer.ScheduleNextTick(this.initialDelay);
+        }
+
+        private IQuickPulseTelemetryProcessor FetchTelemetryProcessor(TelemetryConfiguration configuration)
+        {
+            return configuration.TelemetryProcessors.OfType<IQuickPulseTelemetryProcessor>().SingleOrDefault();
```
SergeyKanzhelev: At this stage I'm questioning the need of telemetry module at all. Why wouldn't processor collect counters by itself? Ideally only one of those two should send data and know serviceUrl. This coordination of two things looks bad for me. And telemetry module is not an actual telemetry module as it doesn't produce any telemetry from SDK's perspective.

Alternatively change model that processor controls the module and make it start producing real SDK telemetry that will only be processed by this telemetry processor.

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
```diff
@@ -0,0 +1,468 @@
...
+            var currentCallbackStarted = this.timeProvider.UtcNow;
+            TimeSpan? timeToNextUpdate = null;
+
+            try
+            {
+                // the first instrumentation key that we get from TelemetryConfiguration.Active
+                // will be our permanent instrumentation key
+                if (string.IsNullOrWhiteSpace(this.instrumentationKey))
+                {
+                    this.instrumentationKey = TelemetryConfiguration.Active.InstrumentationKey;
```
SergeyKanzhelev: Use one that passed in `Initialize`. Also many people initialize instrumentation key in Global.asax after all modules already ininitalized

Src/PerformanceCollector/Net40/Perf.Net40.csproj
tokaplan: Fixed. This slipped in when I habitually used HttpUtility.UrlEncode.

Src/Microsoft.ApplicationInsights.Web.sln.DotSettings
tokaplan: Removed. I'm the only one who uses R#, so no, I don't need it :)

Src/PerformanceCollector/Shared/Implementation/QuickPulse/IQuickPulseServiceClient.cs
tokaplan: Done. Don't see the value, but it looks cooler :)

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
tokaplan: Done. Again, see no value, but looks cool.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseTimeProvider.cs
tokaplan: Done. I understand you're talking about renaming? I've renamed this to Clock.

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
tokaplan: Done. It does exist, doesn't it...

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
tokaplan: I have no reason to freeze those timestamps. Unless the call is associated with performance overhead (and it's not), I see no point in consciously reducing accuracy of the timestamp at any particular point in execution of this code.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
tokaplan: I don't really control this interface, I'm complying with a specification here. I think one of the reasons why this can't be a GET is because they want the information about the instance which is sending the Ping, and they wanted it in the same format as regular data for some reason (maybe it's just easier that way). Zaki would know for sure.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
tokaplan: There is no good way to concatenate URI's (as outlined in many online discussions, e.g. http://stackoverflow.com/questions/372865/path-combine-for-urls). Both Uri constructor and other ways (including UriBuilder) don't fully solve the problem, they still depend on whether there's a slash at the beginning of the relative path, and the end of the absolute one. Since I need to Trim() anyway, I'd rather form the URI manually.
I have changed one of the Trims to only happen once.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
tokaplan: Done here and elsewhere.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
tokaplan: Doesn't look like DataContractJsonSerializer is caching anything. Hence instantiating it every time.

I couldn't find anything that would point to potential security issues with this class (there are some reports of issues when reflection is disabled through CAS, but nothing else). Do you think testing it under the conditions that you stated above would be a sufficient verification?

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
tokaplan: Not sure what you mean, which part would throw and why?

SergeyKanzhelev: `ScheduleNextTick` may get a negative value as an argument if timing is right

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
SergeyKanzhelev: Oh, ok. THis link says so: http://stackoverflow.com/questions/7660439/safest-place-for-the-xmlserializer-to-save-temp-files

Are you sure contract serializer do not cache types? I'd be surprised if they don't have this optimization implemented

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
SergeyKanzhelev: What if datetime.now will return timeA at first line and timeB that is less than time A two lines down due to precision issue. There is no guarantee that it time always goes forward. I just want to make sure there are no hidden logical bombs here

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
tokaplan: Good point, thanks! Done.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
tokaplan: Good point, but not applicable anymore, eliminated the intermediary buffer. Thanks!

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
tokaplan: Neither. This is a completely synchronous, blocking call. By design.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Service contract/MetricPoint.cs
tokaplan: We think this is best efficiency-wise right now. We will consider this option along with many bandwidth-saving ones (like using bond or other kinds of binary telemetry) in future iterations.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
tokaplan: `instanceName` is a unique identifier of the machine, so it never changes. Quick Pulse service uses this value to identify how many servers are reporting data at any given moment. `instrumentationKey` is also always the same, actually, I will refactor for cleaner design a bit later.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
tokaplan: Right now the service doesn't expect the instrumentation key to be set in the body of Ping request, it takes it from the query parameter. This is likely to be adjusted in future iterations, but as long as this weirdness exists, I'd rather keep it here as a reminder of what's going on.

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
tokaplan: I'm thinking the same thing, but do we have a better option?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseEventSource.cs
tokaplan: How do I "add" the event source?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseEventSource.cs
tokaplan: I would like for it to be sent. Why won't it be sent?

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
tokaplan: `private set` won't be in a way of this property being set from `ApplicationInsights.config`. And yes, this is not supposed to be changing at runtime. But you're right, the setter should be public in case the module is instantiated from code. The user must be able to set the endpoint.
Done.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulsePerfCounterList.cs
tokaplan: Done.

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
tokaplan: This will be addressed in future iterations. Only a few exceptions would be thrown though, since as per our existing perf collection logic we stop trying after a few failed attempts.

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
tokaplan: This is supposed to be `TelemetryProcessors`. Good catch, thanks.

Done.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
tokaplan: These are in one place only. Looks kind of ugly, but I really see no value in moving them out, they have to be somewhere.

Src/PerformanceCollector/Shared/PerformanceCollectorModule.cs
tokaplan: There was an incorrect indentation, this is how github shows the difference after fixing it.

Src/PerformanceCollector/Xdt.Tests/Xdt.Tests.csproj
tokaplan: Yup, I tried that, saw no support for C# 6, and reverted it back. Will update later if there's tangible value.

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
tokaplan: Let's take this offline.

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
tokaplan: Let's take this one offline as well.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
```diff
@@ -0,0 +1,215 @@
...
+            {
+                try
+                {
+                    var request = WebRequest.Create(requestUri) as HttpWebRequest;
+                    request.Method = httpVerb;
+                    request.Timeout = (int)this.timeout.TotalMilliseconds;
+
+                    onWriteBody?.Invoke(request.GetRequestStream());
+
+                    var response = request.GetResponse() as HttpWebResponse;
```
SergeyKanzhelev: returning to the previous comment. You cannot say "by design" and lock a thread from a thread pool for whatever amount of time you need - you suppose to release the thread and do calls asynchronously. It affects performance otherwise.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseEventSource.cs
abaranch: Because we send only errors

Src/PerformanceCollector/Shared/PerformanceCollectorModule.cs
abaranch: What about TODO?..

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
tokaplan: Will investigate and see if I can take it once at the beginning.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
tokaplan: As per conversation offline - we will keep a close look at this and if it affects thread pool in any way. I will try to get it in for iteration 1 is time permits, otherwise (and if no issues are detected while testing in Prod services) I'll address it for later iteration.

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
tokaplan: Agreed offline that a module is still somewhat advantageous since it exists in a single instance across telemetry configurations.

Src/PerformanceCollector/Shared/PerformanceCollectorModule.cs
tokaplan: That's still TODO, unrelated to QuickPulse.

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
tokaplan: Done. The processor filters items and only passes those that match the key passed in configuration to the module.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseEventSource.cs
tokaplan: Marked as non user actionable whenever we don't want them to be sent, marked as Error whenever we do.

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
tokaplan: That's why there's Duration() to make sure it becomes positive.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
```diff
@@ -0,0 +1,222 @@
...
+                this.IsCollectingData = keepCollecting ?? this.IsCollectingData;
+            }
+            else
+            {
+                // we are currently idle and pinging the service waiting for it to ask us to start collecting data
+                bool? startCollection = this.serviceClient.Ping(instrumentationKey, this.timeProvider.UtcNow);
+
+                QuickPulseEventSource.Log.TroubleshootingMessageEvent(string.Format(CultureInfo.InvariantCulture, "Ping sent. Response: '{0}'", startCollection));
```
abaranch: Having formatting in verbose traces is not recommended. You add unnecessary noise. Could you please add seperate method, move string in the method argument?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseEventSource.cs
```diff
@@ -0,0 +1,134 @@
...
+        public string ApplicationName { [NonEvent]get; [NonEvent]private set; }
+
+        #region Infra init - success
+
+        [Event(1, Level = EventLevel.Informational, Message = @"QuickPulse infrastructure is being initialized. {0}")]
+        public void ModuleIsBeingInitializedEvent(
+            string message,
+            string applicationName = "dummy")
+        {
+            this.WriteEvent(1, message, this.ApplicationName);
```
abaranch: I'm not sure abot all versions but in some EventSource does not properly handles null values. It would be great if we add ?? string.Empty to all places where you use strings

tokaplan: Done.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
tokaplan: Done. The only one with formatting left is only called once at initialization time.

abaranch:
:shipit:

Issue #26, opened 2016-02-11
Configuring two httpmodules generated two identical requests


SergeyKanzhelev:
HttpModule should use a singleton to generate Begin and End ETW events so we will not send duplicates when misconfigured

abaranch:
We cannot reliably detect if you have identical modules defined in the config, or IIS creats several HttpApplication instances: http://erraticdev.blogspot.com/2011/01/how-to-correctly-use-ihttpmodule-to.html 
@SergeyKanzhelev Let's close this by design.

SergeyKanzhelev:
As discussed offline we may want to try to detect dups inside a single request.

PR #37, opened 2016-02-24
Fix module generates duplicate events if added twice in config

abaranch:
#26

### DIFF
--- a/Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
+++ b/Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
@@ -13,10 +13,6 @@
     /// </summary>
     public sealed class ApplicationInsightsHttpModule : IHttpModule
     {
-        private static int mutex;
-
-        private readonly WebEventsPublisher publisher;
-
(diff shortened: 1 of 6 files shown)


Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
```diff
@@ -28,6 +28,13 @@ public ApplicationInsightsHttpModule()
             {
                 // The call initializes TelemetryConfiguration that will create and Intialize modules
                 TelemetryConfiguration configuration = TelemetryConfiguration.Active;
+
+                // Initialize publisher only in the first instance of the module
+                var result = Interlocked.Increment(ref mutex);
```
SergeyKanzhelev: ideally with this implementation you should be releasing instance you holding on `Dispose` and allow the next module to keep working.

I'm not sure how it will work when there are two modules and one of them for some reason will be "unloaded".

SergeyKanzhelev:
Do you want to take it in stable? For me this change is higher risk for RC->Stable than other pull requests from you

Issue #38, opened 2016-02-26
Consider making requests to QuickPulse service async.


tokaplan:
Currently all requests to QuickPulse service are blocking. Consider making requests to QuickPulse service async.

tokaplan:
This has been addressed by #42

PR #42, opened 2016-03-01
QuickPulse - scheduling

tokaplan:
Moving to the best available scheduling model based on load testing.
During the investigation related to issue #38, it was determined that any kind of ThreadPool-dependent scheduling (be that a timer or an async/await loop) is inappropriate for QuickPulse scheduling. As the load increases, QuickPulse threads start competing for ThreadPool's resources with incoming requests, which results in unacceptable delays in data delivery. Since we only need to create 2 threads once and run them for the entirety of the life cycle of the application, ThreadPool is of little to no value to us anyway, so we might as well create the 2 threads ourselves. This eliminates the need for asynchronous code and has an added benefit of decoupling from ThreadPool which handles incoming requests.

### DIFF
--- a/Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
+++ b/Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
@@ -130,7 +130,7 @@
 
                 bool? keepCollecting = this.serviceClient.SubmitSamples(dataSamplesToSubmit, instrumentationKey);
 
-                QuickPulseEventSource.Log.TroubleshootingMessageEvent(string.Format(CultureInfo.InvariantCulture, "Sample submitted. Response: '{0}'", keepCollecting));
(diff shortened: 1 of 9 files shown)


abaranch:
2 threads is bad:
- that will increase context switching. With timer you will switch only when time is reached. What if you have 2 Cores machine and you load it with :

- Waste compute. Sleeping threads will be scheduled on cores that would be able to do smth useful instead. Synchronous calls as we discussed needs to be removed

You need to use async posts / awaits and use Task.Delay instead of Thread.Sleep. Task.Delay (https://msdn.microsoft.com/en-us/library/hh139096(v=vs.110).aspx) is a hint for a state machine to schedule continuation task after the interval, there will be no waste resources as with Sleep.

abaranch:
Please read this: http://blogs.msmvps.com/peterritchie/2007/04/26/thread-sleep-is-a-sign-of-a-poorly-designed-program/

tokaplan:
Synced up offline, introducing the change discussed.

SergeyKanzhelev:
:Shipit:

PR #43, opened 2016-03-02
Prevent double requests collection

abaranch:
#37


SergeyKanzhelev:
:shipit: Please take a look at perf run after this. Wonder will it affect in perf any way

Issue #44, opened 2016-03-03
CLR deadlock during performance collection


abaranch:

tokaplan:
Mitigated  with PR #45

PR #45, opened 2016-03-04
Disabling perf collection under IIS Express

tokaplan:
Disabling perf collection under IIS Express. #44


Src/PerformanceCollector/Shared/PerformanceCollectorModule.cs
```diff
@@ -133,14 +145,13 @@ public void Initialize(TelemetryConfiguration configuration)
                         this.timer = new Timer(this.TimerCallback);
 
                         // schedule the first tick
-                        // we skip the first collection interval to allow process to load otherwise deadlock may occur 
-                        this.timer.ScheduleNextTick(this.collectionPeriod);
+                        this.timer.ScheduleNextTick(TimeSpan.Zero);
```
SergeyKanzhelev: can we change it after stable version? I don't want to risk this change now

Src/PerformanceCollector/Shared/PerformanceCollectorModule.cs
```diff
@@ -125,6 +131,12 @@ public void Initialize(TelemetryConfiguration configuration)
                             throw new ArgumentNullException("configuration");
                         }
 
+                        if (!this.EnableIISExpressPerformanceCounters && IsRunningUnderIisExpress())
+                        {
+                            PerformanceCollectorEventSource.Log.RunningUnderIisExpress();
```
SergeyKanzhelev: Maybe only report it in debug output and not send event? Otherwise we just came up with a new startup trace message for every developer using Application Insights.

If we really need this message in portal I'd at least roll a dice and only report it in 30% of times using Random function =)

SergeyKanzhelev: Debug output only should be ok since it's development experience anyway

tokaplan: What's the proper way to do "debug output" you're mentioning?

Src/PerformanceCollector/Shared/PerformanceCollectorModule.cs
tokaplan: Done.

Src/PerformanceCollector/Shared/PerformanceCollectorModule.cs
abaranch: Why would we write that at all? It is not an error, it is our by default behavior. We do not want customers to enable it back and get a deadlock. Let's report a verbose trace for diagnostics purposes only. If we want to write anything in debug output that should be "Performance counters collection is not available for issexpess" and that's it.

Src/PerformanceCollector/Shared/PerformanceCollectorModule.cs
```diff
@@ -176,6 +187,21 @@ private static ITelemetry CreateTelemetry(PerformanceCounter pc, string reportAs
...
+            var iisExpressProcessName = "iisexpress";
+
+            try
+            {
+                return Process.GetCurrentProcess().ProcessName.IndexOf(iisExpressProcessName, StringComparison.OrdinalIgnoreCase) >= 0;
+            }
+            catch (Exception)
+            {
+                // we are unable to determine if we're running under IIS Express, assume we are not
+                return false;
```
abaranch: Can we write a trace here?

abaranch:
Does this change affect functional tests?

Src/PerformanceCollector/Shared/PerformanceCollectorModule.cs
tokaplan: Agreed, verbose trace here does sound like the best option.
Done.

Src/PerformanceCollector/Shared/PerformanceCollectorModule.cs
tokaplan: Done.

tokaplan:
Good catch, added the new flag to functional tests and verified that they were failing and are now passing.

abaranch:
:shipit:

Src/PerformanceCollector/Shared/Implementation/PerformanceCollectorEventSource.cs
```diff
@@ -88,6 +88,13 @@ public void CounterReportAsStrippedEvent(string counter, string stripped, string
             this.WriteEvent(8, misconfiguredCountersCount, overallConfiguredCountersCount, e, this.ApplicationName);
         }
 
+        [Event(15, Level = EventLevel.Verbose,
+            Message = @"Performance counters are unavailable when the application is running under IIS Express. Use EnableIISExpressPerformanceCounters element with a value of 'true' within the Performance Collector Module element to override this behavior.")]
+        public void RunningUnderIisExpress(string applicationName = "dummy")
```
SergeyKanzhelev: this event source is the only event source that uses `"dummy"` instead of `"Incorrect"`. I wonder why are trying to be special =)

Src/PerformanceCollector/Shared/PerformanceCollectorModule.cs
SergeyKanzhelev: I think @SoubhagyaDash meant to notify customer in debug output. Like `Debug.Write("Counters disabled on IISexpress")`. I'm fine with Verbose

Src/PerformanceCollector/Shared/Implementation/PerformanceCollectorEventSource.cs
tokaplan: I would have never dreamt of having a "dummy" or "incorrect" variable, so I think I copied it from somewhere in SDK :)

PR #50, opened 2016-03-07
Hardcode file version so it is bigger than all we shipped before

abaranch:


SergeyKanzhelev:
:shipit: merge with develop will take some effort after this, but fine

PR #55, opened 2016-03-11
SDK sends transmission timestamp along with the sample batch.

tokaplan:
SDK sends transmission timestamp along with the sample batch. This is a prerequisite for implementing a better clock synchronization method on the server side.

### DIFF
--- a/Src/PerformanceCollector/Shared/Implementation/QuickPulse/IQuickPulseServiceClient.cs
+++ b/Src/PerformanceCollector/Shared/Implementation/QuickPulse/IQuickPulseServiceClient.cs
@@ -24,7 +24,8 @@
         /// </summary>
         /// <param name="samples">Data samples.</param>
         /// <param name="instrumentationKey">InstrumentationKey for which to submit data samples.</param>
+        /// <param name="timerProvider">Time provider.</param>
         /// <returns><b>true</b> if the client is expected to keep sending data samples, <b>false</b> otherwise.</returns>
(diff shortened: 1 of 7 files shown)


tokaplan:
This is a bit heavyweight given that 99% of all batches would contain a single sample anyway, and restrictive going forward, since we would have to group samples by instrumentation key. We'll probably just include the transmission timestamp into each sample.

tokaplan:
Overriden by #56

Src/PerformanceCollector/Shared/Implementation/QuickPulse/IQuickPulseServiceClient.cs
```diff
@@ -24,7 +24,8 @@ internal interface IQuickPulseServiceClient
         /// </summary>
         /// <param name="samples">Data samples.</param>
         /// <param name="instrumentationKey">InstrumentationKey for which to submit data samples.</param>
+        /// <param name="timerProvider">Time provider.</param>
         /// <returns><b>true</b> if the client is expected to keep sending data samples, <b>false</b> otherwise.</returns>
-        bool? SubmitSamples(IEnumerable<QuickPulseDataSample> samples, string instrumentationKey);
+        bool? SubmitSamples(IEnumerable<QuickPulseDataSample> samples, string instrumentationKey, Clock timerProvider);
```
SergeyKanzhelev: should it be `IClock`?

PR #60, opened 2016-03-18
Working around TelemetryProcessorChain being recreated on Build()

tokaplan:
Whenever a customer adds his own TelemetryProcessor from code, SDK will re-instantiate all TelemetryProcessors, leaving QuickPulseTelemetryProcessor hanging without any telemetry items going through it. To remedy the issue, QuickPulseTelemetryProcessor now registers itself with QuickPulseTelemetryModule, not the other way around, so that when the processor is recreated, it will either automatically find the module and register with it (ApplicationInsights.config scenario), or the customer will manually call module.RegisterTelemetryProcessor() if the module is created in code.

### DIFF
--- a/Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseEventSource.cs
+++ b/Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseEventSource.cs
@@ -58,12 +58,6 @@
         {
             this.WriteEvent(6, e ?? string.Empty, counter ?? string.Empty, this.ApplicationName);
         }
-
-        [Event(7, Keywords = Keywords.UserActionable, Level = EventLevel.Error, Message = @"QuickPulseTelemetryProcessor could not locate a QuickPulseTelemetryModule in configuration. QuickPulse data will not be available. Make sure both QuickPulseTelemetryProcessor and QuickPulseTelemetryModule are in ApplicationInsights.config or otherwise present.")]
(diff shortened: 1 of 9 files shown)


Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseEventSource.cs
```diff
@@ -138,6 +132,12 @@ public void SampleStoredEvent(int bufferLength, string applicationName = "dummy"
             this.WriteEvent(19, bufferLength, this.ApplicationName);
         }
 
+        [Event(7, Level = EventLevel.Verbose, Message = @"QuickPulseTelemetryModule has received a registration request from a QuickPulseTelemetryProcessor.")]
+        public void ProcessorRegistered(int count, string applicationName = "dummy")
+        {
+            this.WriteEvent(7, count, this.ApplicationName);
```
abaranch: Please change to int, sting, string so we do not have this issue again: https://github.com/Microsoft/ApplicationInsights-dotnet/issues/194

abaranch:
:shipit:

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseEventSource.cs
tokaplan: Done, thank you!

PR #64, opened 2016-03-28
Bug fix: QuickPulseTelemetryProcessor leaks QuickPulse dependencies before collection starts.

tokaplan:
Bug fix: QuickPulseTelemetryProcessor leaks QuickPulse dependencies before collection starts


abaranch:
:shipit:

Issue #67, opened 2016-04-07
On .NET 4.5.2 successful HTTP dependencies are logged as failing


AlexBulankou:
Try this example:
```
 static void Main(string[] args)
        {
            new DependencyTrackingTelemetryModule().Initialize(TelemetryConfiguration.Active);
            Test().Wait();
        }

        static async Task Test()
        {
            var result = await new HttpClient().GetAsync("https://www.bing.com/search?q=httpclient");
            var response = await WebRequest.Create("https://www.bing.com/search?q=webrequest").GetResponseAsync();                 
        }
```

These are dependencies that are getting logged on .NET 4.5.2. Note that we get nothing logged on 4.5 and 4.5.1 (probably expected) and on 4.6 everything is logged as expected

```
{"name":"Microsoft.ApplicationInsights.RemoteDependency","time":"2016-04-07T19:0
3:47.1438164Z","tags":{"ai.internal.sdkVersion":"rddf: 2.0.0.25000"},"data":{"ba
seType":"RemoteDependencyData","baseData":{"ver":2,"name":"https://www.bing.com/
search?q=httpclient","id":"R1YUAcsnqk8=","value":167.5677,"dependencyKind":1,"su
ccess":false}}}

{"name":"Microsoft.ApplicationInsights.RemoteDependency","time":"2016-04-07T19:0
3:47.7255489Z","tags":{"ai.internal.sdkVersion":"rddf: 2.0.0.25000"},"data":{"ba
seType":"RemoteDependencyData","baseData":{"ver":2,"name":"https://www.bing.com/
search?q=webrequest","id":"uGxqsnv1HTU=","value":487.8151,"dependencyKind":1,"su
ccess":false}}}
```

AlexBulankou:
From the customer report, looks like this is a regression from 1.2.3

AlexBulankou:
Verified that with 1.2.3 we are not reporting success flag at all:

```
{"name":"Microsoft.ApplicationInsights.RemoteDependency","time":"2016-04-07T20:4
3:33.3314554+00:00","tags":{"ai.internal.sdkVersion":"1.2.3.490"},"data":{"baseT
ype":"RemoteDependencyData","baseData":{"ver":2,"name":"https://www.bing.com/sea
rch?q=httpclient","kind":1,"value":170.2793,"dependencyKind":1,"async":true,"dep
endencySource":0}}}

[...]

PR #72, opened 2016-04-09
Cut HTTP support in 4.5.2

abaranch:
Cutting support completely because not only we do not have success and status code we also see duplicates. This cannot be fixed on our side since it is framework implementation. 4.6 does not have this problem so for monitoring applications that are compiled as 4.5.2 there are 2 options
- Install StatusMonitor
- Run application on the server that has 4.6 installed

Partially resolving #67

### DIFF
--- a/Src/DependencyCollector/Shared/Implementation/FrameworkHttpEventListener.cs
+++ b/Src/DependencyCollector/Shared/Implementation/FrameworkHttpEventListener.cs
@@ -142,15 +142,13 @@ namespace Microsoft.ApplicationInsights.DependencyCollector.Implementation
                 // .NET 4.6 onwards will be passing the following additional params.
                 if (eventData.Payload.Count >= 4)
                 {
+                    // Payload[1] = success
+                    // Payload[2] = synchronous
+                    // Payload[3] = statusCode
                     if (eventData.Payload[1] != null)
                     {
                         success = Convert.ToBoolean(eventData.Payload[1], CultureInfo.InvariantCulture);
                     }
-
-                    ////if (eventData.Payload[2] != null)
-                    ////{
-                    ////    synchronous = Convert.ToBoolean(eventData.Payload[2], CultureInfo.InvariantCulture);
-                    ////}
 
                     if (eventData.Payload[3] != null)
                     {
--- a/Src/DependencyCollector/Shared/Implementation/FrameworkHttpEventListener.cs
+++ b/Src/DependencyCollector/Shared/Implementation/FrameworkHttpEventListener.cs
@@ -137,7 +137,6 @@ namespace Microsoft.ApplicationInsights.DependencyCollector.Implementation
                 long id = Convert.ToInt64(eventData.Payload[0], CultureInfo.InvariantCulture);
 
                 bool? success = null;
-                bool synchronous = false;
                 int? statusCode = null;
 
                 // .NET 4.6 onwards will be passing the following additional params.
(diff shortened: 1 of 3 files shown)


Src/DependencyCollector/Shared/Implementation/FrameworkHttpEventListener.cs
```diff
@@ -148,10 +147,10 @@ private void OnEndGetResponse(EventWrittenEventArgs eventData)
...
                     }
 
-                    if (eventData.Payload[2] != null)
-                    {
-                        synchronous = Convert.ToBoolean(eventData.Payload[2], CultureInfo.InvariantCulture);
-                    }
+                    ////if (eventData.Payload[2] != null)
+                    ////{
+                    ////    synchronous = Convert.ToBoolean(eventData.Payload[2], CultureInfo.InvariantCulture);
+                    ////}
```
dnduffy: Should just delete the commented out code since you can always get it back from history if need be.

Src/DependencyCollector/Shared/Implementation/FrameworkHttpProcessing.cs
```diff
@@ -108,19 +108,23 @@ public void OnEndHttpCallback(long id, bool? success, bool synchronous, int? sta
...
-                telemetry.ResultCode = statusCode.Value > 0 ? statusCode.Value.ToString(CultureInfo.InvariantCulture) : string.Empty;
-
-                // We calculate success on the base of http code and do not use the 'success' method argument
-                // because framework returns true all the time if you use HttpClient to create a request
-                // statusCode == -1 if there is no Response
-                telemetry.Success = (statusCode > 0) && (statusCode < 400);
+                    // We calculate success on the base of http code and do not use the 'success' method argument
+                    // because framework returns true all the time if you use HttpClient to create a request
+                    // statusCode == -1 if there is no Response
+                    telemetry.Success = (statusCode > 0) && (statusCode < 400);
```
dnduffy: What happens with 302?  A redirect is not a failure.

Src/DependencyCollector/Shared/Implementation/FrameworkHttpProcessing.cs
```diff
@@ -108,19 +108,23 @@ public void OnEndHttpCallback(long id, bool? success, bool synchronous, int? sta
...
 
-                ClientServerDependencyTracker.EndTracking(this.telemetryClient, telemetry);
+                    ClientServerDependencyTracker.EndTracking(this.telemetryClient, telemetry);
+                }
+                else
+                {
+                    // This case is for For 4.5.2
+                    // We never not collected statusCode or success
+                    // We also have duplicates if runtime is also 4.5.2 (4.6 runtime has no such problem)
+                    // So starting 2.1.0-beta4 we are cutting support of http dependencies for 4.5.2
```
dnduffy: What is meant by "never not"?

The last sentence would read better as "Starting with 2.1.0-beta4 we are cutting support for HTTP dependencies in .NET 4.5.2."

dnduffy:
:shipit:

Src/DependencyCollector/Shared/Implementation/FrameworkHttpProcessing.cs
abaranch: Yes, 302 would be success.

dnduffy: My apologies, I inverted the condition in my head when I read it. ;)

AlexBulankou:
Yes! cut it

hannesdecorte:
I installed StatusMonitor on the Azure Cloud Service and it's running OK now.
So cut it! ;-)
Thanks for the follow up.

PR #79, opened 2016-04-29
QuickPulse module is now sending stream id

tokaplan:
Currently, QuickPulse is only sending an instance name (which is either a role name or a machine name) with its samples. This leaves a gap when hosting the same application multiple times on the same machine (in different IIS Web Sites, possibly) - those streams will have the same instance name and will be identified as one by the server. Stream Id is solving this problem by introducing an additional identifier to differentiate between such streams (currently a GUID generated at start-up time).


abaranch:
:shipit: But please do not merge till we decide where it goes.

abaranch:
You can merge to develop

tokaplan:
Wrong branch, resubmitting.

PR #86, opened 2016-06-01
Change version from x.x.x.x to x.x.x-x

abaranch:


Src/DependencyCollector/Shared.Tests/Implementation/FrameworkSqlProcessingTest.cs
```diff
@@ -315,6 +185,14 @@ public void Dispose()
...
                 remoteDependencyTelemetryActual.Duration <= TimeSpan.FromMilliseconds(valueMax),
                 string.Format(CultureInfo.InvariantCulture, "Value (dependency duration = {0}) in the sent telemetry should not be significantly bigger than the time duration between start and end", remoteDependencyTelemetryActual.Duration));
+
+            string versonStr = Assembly.GetAssembly(typeof(DependencyTrackingTelemetryModuleTest)).GetCustomAttributes(false)
+                    .OfType<AssemblyFileVersionAttribute>()
+                    .First()
+                    .Version;
+            Version version = new Version(versonStr);
+            string expectedVersion = "rddf:" + version.ToString(3) + "-" + version.Revision;
+            Assert.AreEqual(expectedVersion, remoteDependencyTelemetryActual.Context.GetInternalContext().SdkVersion);
```
dnduffy: Would it make sense to put this block of version lookup code in a helper method that can be reused by the different tests?

Src/PerformanceCollector/Shared/Implementation/SdkVersionUtils.cs
```diff
@@ -1,16 +1,20 @@
 ﻿namespace Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.Implementation
 {
+    using System;
     using System.Linq;
     using System.Reflection;
 
     internal class SdkVersionUtils
     {
-        internal static string GetAssemblyVersion()
+        internal static string GetSdkVersion(string versionPrefix)
```
dnduffy: Are there multiple copies because they are in different projects?  Is there a clean way to share one copy?

dnduffy:
:shipit:

Src/DependencyCollector/Shared.Tests/Implementation/FrameworkSqlProcessingTest.cs
abaranch: Done

---
In reply to: [65463611](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/86#discussion_r65463611) [](ancestors = 65463611)

Src/PerformanceCollector/Shared/Implementation/SdkVersionUtils.cs
abaranch: Yes, we have different projects and code for each is in Shared project. Shared projects do not support referencing items by link. To resolve your concern I moved this to Web.45 project and not I reference it from all other projects (like rdd.40, rdd.45 etc). That is more work but now code is in one place.

---
In reply to: [65464131](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/86#discussion_r65464131) [](ancestors = 65464131)

dnduffy:
:shipit: x2 ;)

PR #91, opened 2016-06-08
Send data from funcs to Validate endpoint

abaranch:


Test/DependencyCollector/FunctionalTests/FuncTest/Helpers/HttpListenerObservable.cs
```diff
@@ -161,5 +168,24 @@ private static string Decompress(string content)
...
+            HttpClient client = new HttpClient();
+            var result = client.PostAsync(
+                "https://dc.services.visualstudio.com/v2/validate",
+                new ByteArrayContent(Encoding.UTF8.GetBytes(items))).GetAwaiter().GetResult();
+
+            if (result.StatusCode != HttpStatusCode.OK)
+            {
+                var response = result.Content.ReadAsStringAsync().GetAwaiter().GetResult();
+                Trace.WriteLine("ERROR! Backend Response: " + response);
+                this.FailureDetected = true;
```
dnduffy: Why set a property instead of just returning true or false from the function?

dnduffy:
:shipit:

Test/DependencyCollector/FunctionalTests/FuncTest/Helpers/HttpListenerObservable.cs
abaranch: Because we do validation in a different place when test is done.

---
In reply to: [66305457](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/91#discussion_r66305457) [](ancestors = 66305457)

Issue #33, opened 2016-02-23
Dependency collector code coverage is less than 70%


abaranch:

MS-TimothyMothra:
@cijothomas Need to validate if this is still the case considering new tests infra.

cijothomas:
This would still be the same. This is unit test coverage.  Codecov is not posting results, so need to enable it first and then we'll know hows the code coverage. i'll investigate why code cov is broken

PR #94, opened 2016-06-13
Fix for bug that caused empty result codes in rddp

madeleinefletcher:


madeleinefletcher:
@dnduffy @abaranch

Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
```diff
@@ -301,9 +301,16 @@ private void OnEnd(object exception, object thisObj, object returnValue)
                 {
                     this.TelemetryTable.Remove(thisObj);
                     DependencyTelemetry telemetry = telemetryTuple.Item1;
-                    
+
                     var responseObj = returnValue as HttpWebResponse;
 
+                    var webException = exception as WebException;
```
AlexBulankou: What if responseObj is already set at this point (after previous line)? Should we be overwriting it? Or perhaps only take it from WebException if not already set?

Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
```diff
@@ -301,9 +301,16 @@ private void OnEnd(object exception, object thisObj, object returnValue)
...
+
+                    if (webException != null)
+                    {
+                        responseObj = webException.Response as HttpWebResponse;
+                    }
```
AlexBulankou: We can also get Http status code from HttpException ([more info](https://msdn.microsoft.com/en-us/library/system.web.httpexception.gethttpcode(v=vs.110).aspx)). Should we attempt to do it in case we are dealing with HttpException and don't have response?

Src/DependencyCollector/Shared.Tests/Implementation/ProfilerHttpProcessingTest.cs
```diff
@@ -115,6 +115,27 @@ public void RddTestHttpProcessingProfilerOnExceptionForGetResponse()
         }
 
         /// <summary>
+        /// Validates HttpProcessingProfiler sends correct telemetry including response code on calling OnExceptionForGetResponse for WebException.
+        /// </summary>
+        [TestMethod]
+        [Description("Validates HttpProcessingProfiler sends correct telemetry including response code on calling OnExceptionForGetResponse for WebException.")]
+        [Owner("cithomas")]
```
dnduffy: Is "cithomas" the owner of this new test case?

Src/DependencyCollector/Shared.Tests/Implementation/ProfilerHttpProcessingTest.cs
```diff
@@ -115,6 +115,27 @@ public void RddTestHttpProcessingProfilerOnExceptionForGetResponse()
...
+        [TestCategory("CVT")]
+        public void RddTestHttpProcessingProfilerOnWebExceptionForGetResponse()
+        {
+            var request = WebRequest.Create(this.testUrl);
+            this.httpProcessingProfiler.OnBeginForGetResponse(request);
+            Thread.Sleep(this.sleepTimeMsecBetweenBeginAndEnd);
```
dnduffy: Why is this sleeping here?  Is "OnBeginForGetResponse" not a handler (usually methods named with 'On*' imply they are handlers for events)?
If you need to wait for something to have happened before proceeding in a test it is much safer (and generally faster) to watch for a condition to be met than to sleep for an arbitrary amount of time.  Sleeping in tests can cause spurious failures due to race conditions, particularly as build infrastructure changes.

Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
```diff
@@ -301,9 +301,16 @@ private void OnEnd(object exception, object thisObj, object returnValue)
                 {
                     this.TelemetryTable.Remove(thisObj);
                     DependencyTelemetry telemetry = telemetryTuple.Item1;
-                    
+
                     var responseObj = returnValue as HttpWebResponse;
 
+                    var webException = exception as WebException;
+
+                    if (webException != null)
```
abaranch: Should we do this only if responseObj is null?

madeleinefletcher:
Ok, I added logic to get status code from HTTPExceptions as well (although, this is apparently only available in NET45). I also made it only use exception to set status code if responseObj is null (I did check though, and OnEnd is never called with both exception and responseObj non-null).

Src/DependencyCollector/Shared.Tests/Implementation/ProfilerHttpProcessingTest.cs
madeleinefletcher: I have no idea why there's a sleep here, it's what all the other tests for this class are doing. Maybe @abaranch knows?

Src/DependencyCollector/Shared.Tests/Implementation/ProfilerHttpProcessingTest.cs
madeleinefletcher: Fixed.

Src/DependencyCollector/Shared.Tests/Implementation/ProfilerHttpProcessingTest.cs
abaranch: These are old tests. I do not know how they work. You try to remove and see where it fails.

---
In reply to: [66872458](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/94#discussion_r66872458) [](ancestors = 66872458)

Src/DependencyCollector/Shared.Tests/Implementation/ProfilerHttpProcessingTest.cs
abaranch: Just FYI, we do not use owners attribute. These are old tests and attributes were added by the previous owner team. We can safely remove it for all the old tests and do not add for the new ones. (I'm not asking to do that now, just commenting for the future.)

---
In reply to: [66872558](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/94#discussion_r66872558) [](ancestors = 66872558)

abaranch:
:shipit:

Src/DependencyCollector/Shared.Tests/Implementation/ProfilerHttpProcessingTest.cs
dnduffy: I would try removing the sleep, see if it was even needed, and if it was needed then look for a status change to watch for instead of sleeping.  Based on the method names I would not expect sleeping here to be needed.

dnduffy:
:shipit:

abaranch:
When you merge you can close this enhancement: https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/33
:+1:

PR #97, opened 2016-06-22
Changed Status Monitor script to take a parameter for path

madeleinefletcher:
This is required to fix the build so it downloads Status Monitor from correct location


abaranch:
:shipit:

Issue #101, opened 2016-07-07
Sql Dependency error message


vladimirlapacek:
The Sql Dependency details are currently limited to a couple fields:

- success
- result code

Although the result code can help with troubleshooting issues, the full error message returned by the SQL server would be more useful, eg.

result code: 2601
vs.
error message: Cannot insert duplicate key row in object 'dbo.Products' with unique index 'UX_ProductSKU'. The duplicate key value is (Selenium SKU Min). The statement has been terminated.

Can the error message be added to the SQL dependency detail?

abaranch:
Message will be added when you have StatusMonitor on the box with the application. Feature will b released with 2.2-beta2

vladimirlapacek:
Hi Anastasia,

I really appreciate your (team) flexibility. It's really amazing :)

Thanks,
Vlad

abaranch:
Welcome :)

PR #115, opened 2016-08-05
Collect Sql error message (profiler only)

abaranch:
#101

### DIFF
--- a/Test/DependencyCollector/FunctionalTests/FuncTest/IIS/Iis.cs
+++ b/Test/DependencyCollector/FunctionalTests/FuncTest/IIS/Iis.cs
@@ -45,11 +45,10 @@ namespace FuncTest.IIS
         /// </summary>
         public static void Start()
         {
-            Process process = new Process { StartInfo = StartIis };
-            process.Start();
-            process.WaitForExit();
+            Stop();
 
-            process = new Process { StartInfo = StartW3Svc };
+            Process process = new Process { StartInfo = StartW3Svc };
+            process.StartInfo.WindowStyle = ProcessWindowStyle.Hidden;
             process.Start();
             process.WaitForExit();
         }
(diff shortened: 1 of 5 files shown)


Test/DependencyCollector/FunctionalTests/FuncTest/IIS/Iis.cs
```diff
@@ -45,11 +45,10 @@ public static void Reset()
         /// </summary>
         public static void Start()
         {
-            Process process = new Process { StartInfo = StartIis };
-            process.Start();
-            process.WaitForExit();
+            Stop();
```
dnduffy: ?

dnduffy:
Start calling Stop looks odd but if my assumption is correct then it is okay but should probably be commented and then :shipit:.

Test/DependencyCollector/FunctionalTests/FuncTest/IIS/Iis.cs
```diff
@@ -60,7 +59,9 @@ public static void Start()
         public static void Stop()
         {
             Process process = new Process { StartInfo = StopIis };
```
dnduffy: So StartIis was removed and only StartW3Svc remains in Start but Stop only has StopIis, should there be a StopW3Svc?

Test/DependencyCollector/FunctionalTests/FuncTest/IIS/Iis.cs
abaranch: Good catch. Thank you. I misread. I thought start is doing stop and than start of the same process.

---
In reply to: [73772634](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/115#discussion_r73772634) [](ancestors = 73772634)

PR #117, opened 2016-08-09
Fix the build (QuickPulse functional tests build break)

tokaplan:


dnduffy:
:shipit: 👍 
Interesting that it builds fine locally but not on the build machines...

PR #128, opened 2016-09-02
Removed rule CA2210.

dnduffy:
It seems the rule didn't get turned off in this one repository.


SergeyKanzhelev:
:shipit: whatever this rule is =)

dnduffy:
This is the rule that assemblies should be signed but because we delay sign them we need the rule turned off since the signing happens after the rules get run as part of the build.

PR #130, opened 2016-09-07
QuickPulse tweaks for Live Failures

tokaplan:


abaranch:
QuickPulseTelemetryModuleSupportsMultipleTelemetryProcessorsForMultipleConfigurations failed

Src/PerformanceCollector/Shared/QuickPulseTelemetryProcessor.cs
```diff
@@ -24,6 +24,12 @@ public class QuickPulseTelemetryProcessor : ITelemetryProcessor, ITelemetryModul
 
         private const int InitialTelemetryQuota = 3;
 
+        private const int MaxFieldLength = 32768;
```
abaranch: >MaxFieldLength [](start = 26, length = 14)

You may want to start thinking about moving sanitization logic in a separate class

Src/PerformanceCollector/Shared/QuickPulseTelemetryProcessor.cs
```diff
@@ -178,55 +184,96 @@ private static ITelemetryDocument ConvertRequestToTelemetryDocument(RequestTelem
...
+            if (telemetry.Properties != null && telemetry.Properties.Count > 0)
+            {
+                properties = new Dictionary<string, string>(MaxPropertyCount + 1);
+
+                foreach (var prop in
+                    telemetry.Properties
+                    .Where(p => !string.Equals(p.Key, specialPropertyName, StringComparison.Ordinal))
+                    .Take(MaxPropertyCount))
+                {
+                    properties.Add(TruncateValue(prop.Key), TruncateValue(prop.Value));
```
abaranch: >TruncateValue [](start = 35, length = 13)

Duplicate key exception

Src/PerformanceCollector/Shared/QuickPulseTelemetryProcessor.cs
```diff
@@ -178,55 +184,96 @@ private static ITelemetryDocument ConvertRequestToTelemetryDocument(RequestTelem
...
+                               exceptionTelemetry.Exception != null
+                                   ? TruncateValue(exceptionTelemetry.Exception.ToString())
+                                   : null,
+                           ExceptionType =
+                               exceptionTelemetry.Exception != null
+                                   ? TruncateValue(exceptionTelemetry.Exception.GetType().FullName)
+                                   : null,
+                           ExceptionMessage =
+                               exceptionTelemetry.Exception != null
+                                   ? TruncateValue(exceptionTelemetry.Exception.Message)
```
abaranch: >Message [](start = 80, length = 7)

exceptionTelemetry has Message property that by default is exception.Message but customer may override it to smth more meaningful

Src/PerformanceCollector/Shared/QuickPulseTelemetryProcessor.cs
dnduffy: Are keys really allowed to be 32768 characters long?  That seems excessive for a key in my opinion.

Src/PerformanceCollector/Unit.Tests/QuickPulse/QuickPulseServiceClientTests.cs
```diff
@@ -247,9 +248,9 @@ public void QuickPulseServiceClientFillsInTelemetryDocumentsWhenSubmittingToServ
                                 new ConcurrentStack<ITelemetryDocument>(
                                 new ITelemetryDocument[]
                                     {
-                                        new RequestTelemetryDocument() { Id = "Request1" },
-                                        new DependencyTelemetryDocument() { Id = "Dependency1" },
-                                        new ExceptionTelemetryDocument() { Exception = "Exception1" }
+                                        new RequestTelemetryDocument() { Name = "Request1", Properties = properties.ToArray() },
+                                        new DependencyTelemetryDocument() { Name = "Dependency1", Properties = properties.ToArray() },
+                                        new ExceptionTelemetryDocument() { Exception = "Exception1", Properties = properties.ToArray() }
```
dnduffy: I know it's just a test but why reuse the same properties as the value?

Src/PerformanceCollector/Unit.Tests/QuickPulse/QuickPulseTelemetryProcessorTests.cs
```diff
@@ -402,22 +404,28 @@ public void QuickPulseTelemetryProcessorCollectsFullTelemetryItems()
             // ACT
             var request = new RequestTelemetry()
                               {
-                                  Id = Guid.NewGuid().ToString(),
+                                  Name = Guid.NewGuid().ToString(),
```
dnduffy: I'm trying to understand, why is the new Guid set to an "Id" property in the others but "Name" here, particularly when this used to be Id?

Src/PerformanceCollector/Unit.Tests/QuickPulse/QuickPulseTelemetryProcessorTests.cs
```diff
@@ -694,5 +710,300 @@ public void QuickPulseTelemetryProcessorDoesNotCollectFullTelemetryItemsWhenSwit
...
+            telemetryProcessor.Process(requestLong);
+            telemetryProcessor.Process(requestShort);
+
+            // ASSERT
+            Assert.AreEqual(
+                ((RequestTelemetryDocument)accumulatorManager.CurrentDataAccumulator.TelemetryDocuments.ToList()[0]).Name,
+                requestShort.Name);
+            Assert.AreEqual(
+                ((RequestTelemetryDocument)accumulatorManager.CurrentDataAccumulator.TelemetryDocuments.ToList()[1]).Name,
+                requestShort.Name);
```
dnduffy: Should one of these be requestLong.Name?

Src/PerformanceCollector/Unit.Tests/QuickPulse/QuickPulseTelemetryProcessorTests.cs
```diff
@@ -694,5 +710,300 @@ public void QuickPulseTelemetryProcessorDoesNotCollectFullTelemetryItemsWhenSwit
...
+            };
+
+            // process in the opposite order to allow for an easier validation order
+            telemetryProcessor.Process(requestLong);
+            telemetryProcessor.Process(requestShort);
+
+            // ASSERT
+            Assert.AreEqual(
+                ((RequestTelemetryDocument)accumulatorManager.CurrentDataAccumulator.TelemetryDocuments.ToList()[0]).Properties.First().Key,
+                requestShort.Properties.First().Key);
```
dnduffy: You should only call "accumulatorManager.CurrentDataAccumulator.TelemetryDocuments.ToList()" once and cache the result.  There is no reason to keep converting to list and wasting memory and GC cycles.

Src/PerformanceCollector/Unit.Tests/QuickPulse/QuickPulseTelemetryProcessorTests.cs
```diff
@@ -694,5 +710,300 @@ public void QuickPulseTelemetryProcessorDoesNotCollectFullTelemetryItemsWhenSwit
...
+            Assert.AreEqual(
+                ((RequestTelemetryDocument)accumulatorManager.CurrentDataAccumulator.TelemetryDocuments.ToList()[0]).Properties.First().Value,
+                requestShort.Properties.First().Value);
+            Assert.AreEqual(
+                ((RequestTelemetryDocument)accumulatorManager.CurrentDataAccumulator.TelemetryDocuments.ToList()[1]).Properties.First().Key,
+                requestShort.Properties.First().Key);
+            Assert.AreEqual(
+                ((RequestTelemetryDocument)accumulatorManager.CurrentDataAccumulator.TelemetryDocuments.ToList()[1]).Properties.First().Value,
+                requestShort.Properties.First().Value);
```
dnduffy: Nothing seems to assert against requestLong...

Src/PerformanceCollector/Unit.Tests/QuickPulse/QuickPulseTelemetryProcessorTests.cs
```diff
@@ -694,5 +710,300 @@ public void QuickPulseTelemetryProcessorDoesNotCollectFullTelemetryItemsWhenSwit
...
+            Assert.AreEqual(
+                ((DependencyTelemetryDocument)accumulatorManager.CurrentDataAccumulator.TelemetryDocuments.ToList()[0]).CommandName,
+                dependencyShort.CommandName);
+            Assert.AreEqual(
+                ((DependencyTelemetryDocument)accumulatorManager.CurrentDataAccumulator.TelemetryDocuments.ToList()[1]).CommandName,
+                dependencyLong.CommandName.Substring(0, MaxFieldLength));
+        }
+
+        [TestMethod]
+        public void QuickPulseTelemetryProcessorTruncatesLongFullDependencyTelemetryItemName()
```
dnduffy: According to the test name the point of the test is to check for truncation so it seems that the test could be greatly simplified and things like dependencyShort could be left out.

Src/PerformanceCollector/Unit.Tests/QuickPulse/QuickPulseTelemetryProcessorTests.cs
```diff
@@ -694,5 +710,300 @@ public void QuickPulseTelemetryProcessorDoesNotCollectFullTelemetryItemsWhenSwit
...
+                new string('c', MaxFieldLength + 1),
+                new string('c', MaxFieldLength + 1),
+                DateTimeOffset.Now,
+                TimeSpan.FromSeconds(1),
+                false)
+            { Context = { InstrumentationKey = instrumentationKey } };
+
+            // process in the opposite order to allow for an easier validation order
+            telemetryProcessor.Process(dependencyLong);
+            telemetryProcessor.Process(dependencyShort);
```
dnduffy: There is a lot of duplicated code in these tests, please make helper methods to remove the duplication and shorten up the code.

Src/PerformanceCollector/Unit.Tests/QuickPulse/QuickPulseTelemetryProcessorTests.cs
```diff
@@ -694,5 +710,300 @@ public void QuickPulseTelemetryProcessorDoesNotCollectFullTelemetryItemsWhenSwit
...
+                dependencyShort.Properties.First().Key);
+            Assert.AreEqual(
+                ((DependencyTelemetryDocument)accumulatorManager.CurrentDataAccumulator.TelemetryDocuments.ToList()[0]).Properties.First().Value,
+                dependencyShort.Properties.First().Value);
+            Assert.AreEqual(
+                ((DependencyTelemetryDocument)accumulatorManager.CurrentDataAccumulator.TelemetryDocuments.ToList()[1]).Properties.First().Key,
+                dependencyShort.Properties.First().Key);
+            Assert.AreEqual(
+                ((DependencyTelemetryDocument)accumulatorManager.CurrentDataAccumulator.TelemetryDocuments.ToList()[1]).Properties.First().Value,
+                dependencyShort.Properties.First().Value);
```
dnduffy: Cache the stuff that is expensive like .ToList() and .First() and reuse instead of recreating and reiterating.

Src/PerformanceCollector/Shared/QuickPulseTelemetryProcessor.cs
tokaplan: I don't think we're quite there yet, but eventually yes.

Src/PerformanceCollector/Unit.Tests/QuickPulse/QuickPulseTelemetryProcessorTests.cs
tokaplan: Of course, but this is a unit test, so I don't care about cpu cycles :8ball:

dnduffy: Faster execution of unit tests means less wait time when running them all.  Plus it would make the lines a little shorter and easier to read.

tokaplan: Done here and elsewhere.

Src/PerformanceCollector/Unit.Tests/QuickPulse/QuickPulseTelemetryProcessorTests.cs
tokaplan: The lack of sign-offs made me care (go figure!), so I've fixed this everywhere.

Src/PerformanceCollector/Unit.Tests/QuickPulse/QuickPulseTelemetryProcessorTests.cs
tokaplan: I've taken another look, and I believe you're talking about the beginning of each test (where we create the objects under test and initialize them). The issue I'm seeing is that should I refactor that into a method, I would not going anything since three objects would be outputs of such method, so readability would still be the same (while all flexibility is lost). If you see a good option for extracting repetitive initialization out, please let me know.

Src/PerformanceCollector/Unit.Tests/QuickPulse/QuickPulseTelemetryProcessorTests.cs
dnduffy: Thank-you. :)

Src/PerformanceCollector/Unit.Tests/QuickPulse/QuickPulseTelemetryProcessorTests.cs
dnduffy: Probably the cleanest I can think of that doesn't involve an ugly helper method with out parameters would be to use the [TestInitialize()] attribute on a test setup method which creates and sets the variables you need each time.  The test framework will run that attributed init method before it runs each test so you could make the variables you need class instance level and then trust the test framework to always run that method before running each test.
If however your tests are only a few out of many in this file then it won't make sense to do that because there is no reason to have those variables re-created for tests which don't use them.

dnduffy:
Commit, push, and :shipit:!

abaranch:
:shipit:

PR #131, opened 2016-09-07
Added common.targets to all projects and added injection of language into XML documentation files.

dnduffy:


AddXmlLanguage.targets
```diff
@@ -0,0 +1,65 @@
...
+                        }
+
+                        if (filePath.EndsWith(".xml", StringComparison.OrdinalIgnoreCase))
+                        {
+                            string[] files = null;
+                            int starIndex = filePath.IndexOf('*');
+                            if (starIndex >= 0)
+                            {
+                                //string[] directories = Directory.GetDirectories("");
+                                //string[] files = Directory.GetFiles(Path.GetDirectoryName(filePath));
```
abaranch: Can be deleted

abaranch:
:shipit:

AddXmlLanguage.targets
dnduffy: Oops, good catch.

PR #133, opened 2016-09-09
Fix ps1 signing

abaranch:


Signing.props
```diff
@@ -11,27 +11,20 @@
...
-  <!-- Sign PowerShell scripts and modules copied to the output directory of the project -->
-  <Target Name="PreparePowerShellScriptsForSigning" BeforeTargets="SignFiles">
     <ItemGroup>
-      <FilesToSign Include="$(OutputPath)\**\*.ps*1">
+      <!-- Automatically sign binary generated by the project that imports Signing.props and all ps1 files -->
+      <FilesToSign Include="$(TargetPath)" >
+        <Authenticode>Microsoft</Authenticode>
+        <StrongName>MsSharedLib72</StrongName>
+      </FilesToSign>
+      <FilesToSign Include="$(OutputPath)*.ps1">
```
dnduffy: Just to be safe, does $(OutputPath) end in a '\' and all the powershell scripts are in that folder such that the ** wildcard is not needed?

dnduffy:
:+1:

Signing.props
abaranch: Yes


---
In reply to: [78254308](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/133#discussion_r78254308) [](ancestors = 78254308)

PR #146, opened 2016-09-23
removed performance counters and session state

SergeyKanzhelev:
Breeze will convert custom metric with the certain names to the performance counters. So UI wouldn't be broken


Src/PerformanceCollector/Shared/PerformanceCollectorModule.cs
```diff
@@ -156,28 +195,20 @@ public void Dispose()
...
-                metricTelemetry.Properties.Add("CounterInstanceName", pc.InstanceName);
-                metricTelemetry.Properties.Add("CustomPerfCounter", "true");
-
-                return metricTelemetry;
-            }
-            else
-            {
-                return new PerformanceCounterTelemetry(pc.CategoryName, pc.CounterName, pc.InstanceName, value);
-            }
+            var metricName = !string.IsNullOrWhiteSpace(reportAs)
```
dnduffy: metricName is just a string, right?
string metricName...?

abaranch:
Please update changelog

PR #149, opened 2016-09-27
Public schema update

SergeyKanzhelev:
Switching to the latest Core with the schema update. On top of this PR will be another that will add telemetry initializer that will parse http dependencies into Azure Storage dependencies.


Src/DependencyCollector/Shared/Implementation/FrameworkHttpProcessing.cs
```diff
@@ -54,6 +54,14 @@ public void OnBeginHttpCallback(long id, string resourceName)
                     return;
                 }
 
+                var url = new Uri(resourceName);
```
abaranch: >resourceName [](start = 34, length = 12)

Is that possible that it is incorrect uri? Are we OK to throw?

Src/DependencyCollector/Shared/Implementation/FrameworkHttpProcessing.cs
```diff
@@ -54,6 +54,14 @@ public void OnBeginHttpCallback(long id, string resourceName)
...
+
+                if (url == null)
+                {
```
abaranch: It would throw for incorrect url

Src/DependencyCollector/Shared/Implementation/FrameworkHttpProcessing.cs
dnduffy: I believe this will throw an exception rather than return null which means the check later won't work and this should be wrapped in a try catch.

Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
```diff
@@ -215,10 +215,16 @@ private object OnBegin(object thisObj, bool isAsyncCall)
                     DependencyCollectorEventSource.Log.UnexpectedCallbackParameter("WebRequest");
                 }
 
-                string url = this.GetUrl(webRequest);
-                string httMethod = webRequest.Method;
+                var url = this.GetUrl(webRequest);
```
dnduffy: Uri url...?
This is not good readability with var.

SergeyKanzhelev:
I'll address comment later. Issue: https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/151

Test/DependencyCollector/FunctionalTests/FuncTest/HttpTests.cs
```diff
@@ -389,7 +389,7 @@ public void TestRddForWin32ApplicationPool()
                             DeploymentAndValidationTools.SleepTimeForSdkToSendEvents);
 
                     var httpItems =
-                        allItems.Where(i => i.data.baseData.dependencyKind == DependencyKind.Http).ToArray();
+                        allItems.Where(i => i.data.baseData.type == "Http").ToArray();
```
dnduffy: Repeating strings like "Http" should be in consts so there is only one copy to maintain.

Test/DependencyCollector/FunctionalTests/FuncTest/HttpTests.cs
```diff
@@ -602,7 +602,7 @@ private void ExecuteAzureSDKTests(TestWebApplication testWebApplication, int cou
             bool successFlagExpected,
             string verb)
         {
-            if (DependencySourceType.Apmc == DeploymentAndValidationTools.ExpectedSource)
+            if ("rddp" == DeploymentAndValidationTools.ExpectedSDKPrefix)
```
dnduffy: "rddp" const

Test/DependencyCollector/FunctionalTests/FuncTest/SqlTests.cs
```diff
@@ -98,11 +98,11 @@ public void TestStoredProcedureNameIsCollected()
                          //// The above request would have trigged RDD module to monitor and create RDD telemetry
                          //// Listen in the fake endpoint and see if the RDDTelemtry is captured                      
                          var allItems = DeploymentAndValidationTools.SdkEventListener.ReceiveAllItemsDuringTimeOfType<TelemetryItem<RemoteDependencyData>>(DeploymentAndValidationTools.SleepTimeForSdkToSendEvents);
-                         var sqlItems = allItems.Where(i => i.data.baseData.dependencyKind == DependencyKind.SQL).ToArray();
+                         var sqlItems = allItems.Where(i => i.data.baseData.type == "SQL").ToArray();
```
dnduffy: "SQL" const

Src/PerformanceCollector/Net40/Perf.Net40.csproj
```diff
@@ -25,7 +25,7 @@
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' " />
   <ItemGroup>
     <Reference Include="Microsoft.ApplicationInsights">
-      <HintPath>..\..\..\..\packages\Microsoft.ApplicationInsights.2.2.0-beta2\lib\net40\Microsoft.ApplicationInsights.dll</HintPath>
+      <HintPath>..\..\..\..\packages\Microsoft.ApplicationInsights.2.2.0-beta2-build31642\lib\net40\Microsoft.ApplicationInsights.dll</HintPath>
```
tokaplan: Is the "build31642" intended?

Src/PerformanceCollector/Net40/packages.config
```diff
@@ -1,12 +1,12 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="Desktop.Analyzers" version="1.1.0" targetFramework="net40" />
   <package id="MicroBuild.Core" version="0.2.0" targetFramework="net40" developmentDependency="true" />
-  <package id="Microsoft.ApplicationInsights" version="2.2.0-beta2" targetFramework="net40" />
+  <package id="Microsoft.ApplicationInsights" version="2.2.0-beta2-build31642" targetFramework="net40" />
```
tokaplan: Intended?

Src/PerformanceCollector/Shared/QuickPulseTelemetryProcessor.cs
```diff
@@ -191,12 +191,10 @@ private static ITelemetryDocument ConvertRequestToTelemetryDocument(RequestTelem
                            Timestamp = requestTelemetry.Timestamp,
                            OperationId = TruncateValue(requestTelemetry.Context?.Operation?.Id),
                            Name = TruncateValue(requestTelemetry.Name),
-                           StartTime = requestTelemetry.StartTime,
                            Success = IsRequestSuccessful(requestTelemetry),
                            Duration = requestTelemetry.Duration,
                            ResponseCode = requestTelemetry.ResponseCode,
                            Url = requestTelemetry.Url,
```
tokaplan: Are you guys removing StartTime and HttpMethod from RequestTelemetry? Are there alternatives?

Global.props
```diff
@@ -96,7 +96,7 @@
         <NugetVersionFilePath>$(MSBuildThisFileDirectory).nugetVersion</NugetVersionFilePath>
         <BuildNugetVersion Condition="Exists($(NugetVersionFilePath))">$([System.IO.File]::ReadAllText($(NugetVersionFilePath)))</BuildNugetVersion>
 
-        <CoreSdkVersion>2.2.0-beta2</CoreSdkVersion>
+        <CoreSdkVersion>2.2.0-beta2-build31642</CoreSdkVersion>
```
SergeyKanzhelev: we need a specific version to make sure the correct fixes in Core are picked up

Schema/PublicSchema/AvailabilityData.bond
```diff
@@ -1,42 +1,40 @@
...
+    41: required string 	 name;
     
     [Description("Duration in TimeSpan 'G' (general long) format: d:hh:mm:ss.fffffff")]
     [CSType("TimeSpan")]
     50: required string 	 duration;
     
-    60: required AI.TestResult 	 result;
+    [ActAsRequired("Renaming result to success.")]
+    [Description("Success flag.")]
+    61: required bool 	 success;
```
SergeyKanzhelev: test either `pass` or `fail`. We do not support `inconclusive`.

Src/DependencyCollector/Shared.Tests/Implementation/FrameworkSqlProcessingTest.cs
```diff
@@ -66,7 +65,8 @@ public void RddTestSqlProcessingFrameworkSendsCorrectTelemetrySqlQuerySucess()
             Assert.AreEqual(1, this.sendItems.Count, "Only one telemetry item should be sent");
             ValidateTelemetryPacket(
                 this.sendItems[0] as DependencyTelemetry,
-                this.resourceName,
+                "ourdatabase.database.windows.net | mydatabase",
```
SergeyKanzhelev: I didn't make it worse =) Let's fix it eventually

Src/PerformanceCollector/Shared/QuickPulseTelemetryProcessor.cs
SergeyKanzhelev: `TimeStamp` for `StartTime`, http method is a part of a `Name` - so removed the duplication

Src/DependencyCollector/Shared.Tests/Implementation/FrameworkSqlProcessingTest.cs
dnduffy: My worry is "eventually" = "never". ;)

PR #154, opened 2016-09-28
Sergkanz/fix build

SergeyKanzhelev:


Global.props
```diff
@@ -96,7 +96,7 @@
         <NugetVersionFilePath>$(MSBuildThisFileDirectory).nugetVersion</NugetVersionFilePath>
         <BuildNugetVersion Condition="Exists($(NugetVersionFilePath))">$([System.IO.File]::ReadAllText($(NugetVersionFilePath)))</BuildNugetVersion>
 
-        <CoreSdkVersion>2.2.0-beta2-build31642</CoreSdkVersion>
+        <CoreSdkVersion>2.2.0-beta3-build34090</CoreSdkVersion>
```
dnduffy: Why are there dependencies on builds with build numbers?
For dependencies I think we should produce builds that are "public facing" without a build number in the name.

NuGet.config
```diff
@@ -17,6 +17,6 @@
     </config>
   <packageSources>
     <add key="nuget.org" value="https://www.nuget.org/api/v2/" />
-    <add key="MyGetApplicationInsights" value="https://www.myget.org/F/applicationinsights/api/v3/index.json" />
+    <add key="MyGetApplicationInsights" value="https://www.myget.org/F/applicationinsights/" />
```
dnduffy: We should update the NuGet exe's and move everything to v3, not downgrade to v2, because v2 it is going away and we will be broken when it does.  Also, there is some new functionality in v3 that is required in the new .NET Core world (and probably will be elsewhere) with regards to NuGet packages that we produce (I don't recall the details off-hand).

SergeyKanzhelev: https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/155

Global.props
SergeyKanzhelev: We always have a specific version. When we build public versions - they wouldn't have - `builldFoo` suffix

SergeyKanzhelev:
@dnduffy any concerns merging it?

PR #157, opened 2016-09-29
Sergkanz/detect http dependencies

SergeyKanzhelev:
This PR contains parsing for Azure Blobs. Seeking for the feedback on telemetry initializer registration approach

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Version 2.2.0-beta3
 - New property `DefaultCounters` in `PerformanceCollectorModule` to control the list of standard counters that will be collected
 - Default performance counters will be reported as metrics
+- By default `DependencyTrackingTelemetryModule` will not detect certain http dependencies as Azure Storage calls. You need to register a telemetry initializer `HttpDependenciesParsingTelemetryInitializer` to enable this functionality. 
 
 ## Version 2.2.0-beta2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Version 2.2.0-beta3
 - New property `DefaultCounters` in `PerformanceCollectorModule` to control the list of standard counters that will be collected
 - Default performance counters will be reported as metrics
-- By default `DependencyTrackingTelemetryModule` will not detect certain http dependencies as Azure Storage calls. You need to register a telemetry initializer `HttpDependenciesParsingTelemetryInitializer` to enable this functionality. 
+- When you instantiate `DependencyTrackingTelemetryModule` in code it will not detect certain http dependencies as Azure Storage calls. You need to register a telemetry initializer `HttpDependenciesParsingTelemetryInitializer` to enable this functionality. This telemetry initializer will be registered automatically during NuGet installation.
 
 ## Version 2.2.0-beta2
 
--- a/Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
+++ b/Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
@@ -31,12 +31,6 @@
         public bool DisableRuntimeInstrumentation { get; set; }
 
         /// <summary>
-        /// Gets or sets a value indicating whether parsing of http dependencies to azure storage dependencies is enabled. 
-        /// This flag is required for the backward compatibility to make sure that Azure Storage dependencies will be recognized as such.
-        /// </summary>
-        public bool DoNotParseHttp { get; set; }
-
-        /// <summary>
         /// IDisposable implementation.
         /// </summary>
         public void Dispose()
@@ -61,24 +55,6 @@
                         try
                         {                            
                             this.telemetryConfiguration = configuration;
-
-                            if (!this.DoNotParseHttp && configuration != null)
-                            {
-                                var initializerRegistered = false;
-                                foreach (var initializer in configuration.TelemetryInitializers)
-                                {
-                                    if (initializer is HttpDependenciesParsingTelemetryInitializer)
-                                    {
-                                        initializerRegistered = true;
-                                        break;
-                                    }
-                                }
-
-                                if (!initializerRegistered)
-                                {
-                                    configuration.TelemetryInitializers.Add(new HttpDependenciesParsingTelemetryInitializer());
-                                }
-                            }
 
                             // Net40 only supports runtime instrumentation
                             // Net45 supports either but not both to avoid duplication
--- a/Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
+++ b/Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
@@ -24,11 +24,17 @@
         private TelemetryConfiguration telemetryConfiguration;
         private bool isInitialized = false;
         private bool disposed = false;
-        
+
         /// <summary>
         /// Gets or sets a value indicating whether to disable runtime instrumentation.
         /// </summary>
         public bool DisableRuntimeInstrumentation { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether parsing of http dependencies to azure storage dependencies is enabled. 
+        /// This flag is required for the backward compatibility to make sure that Azure Storage dependencies will be recognized as such.
+        /// </summary>
+        public bool DoNotParseHttp { get; set; }
 
         /// <summary>
         /// IDisposable implementation.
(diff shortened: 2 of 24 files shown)


Src/DependencyCollector/Shared/HttpDependenciesParsingTelemetryInitializer.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        /// <summary>
+        /// If telemetry item is http dependency - converts it to the well-known type of the dependency.
+        /// </summary>
+        /// <param name="telemetry">Telemetry item to convert.</param>
+        public void Initialize(ITelemetry telemetry)
+        {
+            var httpDependency = telemetry as DependencyTelemetry;
+
+            if (httpDependency != null && httpDependency.Type != null && httpDependency.Type.ToUpperInvariant() == "HTTP")
```
abaranch: >"HTTP" [](start = 115, length = 6)

Do we have a enum somewhere o at least consts?..

Src/DependencyCollector/Shared/HttpDependenciesParsingTelemetryInitializer.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                string host = httpDependency.Target;
+
+                if (!string.IsNullOrEmpty(host))
+                {
+                    if (host.EndsWith("blob.core.windows.net", StringComparison.OrdinalIgnoreCase))
+                    {
+                        httpDependency.Type = "Azure blob";
```
abaranch: >"Azure blob" [](start = 46, length = 12)

This we also need to move somewhere

Src/DependencyCollector/Shared/HttpDependenciesParsingTelemetryInitializer.cs
```diff
@@ -0,0 +1,70 @@
...
+                    ////}
+                    ////else if (host.EndsWith("queue.core.windows.net", StringComparison.OrdinalIgnoreCase))
+                    ////{
+                    ////    httpDependency.Type = "Azure queue";
+                    ////}
+                }
+            }
+        }
+
+        private static string GetVerb(string name, out string nameWithoutVerb)
```
abaranch: >GetVerb [](start = 30, length = 7)

Why are you sure there will be a verb? Can customer track http dependency without it and you cut it?..

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
```diff
@@ -56,6 +62,24 @@ public void Initialize(TelemetryConfiguration configuration)
                         {                            
                             this.telemetryConfiguration = configuration;
 
+                            if (!this.DoNotParseHttp && configuration != null)
```
abaranch: >DoNotParseHttp [](start = 38, length = 14)

Why we create a new mechanizm for adding initializers? Why hidden one? Why not in config?

Src/DependencyCollector/Shared/HttpDependenciesParsingTelemetryInitializer.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            var result = string.Empty;
+            nameWithoutVerb = name;
+
+            var idx = name.IndexOf(' ') + 1;
```
Dmitry-Matveev: This does not check for VERB, it checks for space. You'd need to validate result against the known list of verbs to be certain that it's a verb and not a dependency "Name" with a space in it.

Src/DependencyCollector/Shared/HttpDependenciesParsingTelemetryInitializer.cs
```diff
@@ -0,0 +1,70 @@
...
+                    if (host.EndsWith("blob.core.windows.net", StringComparison.OrdinalIgnoreCase))
+                    {
+                        httpDependency.Type = "Azure blob";
+
+                        string nameWithoutVerb = httpDependency.Name;
+                        var verb = GetVerb(httpDependency.Name, out nameWithoutVerb);
+
+                        var isFirstSlash = nameWithoutVerb[0] == '/' ? 1 : 0;
+                        var idx = nameWithoutVerb.IndexOf('/', isFirstSlash); // typically first symbol of the path is '/'
+                        string container = idx != -1 ? nameWithoutVerb.Substring(isFirstSlash, idx - isFirstSlash) : string.Empty;
```
Dmitry-Matveev: What about first two slashes? This code looks suspiciously "breezy" if you know what I mean. Do we have a set of the values generally seen as Name to base our decisions on? Or this is an educated guess?

Src/DependencyCollector/Shared/HttpDependenciesParsingTelemetryInitializer.cs
dnduffy: Would it be more efficient to do a case insensitive compare than to create a new string that is uppercase?

SergeyKanzhelev: Done

Src/DependencyCollector/Shared/HttpDependenciesParsingTelemetryInitializer.cs
```diff
@@ -0,0 +1,70 @@
...
+                    }
+
+                    ////else if (host.EndsWith("table.core.windows.net", StringComparison.OrdinalIgnoreCase))
+                    ////{
+                    ////    httpDependency.Type = "Azure table";
+                    ////}
+                    ////else if (host.EndsWith("queue.core.windows.net", StringComparison.OrdinalIgnoreCase))
+                    ////{
+                    ////    httpDependency.Type = "Azure queue";
+                    ////}
```
SergeyKanzhelev: It needs to be implemented. I only implemented blobs for now

Src/DependencyCollector/Shared/HttpDependenciesParsingTelemetryInitializer.cs
SergeyKanzhelev: Aha. Done

Src/DependencyCollector/Shared/HttpDependenciesParsingTelemetryInitializer.cs
SergeyKanzhelev: Customer can. This is the logic from Breeze. What would be your suggestion?

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
SergeyKanzhelev: That was the main qq in this PR. Today when you do `new DependencyTelemetryModule()` you'll collect http dependenies as Azure Storage because Breeze will do transformation. Since logic moved from Breeze to SDK I was thinking it will be logical to register it be default.

Do you think it reasonable or we need to write release notes?

Src/DependencyCollector/Shared/HttpDependenciesParsingTelemetryInitializer.cs
SergeyKanzhelev: Do you mean the list of VERBs that Azure blob supports? I guess it can be done

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
abaranch: I think it is reasonable to add it by default. Let's add it by default in configuration file so it is not hidden from customers and can be removed by deleting from there rather than some undocumented property.

SergeyKanzhelev: Fixed

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Version 2.2.0-beta3
 - New property `DefaultCounters` in `PerformanceCollectorModule` to control the list of standard counters that will be collected
 - Default performance counters will be reported as metrics
+- By default `DependencyTrackingTelemetryModule` will not detect certain http dependencies as Azure Storage calls. You need to register a telemetry initializer `HttpDependenciesParsingTelemetryInitializer` to enable this functionality.
```
abaranch: This is confusing... You are adding it by default in config. I code-based instrumentation is used than you do not get it by default.

SergeyKanzhelev:
I'll merge these changes and will work on supporting other storage account types. This way PR will be much smaller in size

PR #169, opened 2016-10-05
Web SDK may lose data because HttpContext is null on EndRequest

abaranch:
#162


Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
```diff
@@ -67,7 +88,16 @@ private void OnBeginRequest(object sender, EventArgs eventArgs)
         {
             if (this.isEnabled)
             {
-                this.TraceCallback("OnBegin", (HttpApplication)sender);
+                HttpApplication httpApplication = (HttpApplication)sender;
```
dnduffy: Just curious if this cast is always true or if using "as HttpApplication" would be safer? (If it's always true then leave it this way)

Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
```diff
@@ -26,6 +32,21 @@ public ApplicationInsightsHttpModule()
...
+
+                foreach (var module in TelemetryModules.Instance.Modules)
+                {
+                    if (module is RequestTrackingTelemetryModule)
+                    {
+                        this.requestModule = (RequestTrackingTelemetryModule)module;
+                    }
+                    else
+                    {
+                        if (module is ExceptionTrackingTelemetryModule)
```
dnduffy: Why not just "else if" instead of nested "if"?

Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
```diff
@@ -81,6 +111,17 @@ private void OnEndRequest(object sender, EventArgs eventArgs)
 
                 if (this.IsFirstRequest(httpApplication))
```
dnduffy: Something feels weird to me about "is first" logic but I don't currently recall why...

Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
abaranch: It is safe to cast.

Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
abaranch: :) I did not change that. You can talk to @madeleinefletcher if you understand what is the question

Issue #127, opened 2016-09-02
Number of arguments were changed for some of internal SQL methods we are using


SergeyKanzhelev:
`BeginExecuteNonQueryInternal` [expects](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/7c86689810be38a8a8a412c0720a4f2614d7207d/Src/DependencyCollector/Shared/Implementation/ProfilerRuntimeInstrumentation.cs#L122) 4 arguments. Actual signature:

``` csharp
// System.Data.SqlClient.SqlCommand
private IAsyncResult BeginExecuteNonQueryInternal(CommandBehavior behavior, AsyncCallback callback, object stateObject, int timeout, bool inRetry, bool asyncWrite = false)
```

Same for some other `BeginFoo` methods

PR #170, opened 2016-10-05
Abaranch/sql bug

abaranch:
#127


Src/DependencyCollector/Shared.Tests/Implementation/ProfilerSqlProcessingTest.cs
```diff
@@ -45,566 +49,256 @@ public void TestInitialize()
...
 
-        /// <summary>
-        /// Validates SQLProcessingProfiler sends correct telemetry on calling OnExceptionForExecuteReader.
-        /// </summary>
         [TestMethod]
-        [Description("Validates SQLProcessingProfiler sends correct telemetry on calling OnExceptionForExecuteReader.")]
-        [Owner("cithomas")]
-        [TestCategory("CVT")]
-        public void RddTestSqlProcessingProfilerBeginAndForBeginExecuteReaderInternalException()
+        public void OnExceptionSendsCorrectTelemetry3()
```
dnduffy: Is there anything more we can add to distinguish these test names with instead of 1, 2, 3?

Src/DependencyCollector/Shared/Implementation/ProfilerRuntimeInstrumentation.cs
```diff
@@ -69,114 +69,249 @@ internal static void DecorateProfilerForHttp(ref ProfilerHttpProcessing httpCall
...
-                sqlCallbacks.OnEndForSync,
-                sqlCallbacks.OnExceptionForSync,
+                sqlCallbacks.OnBeginForOneParameter,
+                sqlCallbacks.OnEndForOneParameter,
+                sqlCallbacks.OnExceptionForOneParameter,
                 isStatic: false);
 
-            // Decorate Sql ExecuteReader, 2 params
+            // Decorate Sql ExecuteNonQueryAsync(CancellationToken)
+            // Only latest instrumentation engine supports Tasks so we need to uncomment this code and add funcs when StatusMonitor is out
```
dnduffy: Should put a "// TODO: `<alias>` - 2016/10/05 - ..." comment.

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessing.cs
```diff
@@ -52,154 +51,86 @@ internal ProfilerSqlProcessing(TelemetryConfiguration configuration, string agen
...
-        /// On exception callback for ExecuteReader.
+        /// On begin callback for methods with 4 parameter.
         /// </summary>
-        /// <param name="context">The context.</param>
-        /// <param name="exception">The exception.</param>
-        /// <param name="thisObj">This object.</param>
-        /// <param name="behavior">The callback parameter.</param>
-        /// <param name="method">The state parameter.</param>
-        public void OnExceptionForExecuteReader(object context, object exception, object thisObj, object behavior, object method)
+        public object OnBeginForFourParameters(object thisObj, object parameter1, object parameter2, object parameter3)
```
dnduffy: I admit the previous param doc comments were not detailed but it would be nice to have StyleCop compliance with good ones.

Test/DependencyCollector/FunctionalTests/FuncTest/SqlTests.cs
```diff
@@ -372,7 +387,7 @@ private void ExecuteSyncSqlTests(TestWebApplication testWebApplication, int expe
                     //// The above request would have trigged RDD module to monitor and create RDD telemetry
                     //// Listen in the fake endpoint and see if the RDDTelemtry is captured
 
-                    var allItems = DeploymentAndValidationTools.SdkEventListener.ReceiveAllItemsDuringTimeOfType<TelemetryItem<RemoteDependencyData>>(DeploymentAndValidationTools.SleepTimeForSdkToSendEvents);
+                    var allItems = DeploymentAndValidationTools.SdkEventListener.ReceiveAllItemsDuringTimeOfType<TelemetryItem<RemoteDependencyData>>(DeploymentAndValidationTools.SleepTimeForSdkToSendEvents).ToArray();
                     var sqlItems = allItems.Where(i => i.data.baseData.type == "SQL").ToArray();
```
dnduffy: Array of what?  These usages of var are not clear and should have concrete types for human readability.

Test/DependencyCollector/FunctionalTests/FuncTest/SqlTests.cs
```diff
@@ -372,7 +387,7 @@ private void ExecuteSyncSqlTests(TestWebApplication testWebApplication, int expe
                     //// The above request would have trigged RDD module to monitor and create RDD telemetry
                     //// Listen in the fake endpoint and see if the RDDTelemtry is captured
 
-                    var allItems = DeploymentAndValidationTools.SdkEventListener.ReceiveAllItemsDuringTimeOfType<TelemetryItem<RemoteDependencyData>>(DeploymentAndValidationTools.SleepTimeForSdkToSendEvents);
+                    var allItems = DeploymentAndValidationTools.SdkEventListener.ReceiveAllItemsDuringTimeOfType<TelemetryItem<RemoteDependencyData>>(DeploymentAndValidationTools.SleepTimeForSdkToSendEvents).ToArray();
```
dnduffy: What is the value of this first ToArray()? The next line will just iterate again in the Where which should make this an expensive waste of a conversion and object creation.

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessing.cs
abaranch: Stylecop does not complain about missing parameters if they all miss. Xml documentation would be useless in this case because it is actually param1,2,3 etc The meaning will vary depending on when method you instrument. Here we can write only something like "First argument of callback method. Second argument ..." but stylecop will complain that descriptions differ only by one word.

---
In reply to: [82120587](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/170#discussion_r82120587) [](ancestors = 82120587)

Test/DependencyCollector/FunctionalTests/FuncTest/SqlTests.cs
abaranch: It is easier to debug this way and we do not need to be performant in functional tests

---
In reply to: [82121011](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/170#discussion_r82121011) [](ancestors = 82121011)

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessing.cs
dnduffy: Interesting, my version of StyleCop complains about missing parameters (or at least it used to).

Issue #151, opened 2016-09-27
Parsing url may throw if url is incorrect in FramewrokHttpProcessor


SergeyKanzhelev:
This is a feedback on PR. I create an issue to track this change and merge PR

PR #173, opened 2016-10-06
Try/catch for url creation

abaranch:
#151

### DIFF
--- a/Src/DependencyCollector/Shared.Tests/Implementation/FrameworkHttpProcessingTest.cs
+++ b/Src/DependencyCollector/Shared.Tests/Implementation/FrameworkHttpProcessingTest.cs
@@ -51,6 +51,12 @@ namespace Microsoft.ApplicationInsights.DependencyCollector.Implementation
 
         #region BeginEndCallBacks
 
+        [TestMethod]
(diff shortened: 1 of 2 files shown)


Src/DependencyCollector/Shared.Tests/Implementation/FrameworkHttpProcessingTest.cs
```diff
@@ -51,6 +51,12 @@ public void Cleanup()
 
         #region BeginEndCallBacks
 
+        [TestMethod]
+        public void OnBeginDoesNotThrowForIncorrectUrl()
+        {
+            this.httpProcessingFramework.OnBeginHttpCallback(100, "BadUrl"); // Should not throw
```
dnduffy: Just curious why a bad URL should not throw? (I'm sure I'm just missing some context)
Is it because the below code in FrameworkHttpProcessing.cs is doing the try catch now?  That looks good. :)

abaranch: Yes, no it has try/catch. In general we should never throw and crash customer app.

PR #178, opened 2016-10-19
add azure table support

SergeyKanzhelev:


Src/DependencyCollector/Shared/HttpDependenciesParsingTelemetryInitializer.cs
```diff
@@ -96,5 +98,49 @@ private bool TryParseAzureBlob(string host, string name, string url, out string
...
+        {
+            bool result = false;
+
+            account = null;
+            verb = null;
+            tableName = null;
+
+            if (name != null && host != null && url != null)
+            {
+                if (host.EndsWith("table.core.windows.net", StringComparison.OrdinalIgnoreCase))
```
dnduffy: Would it work and be safer to add a period in front of the domain ".table.core..."?

Src/DependencyCollector/Shared/HttpDependenciesParsingTelemetryInitializer.cs
```diff
@@ -96,5 +98,49 @@ private bool TryParseAzureBlob(string host, string name, string url, out string
...
+                        var verbPrefix = this.azureTableVerbPrefixes[i];
+                        if (name.StartsWith(verbPrefix, StringComparison.OrdinalIgnoreCase))
+                        {
+                            verb = name.Substring(0, verbPrefix.Length);
+                            nameWithoutVerb = name.Substring(verbPrefix.Length);
+                            break;
+                        }
+                    }
+
+                    var isFirstSlash = nameWithoutVerb[0] == '/' ? 1 : 0;
```
dnduffy: Is this supposed to be an int or a bool?

Src/DependencyCollector/Shared/HttpDependenciesParsingTelemetryInitializer.cs
```diff
@@ -96,5 +98,49 @@ private bool TryParseAzureBlob(string host, string name, string url, out string
...
+                    var idx = nameWithoutVerb.IndexOf('/', isFirstSlash); // typically first symbol of the path is '/'
```
dnduffy: Looks like this should be of type int.

Src/DependencyCollector/Shared/HttpDependenciesParsingTelemetryInitializer.cs
```diff
@@ -96,5 +98,49 @@ private bool TryParseAzureBlob(string host, string name, string url, out string
...
+
+                    string nameWithoutVerb = name;
+
+                    for (int i = 0; i < this.azureTableVerbPrefixes.Length; i++)
+                    {
+                        var verbPrefix = this.azureTableVerbPrefixes[i];
+                        if (name.StartsWith(verbPrefix, StringComparison.OrdinalIgnoreCase))
+                        {
+                            verb = name.Substring(0, verbPrefix.Length);
+                            nameWithoutVerb = name.Substring(verbPrefix.Length);
```
dnduffy: Why are the names prefixed with verbs like "GET"?
What happens if someone names their table something like "GettingDataTable" and you strip off the "Get" portion of the name?

Src/DependencyCollector/Shared/HttpDependenciesParsingTelemetryInitializer.cs
SergeyKanzhelev: aha

Src/DependencyCollector/Shared/HttpDependenciesParsingTelemetryInitializer.cs
dnduffy: Ah sorry, I didn't notice the space.

Src/DependencyCollector/Shared/HttpDependenciesParsingTelemetryInitializer.cs
```diff
@@ -86,9 +89,95 @@ private bool TryParseAzureBlob(string host, string name, string url, out string
...
+                            break;
+                        }
+                    }
+
                     var isFirstSlash = nameWithoutVerb[0] == '/' ? 1 : 0;
                     var idx = nameWithoutVerb.IndexOf('/', isFirstSlash); // typically first symbol of the path is '/'
```
dnduffy: nitpick: I would still really like to be able to read "bool" and "int" instead of "var".

PR #184, opened 2016-10-24
Implemenatation for populating source and target fields in request an…

nizarq:
…d dependency telemetry automatically.


Src/DependencyCollector/Shared/Implementation/RemoteDependencyConstants.cs
```diff
@@ -4,6 +4,7 @@ internal static class RemoteDependencyConstants
     {
         public const string SQL = "SQL";
         public const string HTTP = "Http";
+        public const string AI = "ApplicationInsights";
```
SergeyKanzhelev: value name is too long. Can you think of something shorter? Maybe `App`?

Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
```diff
@@ -92,13 +93,41 @@ private void OnBeginRequest(object sender, EventArgs eventArgs)
...
+        }
+
+        /// <summary>
+        /// When sending the response headers, allow request module to add the IKey's target hash.
+        /// </summary>
+        /// <param name="httpApplication">HttpApplication instance.</param>
+        private void HookOnSendRequestHeaders(HttpApplication httpApplication)
+        {
+            // We use reflection here because 'AddOnSendingHeaders' is only available post .net framework 4.5.2. Hence we call it if we can find it.
+            MethodInfo addOnSendingHeadersMethod = httpApplication.Response.GetType().GetMethod("AddOnSendingHeaders");
```
SergeyKanzhelev: have try{}catch to handle exceptions and keep working even if this failed. Also check `httpApplication` for `null`, same as `Response`

SergeyKanzhelev: possible exception just to give an example is not enough rights to use reflection.

Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
```diff
@@ -92,13 +93,41 @@ private void OnBeginRequest(object sender, EventArgs eventArgs)
...
+
+            if (addOnSendingHeadersMethod != null)
+            {
+                var parameters = new object[]
+                {
+                    new Action<HttpContext>((httpContext) =>
+                    {
+                        if (this.requestModule != null)
```
SergeyKanzhelev: look how we wrap all our callbacks in `try{}catch`. Do not allow runaway exceptions

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -84,10 +85,46 @@ public void OnEndRequest(HttpContext context)
...
+                // and get they instrumentation key from all possible sources in the process. Let's do that now.
+
+                // Todo (nizarq): Revisit this strategy, we are calling a hidden method where in the implemenation it directly says don't call this guy.
+                // Also this means requestTelemetry is getting initialized twice - may be that will cause issues. 
+                this.telemetryClient.Initialize(requestTelemetry);
+            }
+
+            if (!string.IsNullOrEmpty(requestTelemetry.Context.InstrumentationKey))
+            {
+                context.Response.Headers[ComponentCorrelationConstants.TargetInstrumentationKeyHeader] = InstrumentationKeyHashLookupHelper.GetInstrumentationKeyHash(requestTelemetry.Context.InstrumentationKey);
```
SergeyKanzhelev: same as before - do not override if it was set already

Src/Common/InstrumentationKeyHashLookupHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+        public static string GetInstrumentationKeyHash(string instrumentationKey)
+        {
+            if (string.IsNullOrEmpty(instrumentationKey))
+            {
+                return null;
+            }
+
+            if (!knownIKeyHashes.ContainsKey(instrumentationKey))
+            {
+                knownIKeyHashes[instrumentationKey] = GenerateSHA256Hash(instrumentationKey);
```
dnduffy: Constant or a variable?  I agree with returning a variable from here rather than falling through to the return below.

Src/Common/InstrumentationKeyHashLookupHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// Computes the SHA256 hash for a given value.
+        /// </summary>
+        /// <param name="value">Value for which the hash is to be computed.</param>
+        /// <returns>Hash string.</returns>
+        private static string GenerateSHA256Hash(string value)
+        {
+            string hashString = string.Empty;
+
+            var sha256 = SHA256Managed.Create();
+            byte[] hash = sha256.ComputeHash(Encoding.UTF8.GetBytes(value));
```
dnduffy: The SHA256 hash will produce different values for different casings of the instrumentation key so perhaps the code should always toLower or toUpper it (whichever FxCop/StyleCop prefers).

Src/Common/InstrumentationKeyHashLookupHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            foreach (byte x in hash)
+            {
+                hashString += string.Format(CultureInfo.InvariantCulture, "{0:x2}", x);
```
dnduffy: This concatenation of strings in a loop is a .NET (and Java) really bad practice.
At a minimum you should use a string builder constructed with the known length of the hash byte array.

Src/DependencyCollector/Shared/Implementation/RemoteDependencyConstants.cs
dnduffy: "AIKey"?

Src/Common/InstrumentationKeyHashLookupHelper.cs
dnduffy: IKeys are just GUID's so they are not case sensitive, but this process would get different values depending on the ASCII casing of the GUID text.
I now wonder why this needs to be hashed at all?  What is the point of the hash?

Src/Common/InstrumentationKeyHashLookupHelper.cs
nizarq: fixed

Src/Common/InstrumentationKeyHashLookupHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+{
+    using System.Collections.Concurrent;
+    using System.Globalization;
+    using System.Security.Cryptography;
+    using System.Text;
+
+    /// <summary>
+    /// A store for instrumentation key hashes. This is order so as to optimize the computation of hashes.
+    /// </summary>
+    public static class InstrumentationKeyHashLookupHelper
```
nizarq: fixed

Src/Common/InstrumentationKeyHashLookupHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+        private static ConcurrentDictionary<string, string> knownIKeyHashes = new ConcurrentDictionary<string, string>();
```
nizarq: done

Src/Common/InstrumentationKeyHashLookupHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// <param name="instrumentationKey">Instrumentation key string.</param>
+        /// <returns>SHA256 hash for the instrumentation key.</returns>
+        public static string GetInstrumentationKeyHash(string instrumentationKey)
+        {
+            if (string.IsNullOrEmpty(instrumentationKey))
+            {
+                return null;
+            }
+
+            if (!knownIKeyHashes.ContainsKey(instrumentationKey))
```
nizarq: done

Src/Common/InstrumentationKeyHashLookupHelper.cs
nizarq: done

Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
```diff
@@ -261,6 +263,20 @@ private object OnBegin(object thisObj, bool isAsyncCall)
...
+                    // Instrumentation key is probably empty, because the context has not yet had a chance to associate the requestTelemetry to the telemetry client yet.
+                    // and get they instrumentation key from all possible sources in the process. Let's do that now.
+                    this.telemetryClient.Initialize(telemetry);
+                }
+
+                // Add the source instrumentation key header
+                // We should probably store a hash rather than the whole key
+                if (!string.IsNullOrEmpty(telemetry.Context.InstrumentationKey))
+                {
+                    webRequest.Headers.Add(ComponentCorrelationConstants.SourceInstrumentationKeyHeader, InstrumentationKeyHashLookupHelper.GetInstrumentationKeyHash(telemetry.Context.InstrumentationKey));
```
nizarq: change done - unit test pending - will add it along with other unit tests.

Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
```diff
@@ -343,6 +360,13 @@ private void OnEnd(object exception, object thisObj, object returnValue)
                             // ObjectDisposedException is expected here in the following sequence: httpWebRequest.GetResponse().Dispose() -> httpWebRequest.GetResponse()
                             // on the second call to GetResponse() we cannot determine the statusCode.
                         }
+
+                        var targetIkeyHash = responseObj.Headers[ComponentCorrelationConstants.TargetInstrumentationKeyHeader];
```
nizarq: Are you specifically talking about the line var targetIkeyHash = responseObj.Headers[ComponentCorrelationConstants.TargetInstrumentationKeyHeader]; 
Why do you think it will ever throw? Or are you talking about the catch block above where we are swallowing ObjectDisposeException?

Src/DependencyCollector/Shared/Implementation/RemoteDependencyConstants.cs
nizarq: What's the problem with a long descriptive value name? 
I don't like App because it doesn't make abundantly clear to the customer that this is a dependency type that says I called another app that's writing to an application insights resource. 
AIKey might be too cryptic.

Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
nizarq: comment updated - is that what you were looking for?

Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
nizarq: fixed.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
nizarq: fixed.

Src/Common/InstrumentationKeyHashLookupHelper.cs
nizarq: Just checked. Case insensitive. Got telemetry regardless of ikey casing. Fixed the code here to always deal with lowercase when computing hashes.

Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
```diff
@@ -261,6 +263,20 @@ private object OnBegin(object thisObj, bool isAsyncCall)
...
+                if (string.IsNullOrEmpty(telemetry.Context.InstrumentationKey))
+                {
+                    // Instrumentation key is probably empty, because the context has not yet had a chance to associate the requestTelemetry to the telemetry client yet.
+                    // and get they instrumentation key from all possible sources in the process. Let's do that now.
+                    this.telemetryClient.Initialize(telemetry);
+                }
+
+                // Add the source instrumentation key header
+                // We should probably store a hash rather than the whole key
+                if (!string.IsNullOrEmpty(telemetry.Context.InstrumentationKey))
```
nizarq: pending

Src/Common/RequestResponseHeaders.cs
```diff
@@ -0,0 +1,53 @@
...
+            /// <summary>
+            /// Standard root id header.
+            /// </summary>
+            public const string StandardRootIdHeader = "x-ms-request-root-id";
+        }
+
+        /// <summary>
+        /// Headers related to quick pulse service.
+        /// </summary>
+        internal static class QuickPulseService
```
dnduffy: I am not a fan of nested classes pretending to be namespaces...

nizarq: I am not a fan of nested classes either, except for classes only containing constants, where I feel it works really well in grouping related constants. It's your codebase - if you have a strong preference against it, I can change - otherwise, I think for classes containing constants - I like the pattern Constant.Group.Name

Src/Web/Web.Shared.Net/Implementation/WebEventSource.cs
```diff
@@ -331,6 +331,18 @@ public void NoHttpContextWarning(string appDomainName = "Incorrect")
                 this.ApplicationName);
         }
 
+        [Event(
+            35,
+            Message = "Failed to hook onto AddOnSendingHeaders with exception {0}",
```
nizarq: fixed

Src/Common/InstrumentationKeyHashLookupHelper.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// Computes the SHA256 hash for a given value and returns it in the form of a base64 encoded string.
+        /// </summary>
+        /// <param name="value">Value for which the hash is to be computed.</param>
+        /// <returns>Base64 encoded hash string.</returns>
+        private static string GenerateEncodedSHA256Hash(string value)
+        {
+            var sha256 = SHA256Managed.Create();
+            byte[] hash = sha256.ComputeHash(Encoding.UTF8.GetBytes(value));
+
+            return Convert.ToBase64String(hash);
```
dnduffy: Oh I agree 100% it should be treated as a secret.
Currently it is plainly exposed in the JavaScript though... ;)

Src/Common/RequestResponseHeaders.cs
dnduffy: When reading it with no context that reads nicely, but when trying to find it (without F12 in VS) you would expect there to be a file named "Group.cs" in a folder named "Constant" but neither would exist.  I personally find that to be inconsistent with the rest of the source tree and feel that it makes being a code librarian harder.  @SergeyKanzhelev do you have a strong preference?

Src/Common/InstrumentationKeyHashLookupHelper.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <returns>SHA256 hash for the instrumentation key.</returns>
+        public static string GetInstrumentationKeyHash(string instrumentationKey)
+        {
+            if (string.IsNullOrEmpty(instrumentationKey))
+            {
+                return null;
+            }
+
+            // Let's make sure we don't get different hashes for different casing.
+            instrumentationKey = instrumentationKey.ToLowerInvariant();
```
nizarq: yes I did!!!

Src/Common/InstrumentationKeyHashLookupHelper.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <summary>
+        /// Retrieves the hash for a given instrumentation key.
+        /// </summary>
+        /// <param name="instrumentationKey">Instrumentation key string.</param>
+        /// <returns>SHA256 hash for the instrumentation key.</returns>
+        public static string GetInstrumentationKeyHash(string instrumentationKey)
+        {
+            if (string.IsNullOrEmpty(instrumentationKey))
+            {
+                return null;
```
nizarq: Already checking for null before calling this. Changed to throwing null-ref.

Src/Common/InstrumentationKeyHashLookupHelper.cs
nizarq: done

Src/Common/InstrumentationKeyHashLookupHelper.cs
```diff
@@ -0,0 +1,66 @@
...
+
+            if (!found)
+            {
+                // Simplistic cleanup to guard against this becoming a memory hog.
+                if (knownIKeyHashes.Keys.Count >= MAXSIZE)
+                {
+                    knownIKeyHashes.Clear();
+                }
+
+                knownIKeyHashes[instrumentationKey] = hash = GenerateEncodedSHA256Hash(instrumentationKey);
```
nizarq: done

Src/Common/InstrumentationKeyHashLookupHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+                if (!found)
+                {
+                    // Simplistic cleanup to guard against this becoming a memory hog.
+                    if (knownIKeyHashes.Keys.Count >= MAXSIZE)
+                    {
+                        knownIKeyHashes.Clear();
+                    }
+
+                    hash = GenerateEncodedSHA256Hash(instrumentationKey);
+                    knownIKeyHashes[instrumentationKey] = hash;
```
SergeyKanzhelev: In order for the first lookup to work you need to use original instrumentation key as a key in the Dictionary

Src/Common/RequestResponseHeaders.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// Standard root id header.
+        /// </summary>
+        public const string StandardRootIdHeader = "x-ms-request-root-id";
+
+        /// <summary>
+        /// Subscribed header.
+        /// </summary>
+        public const string XMsQpsSubscribedHeaderName = "x-ms-qps-subscribed";
```
SergeyKanzhelev: As I said, let's only keep headers we will use in user's http communication in this file. QPS-specific headers should go back to QPS code

Src/DependencyCollector/Shared.Tests/Implementation/ProfilerHttpProcessingTest.cs
```diff
@@ -96,6 +98,64 @@ public void RddTestHttpProcessingProfilerOnEndForGetResponse()
         }
 
         /// <summary>
+        /// Validates if DependencyTelemetry sent contains the cross component instrumentation key hash.
+        /// </summary>
+        [TestMethod]
+        [Description("Validates if DependencyTelemetry sent contains the cross component IKey hash.")]
+        public void DependencyTargetContainsIkeyTest()
```
SergeyKanzhelev: Name it "ActionAssertResult". See how we name other unit test methods

Src/Web/Web.Shared.Net.Tests/RequestTrackingTelemetryModuleTest.cs
```diff
@@ -272,6 +273,24 @@ public void SdkVersionHasCorrectFormat()
             Assert.Equal(expectedVersion, context.GetRequestTelemetry().Context.GetInternalContext().SdkVersion);
         }
 
+        [TestMethod]
+        public void EnsureSourceFieldFilled()
```
SergeyKanzhelev: Same comment about the test method name

Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
```diff
@@ -92,13 +93,52 @@ private void OnBeginRequest(object sender, EventArgs eventArgs)
...
+                    // Not using reflection would result in MissingMethodException when 4.5 or 4.5.1 is present. 
+                    MethodInfo addOnSendingHeadersMethod = httpApplication.Response.GetType().GetMethod("AddOnSendingHeaders");
+
+                    if (addOnSendingHeadersMethod != null)
+                    {
+                        var parameters = new object[]
+                        {
+                            new Action<HttpContext>((httpContext) =>
+                            {
+                                if (this.requestModule != null)
```
SergeyKanzhelev: I still do not see try{}catch here. Runaway exception in this callback will fail the customer's code

nizarq: fixed.

Src/Common/RequestResponseHeaders.cs
nizarq: moved it to a QPS specific constants file in the same folder where other QPS code is located.

Src/DependencyCollector/Shared.Tests/Implementation/ProfilerHttpProcessingTest.cs
nizarq: done

Src/Web/Web.Shared.Net.Tests/RequestTrackingTelemetryModuleTest.cs
nizarq: done

Src/Common/InstrumentationKeyHashLookupHelper.cs
nizarq: done

Src/Common/InstrumentationKeyHashLookupHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+
+                if (!found)
+                {
+                    // Simplistic cleanup to guard against this becoming a memory hog.
+                    if (knownIKeyHashes.Keys.Count >= MAXSIZE)
+                    {
+                        knownIKeyHashes.Clear();
+                    }
+
+                    hash = GenerateEncodedSHA256Hash(instrumentationKey);
```
SergeyKanzhelev: now you forgot to do tolowerinvariant for the value you calculate the hash for. At this momemt I curious whether you have tests checking different cases. Please add if you don't have them yet

CONTRIBUTING.md
```diff
@@ -52,7 +52,8 @@ If all or most of the Dependency Collector functional tests fail with messages l
 ## Debugging the SDK
 
 * Build the project using ```buildDebug.cmd``` 
-* If the build was successful, you'll find that it generated NuGet packages in <repository root>\..\bin\Debug\NuGet 
+* If the build was successful, you'll find that it generated NuGet packages in <repository root>\..\bin\Debug\NuGet
```
SergeyKanzhelev: Also update `CHANGELOG.md` with the short description of your change

Src/Common/InstrumentationKeyHashLookupHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            string hash;
+            var found = knownIKeyHashes.TryGetValue(instrumentationKey, out hash);
+
+            if (!found)
+            {
+                var instrumentationKeyLowerCase = instrumentationKey.ToLowerInvariant();
+                
+                // Instrumentation key is not case sensitive, let's now try for the lowercase - may be that's found.
+                found = knownIKeyHashes.TryGetValue(instrumentationKeyLowerCase, out hash);
```
SergeyKanzhelev: If you looking it up in dictionary - add it to the dictionary as well. Otherwise don't do another lookup

PR #192, opened 2016-10-28
Added messages to invalid operation exceptions, and tweaked one existing message.

dnduffy:


Src/WindowsServer/WindowsServer.Shared/Implementation/RoleInstance.cs
```diff
@@ -42,7 +42,7 @@ public Role Role
             {
                 if (this.TargetObject == null)
                 {
-                    throw new NotSupportedException();
+                    throw new NotSupportedException("Cannot get a role when no TargetType is set.");
```
OsvaldoRosado: Should this and the above be TargetObject? Should role be uppercased or Id lowercased?

PR #194, opened 2016-11-02
Add ratio gauge and the supported counters

kartang:


Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/CacheHelper.cs
```diff
@@ -85,7 +85,7 @@ public int GetCounterValue(string name, AzureWebApEnvironmentVariables environme
                     return 0;
                 }
 
-                CacheHelper.Instance.SaveToCache(name, uncachedJson, DateTimeOffset.Now.AddSeconds(5.0));
+                CacheHelper.Instance.SaveToCache(name, uncachedJson, DateTimeOffset.Now.AddMilliseconds(500));
```
dnduffy: Just curious why it is going from 5 seconds to half a second?

kartang: Quick pulse is actually requesting the values every second, so we have reduced the time to 500 milli seconds. Thought that 1 second might be too tight, as there could be a possibility of two requests happening consecutively with the timer timeouts.

Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/RatioCounterGauge.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        /// <summary>
+        /// Scale to measure the percentage or other increase the scaling of the ratio.
+        /// </summary>
+        private double scale;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RatioCounterGauge"/> class.
+        /// </summary>
+        /// <param name="name"> Name of the SumUpCountersGauge.</param>
```
dnduffy: I thought this was the RatioCounterGauge?

Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/RatioCounterGauge.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// <param name="numeratorGauge">The numerator for computing the ratio.</param>
+        /// <param name="denominatorGauge">The denominator for computing the ratio.</param>
+        /// <param name="scale">Scale to measure the percentage or other increase the scaling of the ratio.</param>
```
dnduffy: The grammar of this comment doesn't make sense.
But I love that there are comments! :)

Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/RatioCounterGauge.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// <summary>
+        /// Returns the current value of the sum of all different gauges attached to this one and resets their values.
+        /// </summary>
+        /// <returns>The value of the target metric.</returns>
+        public double GetValueAndReset()
+        {
+            if ((this.numeratorGauge != null) && (this.denominatorGauge != null))
+            {
+                double denominatorValue = this.denominatorGauge.GetValueAndReset();
+                return (denominatorValue == 0) ? 0 : (this.numeratorGauge.GetValueAndReset() / denominatorValue) * this.scale;
```
Dmitry-Matveev: Does it mean that numeratorGauge is not reset if denominatorGauge is 0? Is there some negative consequences to that?

PR #200, opened 2016-11-04
Fix functional tets

kartang:


Test/PerformanceCollector/FunctionalTests/FunctionalTests.csproj
```diff
@@ -120,7 +120,9 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="app.config" />
-    <EmbeddedResource Include="IisExpress\applicationhost.config" />
+    <EmbeddedResource Include="IisExpress\applicationhost.config">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
kartang: I had to make it to copy if newer, because tests fail becauseof that

PR #201, opened 2016-11-04
Edge case scenario handling for custom perf counter collection for azure web apps

kartang:


Src/PerformanceCollector/Shared/Implementation/PerformanceCounterUtility.cs
```diff
@@ -272,6 +272,12 @@ internal static IList<string> GetClrProcessInstances()
             }
 
             usesPlaceholder = true;
+
+            if (IsWebAppRunningInAzure())
+            {
+                return instanceName;
```
SergeyKanzhelev: styling: please try to avoid returning from the middle of the method. I think you need to check this condition first before any other conditions.

Src/PerformanceCollector/Shared/Implementation/PerformanceCounterUtility.cs
```diff
@@ -272,6 +272,12 @@ internal static IList<string> GetClrProcessInstances()
             }
 
             usesPlaceholder = true;
+
```
SergeyKanzhelev: why this method was not moved to "StandardPerfCollector" folder. It doesn't make sense to have it in the common folder

PR #219, opened 2016-11-18
Switch Core AI Sdk version to 2.3.0-beta1

vitalyf007:


Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
```diff
@@ -318,7 +318,7 @@ private void InitializeServiceClient(TelemetryConfiguration configuration)
         private static string GetInstanceName(TelemetryConfiguration configuration)
         {
             // we need to initialize an item to get instance information
-            var fakeItem = new MetricTelemetry();
```
tokaplan: Can you please check with chucknorris if this is exactly the same for all cases? That there are no checks in initializers that won't add the context we need to an Event?

tokaplan:
Reviewed the perf/qps part, left a comment, but looks good.

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
vitalyf007: We’re definitely safe for RoleInstance attribute. Initializers can, naturally, check the type of the item but Azure one does not

SergeyKanzhelev: It's ok. You can even create a new implementation of ITelemetry so it will not be sent to the server even if somehow will reach the channel.

vitalyf007: QPS just calls "Initialize" on it, not track. Should be fine...

PR #223, opened 2016-11-21
Dependency FuncTest -fail fast if initialization failed

cijothomas:
If the test app failed to deploy, its better to fail fast. Instead of hiding exception, throw it, and let tests fail fast.


Test/DependencyCollector/FunctionalTests/FuncTest/Helpers/TestWebApplication.cs
```diff
@@ -110,7 +110,8 @@ internal void Deploy(bool enableWin32Mode = false)
             }
             catch (Exception ex)
             {
-                Trace.TraceError("Exception occured while attempting to deploy application {0}: {1}.", this.AppName, ex);
+                Trace.TraceError("Exception occured while attempting to deploy application {0}: {1}. Tests wont continue as they are guaranteed to fail.", this.AppName, ex);
```
dnduffy: nitpick: "wont" should be either "will not" or "won't" (I prefer "will not" in this sentence).

Issue #226, opened 2016-11-22
2.2.0-beta4 does not distinguish Web App isntances


zakimaksyutov:
I have WebApp with 2 instances. Azure Portal Application Insights Servers Page shows only one.

cijothomas:
Reproed this in cijosdktest1 (AI application name).

cijothomas:
Newly added AzureWebAppRoleEnvironmentTelemetryInitializer is setting roleinstancename from the Azure ENV variable "WEBSITE_HOSTNAME", this will be same across instances, so all instances would be treated as same.
Need to Investigate why AzureRoleEnvironmentTelemetryInitializer did not initialize the roleinstancename-AzureWebAppRoleEnvironmentTelemetryInitializer wont initialize if already populated.

cijothomas:
AzureRoleEnvironmentTelemetryInitializer is valid only for Azure CloudServices. so AzureWebAppRoleEnvironmentTelemetryInitializer will populate roleinstance name to something not unique across instances.
The fix would be to populate the environment variable "WEBSITE_INSTANCE_ID", instead of "WEBSITE_HOSTNAME".
Ref:https://github.com/projectkudu/kudu/wiki/Azure-runtime-environment#website-environment-variables

cijothomas:
Commenting out WebAppRole initializer in AI.config (shown below) will be the workaround until the fixed sdk is published.

PR #229, opened 2016-11-23
AzureWebAppRoleEnvironmentTelemetryInitializer fix

cijothomas:
AzureWebAppRoleEnvironmentTelemetryInitializer populated roleinstanceid from env var WEBSITE_HOSTNAME, which is same acros instances, leading to AI backend not able to distinguish between instances.
This fix populates roleInstance name from WEBSITEINSTANCEID which is unique across instances.

Fixes the below bug:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/226

### DIFF
--- a/Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
+++ b/Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
@@ -16,8 +16,8 @@
         /// <summary>Azure Web App name corresponding to the resource name.</summary>
         private const string WebAppNameEnvironmentVariable = "WEBSITE_SITE_NAME";
 
-        /// <summary>Azure Web App host that contains site name and slot: site-slot.</summary>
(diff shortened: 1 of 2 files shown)


Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
```diff
@@ -16,8 +16,8 @@ public class AzureWebAppRoleEnvironmentTelemetryInitializer : ITelemetryInitiali
         /// <summary>Azure Web App name corresponding to the resource name.</summary>
         private const string WebAppNameEnvironmentVariable = "WEBSITE_SITE_NAME";
 
-        /// <summary>Azure Web App host that contains site name and slot: site-slot.</summary>
-        private const string WebAppHostNameEnvironmentVariable = "WEBSITE_HOSTNAME";
+        /// <summary>Azure Web App Instance Id representing the VM. Each instance will have different id.</summary>
```
dnduffy: ...Each instance will have **a** different id.

cijothomas:
@SergeyKanzhelev Can you also take a look?

Issue #225, opened 2016-11-22
2.2.0-beta4 stops reporting CPU for web apps


zakimaksyutov:
I have WebApp with 2 instances. First, one instance stopped reporting CPU. In some time both instances stopped reporting CPU.

SergeyKanzhelev:
I have suspicion that this bug can be related to https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/209. We have a report that sometimes exception happens when we are trying to get the counter value. This exception may cause the collection logic to mark counter as "bad". I remember we had this feature to mark counters as "bad" for the regular performance counters after the number of attempts.

Just a thought. To validate - I'd suggest to enable warning or even informational level traces for the diagnostics listener module.

cijothomas:
ok.. so i was able to REPRO this issue.. CPU simply stopped reporting after few minutes, whereas other metric (eg:PROCESS PRIVATE BYTES) are fine..

cijothomas:
Restarting the WebApp temporarily brings back the CPU Counters ...

cijothomas:
The root cause is found. Overflow of 'int' variable used to hold cpu time caused this. Will send pr with the fix.

cijothomas:
For AzureWebApp we calculate CPU% by reading CPUTime+KernelTime from AzureWebApp environment variables, and doing the delta between consecutive reads and dividing it by the time between the reads. 

The variable used to store the value read from AzureWebApp environment variable was an int – after few minutes of starting, the value returned would overflow int, and we hit exception.
Fix is trivial - replace int with double. (all counters are double anyways)

cijothomas:
Fixed here: https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/232
(it also fixed several other Exception handling logic as well)

PR #232, opened 2016-11-24
Fixed CPU Metric disappearing in AzureWebApp

cijothomas:
Fixed an issue which caused CPU metric to be missing in azurewebapp due to overflow of int variable used to read cpu times.

https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/225

### DIFF
--- a/Src/PerformanceCollector/Shared/Implementation/PerformanceCollectorEventSource.cs
+++ b/Src/PerformanceCollector/Shared/Implementation/PerformanceCollectorEventSource.cs
@@ -177,9 +177,9 @@
         }
 
         [Event(19, Level = EventLevel.Error, Message = @"Web App CPU Usage has unexpected (negative) value. Last Collected Value:{0} Previous Value:{1}. To avoid negative CPU percentage, this will be reported as zero instead.")]
-        public void WebAppCPUUsedNegativeValue(double lastCollectedValue, double previouslyCollectedValue, string applicationName = "dummy")
+        public void WebAppCPUUsedNegativeValue(string lastCollectedValue, string previouslyCollectedValue, string applicationName = "dummy")
         {
-            this.WriteEvent(19, lastCollectedValue.ToString(), previouslyCollectedValue.ToString(), this.ApplicationName);
+            this.WriteEvent(19, lastCollectedValue, previouslyCollectedValue, this.ApplicationName);
         }
 
         #endregion
--- a/Src/PerformanceCollector/Shared/Implementation/PerformanceCollectorEventSource.cs
+++ b/Src/PerformanceCollector/Shared/Implementation/PerformanceCollectorEventSource.cs
@@ -177,7 +177,7 @@
         }
 
         [Event(19, Level = EventLevel.Error, Message = @"CounterName:{2} has unexpected (negative) value. Last Collected Value:{0} Previous Value:{1}. To avoid negative value, this will be reported as zero instead.")]
-        public void WebAppCounterNegativeValue(string lastCollectedValue, string previouslyCollectedValue, string counterName, string applicationName = "dummy")
+        public void WebAppCounterNegativeValue(double lastCollectedValue, double previouslyCollectedValue, string counterName, string applicationName = "dummy")
         {
             this.WriteEvent(19, lastCollectedValue, previouslyCollectedValue, counterName, this.ApplicationName);
         }
--- a/Src/PerformanceCollector/Shared/Implementation/PerformanceCollectorEventSource.cs
+++ b/Src/PerformanceCollector/Shared/Implementation/PerformanceCollectorEventSource.cs
@@ -176,10 +176,10 @@
             this.WriteEvent(18, e, counter, this.ApplicationName);
         }
 
-        [Event(19, Level = EventLevel.Error, Message = @"Web App CPU Usage has unexpected (negative) value. Last Collected Value:{0} Previous Value:{1}. To avoid negative CPU percentage, this will be reported as zero instead.")]
-        public void WebAppCPUUsedNegativeValue(string lastCollectedValue, string previouslyCollectedValue, string applicationName = "dummy")
+        [Event(19, Level = EventLevel.Error, Message = @"CounterName:{2} has unexpected (negative) value. Last Collected Value:{0} Previous Value:{1}. To avoid negative value, this will be reported as zero instead.")]
+        public void WebAppCounterNegativeValue(string lastCollectedValue, string previouslyCollectedValue, string counterName, string applicationName = "dummy")
         {
-            this.WriteEvent(19, lastCollectedValue, previouslyCollectedValue, this.ApplicationName);
+            this.WriteEvent(19, lastCollectedValue, previouslyCollectedValue, counterName, this.ApplicationName);
         }
 
         #endregion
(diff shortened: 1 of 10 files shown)


dnduffy:
I don't have complete context so please forgive me if these are naive questions.
What are the downstream effects of this change?
Will we send telemetry data with a value like "1.00000001" or "0.999999999" instead of "1" now?
Are there other counters we are getting that don't make sense to represent as double?
Why not use long instead of double?
Again I apologize I don't have the full context but the last time I worked with counters I recall that they can be of different types so picking just one type to represent all possible counter values doesn't seem appropriate to me even if double is the appropriate type for the counter in question where the overflow occurred in the code.

cijothomas:
@dnduffy  - No downstream effects. Double is used everywhere in perfcounter module for storing the counter values. The int->double change I made was in a helper file which is parsing counter value from a json string. Infact the method invoking this helper method is expecting double.
There is no change in the way we send data.
(I don't see any reason why long would not work here. I think it would be better to use unsigned long here as the value we are parsing is the cpu/kernel time, which is never -ve).
Will wait if others have any other points to suggest about using unsigned long.

cijothomas:
Cant really use unsigned long, as there could be counters which need double.

cijothomas:
@SergeyKanzhelev replaced double with long as we only need long. And we are unlikely to overflow long as it can store months.

Src/PerformanceCollector/Shared/Implementation/PerformanceCollectorEventSource.cs
```diff
@@ -176,6 +176,12 @@ public void WebAppCounterRegistrationFailedEvent(string e, string counter, strin
             this.WriteEvent(18, e, counter, this.ApplicationName);
         }
 
+        [Event(19, Level = EventLevel.Error, Message = @"Web App CPU Usage has unexpected (negative) value. Last Collected Value:{0} Previous Value:{1}. To avoid negative CPU percentage, this will be reported as zero instead.")]
```
SergeyKanzhelev: @dnduffy - please advice on wording. This message doesn't sound ok

Src/PerformanceCollector/Shared/Implementation/PerformanceCollectorEventSource.cs
```diff
@@ -176,6 +176,12 @@ public void WebAppCounterRegistrationFailedEvent(string e, string counter, strin
...
+        public void WebAppCPUUsedNegativeValue(double lastCollectedValue, double previouslyCollectedValue, string applicationName = "dummy")
+        {
+            this.WriteEvent(19, lastCollectedValue.ToString(), previouslyCollectedValue.ToString(), this.ApplicationName);
```
SergeyKanzhelev: EventSource shouldn't require `ToString()` on doubles.

SergeyKanzhelev: Also you switched to long. Switch it here as well

Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/CPUPercenageGauge.cs
```diff
@@ -48,7 +48,18 @@ public double GetValueAndReset()
                 var baseValue = this.lastCollectedTime.Ticks - previouslyCollectedTime.Ticks;
                 baseValue = baseValue != 0 ? baseValue : 1;
 
-                value = (float)((this.lastCollectedValue - previouslyCollectedValue) / baseValue * 100.0);
+                var diff = this.lastCollectedValue - previouslyCollectedValue;
+
+                if (diff < 0)
```
SergeyKanzhelev: Is it applicable for other gauges?

Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/CacheHelper.cs
```diff
@@ -57,9 +57,9 @@ public int PerformanceCounterValue(string performanceCounterName, string json)
                 startingIndex++;
             }
 
-            if (!int.TryParse(valueString.ToString(), out value))
+            if (!long.TryParse(valueString.ToString(), out value))
             {
-                throw new System.InvalidCastException("The value of the counter cannot be converted to integer type.");
+                throw new System.InvalidCastException("The value of the counter cannot be converted to long type. Value:" + valueString);
```
SergeyKanzhelev: so what will happen here? Counter will be marked as bad and wouldn't be reported for 5 minutes? Or something else will happen?

Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/CPUPercenageGauge.cs
cijothomas: No other gauges are doing a delta calculation,

SergeyKanzhelev: We have rate counters https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/RateCounterGauge.cs#L100 It will not be logical to get negative number of requests/sec:

https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/75373e57dcf8d646c54ee188461c373f2cc98939/Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/CounterFactory.cs#L31-L35

Src/PerformanceCollector/Shared/Implementation/PerformanceCollectorEventSource.cs
cijothomas: Agreed.

cijothomas:
Yea. thanks for pointing out. I will do similar tracing for RateCounters as well.

Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/RateCounterGauge.cs
```diff
@@ -69,39 +73,31 @@ internal RateCounterGauge(string name, string jsonId, AzureWebApEnvironmentVaria
...
+                    PerformanceCollectorEventSource.Log.WebAppCounterNegativeValue(
+                    this.lastValue.ToString(CultureInfo.InvariantCulture),
+                    previouslyCollectedValue.ToString(CultureInfo.InvariantCulture),
+                    this.name);
                 }
                 else
-                {
-                    this.lastValue = this.counter.GetValueAndReset();
+                {                    
+                    value = timeDifferenceInSeconds != 0 ? (double)(diff * 100.0 / timeDifferenceInSeconds) : 0;
```
SergeyKanzhelev: no need to multiple by 100.0

Src/PerformanceCollector/Shared/Implementation/PerformanceCollectorEventSource.cs
```diff
@@ -176,6 +176,12 @@ public void WebAppCounterRegistrationFailedEvent(string e, string counter, strin
             this.WriteEvent(18, e, counter, this.ApplicationName);
         }
 
+        [Event(19, Level = EventLevel.Error, Message = @"CounterName:{2} has unexpected (negative) value. Last Collected Value:{0} Previous Value:{1}. To avoid negative value, this will be reported as zero instead.")]
+        public void WebAppCounterNegativeValue(string lastCollectedValue, string previouslyCollectedValue, string counterName, string applicationName = "dummy")
```
SergeyKanzhelev: use `long` instead of `string`

PR #233, opened 2016-11-25
Fix Dependency Functional test for sql

cijothomas:
Tests at the end fail occasionaly as the time taken to access local sql db was ~1msec or less, leading to sdk reporting it as zero. This fix modifies the sql query to return ~20 rows instead of just a couple rows. In the last 7 test runs, this took more than 1 msec and tests are green.

Updated: Introduced a 1msec delay using WAITFOR sql command. This should be good enough.


Test/DependencyCollector/FunctionalTests/TestApps/Aspx451/ExternalCalls.aspx.cs
```diff
@@ -22,7 +22,7 @@ public partial class ExternalCalls : System.Web.UI.Page
         /// <summary>
         /// Valid SQL Query.
         /// </summary> 
-        private const string ValidSqlQueryToApmDatabase = "select top 2 * from dbo.Messages";
+        private const string ValidSqlQueryToApmDatabase = "WAITFOR DELAY '00:00:00:001'; select * from dbo.Messages";
```
dnduffy: So we're still racing the race condition?

cijothomas: There is no race condition here.
The product cannot measure dependency below 1 msec, and the actual time as <1msec. This delay will ensure a minimum 1msec dependency duration.

PR #234, opened 2016-11-28
Adjusted WaitTime for SQL query to be 6msec

cijothomas:
Adjusted WaitTime for SQL query to be 6msec, as sql is not accurate below 3 msec.

### DIFF
--- a/Test/DependencyCollector/FunctionalTests/TestApps/Aspx451/ExternalCalls.aspx.cs
+++ b/Test/DependencyCollector/FunctionalTests/TestApps/Aspx451/ExternalCalls.aspx.cs
@@ -20,9 +20,9 @@
         private const string ConnectionString = @"Data Source=.\SQLEXPRESS;Initial Catalog=RDDTestDatabase;Integrated Security=True";
 
         /// <summary>
-        /// Valid SQL Query.
+        /// Valid SQL Query. The wait for delay of 6msec is used to prevent access time of less than 1msec. SQL is not accurate below 3, so used 6 msec delay.
         /// </summary> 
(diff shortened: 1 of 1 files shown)


Test/DependencyCollector/FunctionalTests/TestApps/Aspx451/ExternalCalls.aspx.cs
```diff
@@ -20,9 +20,9 @@ public partial class ExternalCalls : System.Web.UI.Page
         private const string ConnectionString = @"Data Source=.\SQLEXPRESS;Initial Catalog=RDDTestDatabase;Integrated Security=True";
 
         /// <summary>
-        /// Valid SQL Query.
+        /// Valid SQL Query. The wait for delay of 6msec is used to prevent access time of less than 1msec. SQL is not accurate below 3, so used 6 msec delay.
         /// </summary> 
-        private const string ValidSqlQueryToApmDatabase = "WAITFOR DELAY '00:00:00:001'; select * from dbo.Messages";
+        private const string ValidSqlQueryToApmDatabase = "WAITFOR DELAY '00:00:00:006'; select * from dbo.Messages";
```
dnduffy: This is what I'm talking about when I say it is a race condition.
You are chasing a timing value that "happens to work" currently and this is the third time changing it.
What about changing the criteria to not fail super low values?
Why is "zero" a problem?

cijothomas: We need a access time greater than minimum measurable value supported by product (which is 1 msec) And SQL Server is not accurate below 3 msec, so changed to 6msec.

If the test criteria is changed to accept zero, then we might fail to catch genuine bugs. These tests are counting on bing.com/local sql server dependencies, and the best we can do is to make sure the access time is between a reasonable range.

PR #235, opened 2016-11-29
Fix SQL delay in all the queries/stored proce

cijothomas:
Continuation of this:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/234

Added SQL delay in all the queries. Above PR modified only 1 query.

### DIFF
--- a/Test/DependencyCollector/FunctionalTests/FuncTest/IIS/IisApplicationPool.cs
+++ b/Test/DependencyCollector/FunctionalTests/FuncTest/IIS/IisApplicationPool.cs
@@ -96,6 +96,7 @@ namespace FuncTest.IIS
             catch (Exception ex)
             {
                 Trace.TraceError("Exception occured while creating IIS Application Pool: {0}. The exception thrown was {1}", poolName, ex.Message);
+                throw ex;
             }
         }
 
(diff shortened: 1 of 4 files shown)


Test/DependencyCollector/FunctionalTests/FuncTest/IIS/IisApplicationPool.cs
```diff
@@ -96,6 +96,7 @@ public IisApplicationPool(string poolName, string managedRuntimeVersion = Defaul
             catch (Exception ex)
             {
                 Trace.TraceError("Exception occured while creating IIS Application Pool: {0}. The exception thrown was {1}", poolName, ex.Message);
+                throw ex;
```
cijothomas: This was meant to be a diff PR as this was unrelated. but i did not change branches while making this change.

cijothomas: btw - this is simply a fix to makes tests fail instantly if initialize has issues

dnduffy: Just use "throw;" so you don't lose the stack trace.

cijothomas: will modify.

PR #239, opened 2016-11-29
Removed second count parameter that was getting set to 0.

dnduffy:
Removed second count parameter that was getting set to 0.
Checking in updated CHANGELOG.md that doesn't seem to have been committed previously.

### DIFF
--- a/Src/PerformanceCollector/Shared/Implementation/PerformanceCollectorEventSource.cs
+++ b/Src/PerformanceCollector/Shared/Implementation/PerformanceCollectorEventSource.cs
@@ -71,15 +71,14 @@
         }
 
         [Event(8, Keywords = Keywords.UserActionable, Level = EventLevel.Error,
-            Message = @"Error collecting {0} out of {1} configured performance counters. Please check the configuration.
+            Message = @"Error collecting {0} of the configured performance counters. Please check the configuration.
 {2}")]
         public void CounterCheckConfigurationEvent(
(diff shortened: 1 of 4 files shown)


Src/PerformanceCollector/Shared/Implementation/PerformanceCollectorEventSource.cs
```diff
@@ -71,15 +71,14 @@ public void CounterParsingFailedEvent(string e, string counter, string applicati
...
+            Message = @"Error collecting {0} of the configured performance counters. Please check the configuration.
 {2}")]
         public void CounterCheckConfigurationEvent(
             string misconfiguredCountersCount,
-            string overallConfiguredCountersCount,
             string e,
             string applicationName = "dummy")
         {
-            this.WriteEvent(8, misconfiguredCountersCount, overallConfiguredCountersCount, e, this.ApplicationName);
+            this.WriteEvent(8, misconfiguredCountersCount, e, this.ApplicationName);
```
cijothomas: why remove the overall counter count instead of fixing it to show the actual value?
(adding default+configured should give us the total count right? )

cijothomas: i saw your comment in the issue. yea lets do the quick fix now, and keep the issue open to investigate and fix in 2.3beta1

PR #251, opened 2016-12-07
Nizarq

nizarq:
A few notes: 
- For the .net storage client, I could not reproduce any failures when the headers were placed, as described in https://github.com/Microsoft/ApplicationInsights-node.js/issues/143.
- Still I went ahead and implemented this change. 
- I contemplated adding default values in the Excluded domains - for example - the storage endpoints for table, blob, queue etc. But the host names are of the format http://<storage account>.table.core.windows.net/. Supporting such defaults would involve supporting wild cards in the provided domain names, such as http://*.table.core.windows.net - which is significantly more complicated than comparing against host names. Hence, no default excluded domain logic is added for now.
- This changeset already goes beyond the triaged minimum requirement of providing a config to enable/disable the collection. It does provide an exclusion list that the user can provide. So for example, if a user wanted to disable adding headers to multiple storage account endpoint - (s)he can just enumerate all endpoints in the config.


Src/DependencyCollector/Shared/ComponentCorrelation.cs
```diff
@@ -0,0 +1,28 @@
+﻿namespace Microsoft.ApplicationInsights.DependencyCollector
+{
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Component Correlation configuration.
+    /// </summary>
+    public class ComponentCorrelation
```
SergeyKanzhelev: I really do not like another public class in this namespace. Also name is not informative. `CrossComponentCorrelationSettings` may sound better. 

I also do not like two classes with the same name in different namespaces.

I'd rather have those properties directly in the module itself.

Src/DependencyCollector/Shared/ComponentCorrelation.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        private ICollection<string> excludedDomains = new SanitizedHostList();
+
+        /// <summary>
+        /// Gets or sets a value indicating whether component correlation collection is enabled. That is should there be an attempt to add component correlation related headers to outgoing responses.
+        /// </summary>
+        public bool CollectionEnabled { get; set; } = true;
```
SergeyKanzhelev: In this case it's not collection - it is propagation. Call it `SetCorrelationHttpHeaders` or something like this

Src/DependencyCollector/Shared/ComponentCorrelation.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        /// <summary>
+        /// Gets the list object that is meant to hold domains to exclude.
+        /// </summary>
+        public ICollection<string> ExcludedDomains
```
SergeyKanzhelev: Will `IEnumerable` be enough here? Why do you need `ICollection`?

Src/DependencyCollector/Shared/SanitizedHostList.cs
```diff
@@ -0,0 +1,104 @@
...
+                return;
+            }
+
+            Uri uriResult;
+            if (Uri.TryCreate(item, UriKind.Absolute, out uriResult))
+            {
+                this.AddIfNotExist(uriResult.Host);
+                return;
+            }
+            else if (Uri.TryCreate("http://" + item, UriKind.Absolute, out uriResult))
```
SergeyKanzhelev: I do not understand why you need this complicated logic. Can we just have a list of strings and compare the host by `string.Contains` instead?

Src/Web/Web.Shared.Net/ComponentCorrelation.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Microsoft.ApplicationInsights.Web
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+
+    /// <summary>
+    /// Component Correlation configuration.
+    /// </summary>
+    public class ComponentCorrelation
```
SergeyKanzhelev: same comment as above.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -118,7 +130,9 @@ public void AddTargetHashForResponseHeader(HttpContext context)
                 this.telemetryClient.Initialize(requestTelemetry);
             }
 
-            if (!string.IsNullOrEmpty(requestTelemetry.Context.InstrumentationKey) && context.Response.Headers[RequestResponseHeaders.TargetInstrumentationKeyHeader] == null)
+            if (ComponentCorrelation.CollectionEnabled
```
SergeyKanzhelev: This flag should disable subscribing to the callback completely, not just setting the header.

Src/DependencyCollector/Shared/SanitizedHostList.cs
nizarq: this is in order to let the user work with http://www.bbc.com and www.bbc.com and http://www.bbc.com/sports - All of these will result in the host getting added.

Src/DependencyCollector/Shared/ComponentCorrelation.cs
nizarq: The class is public because the property that matches the config is public. 
Also it has to be it's own class rather than a property because of the way we decided to form the config: 
```<Add Type="Microsoft.ApplicationInsights.DependencyCollector.DependencyTrackingTelemetryModule, Microsoft.AI.DependencyCollector">
      <ComponentCorrelation>
        <CollectionEnabled>true</CollectionEnabled>
        <ExcludedDomains>
          <Add>www.bbc.com</Add>
          <Add>http://www.msn.com.cn/news</Add>
        </ExcludedDomains>
      </ComponentCorrelation>
    </Add>
```
Since the logic that processes the config is out of our control - in order to parse this, we have to define ComponentCorrelation as a property which in turn contains CollectionEnabled and ExcludedDomains.

Are you saying you would rather have something like the following - which would allow these to become properties on the modules
```
    <Add Type="Microsoft.ApplicationInsights.DependencyCollector.DependencyTrackingTelemetryModule, Microsoft.AI.DependencyCollector">
      <ComponentCorrelationCollectionEnabled>true</ComponentCorrelationCollectionEnabled>
      <ComponentCorrelationExcludedDomains>
        <Add>www.bbc.com</Add>
        <Add>http://www.msn.com.cn/news</Add>
      </ComponentCorrelationExcludedDomains>
    </Add>
```

Src/DependencyCollector/Shared/ComponentCorrelation.cs
nizarq: Not enough because Add needs to be called on this.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
nizarq: makes sense

Src/DependencyCollector/Shared/SanitizedHostList.cs
SergeyKanzhelev: are there unit tests for this collection so I understand what I'm missing here. I thought simply include `bbc.com` will disable it all, `bbc.com/sports` will disable sports part of it. Simple substring

PR #255, opened 2016-12-07
Exceptions statistcs implementation for 2.3.0-beta1

SergeyKanzhelev:


Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -0,0 +1,188 @@
...
+
+                WindowsServerEventSource.Log.FirstChanceExceptionCallbackCalled();
+
+                var type = exception.GetType().FullName;
+
+                // obtaining the operation name. At this stage we have no intention to send this telemetry item
+                ExceptionTelemetry fakeTelemetry = new ExceptionTelemetry(exception);
+                this.telemetryClient.Initialize(fakeTelemetry);
+
+                var operation = fakeTelemetry.Context.Operation.Name;
```
Dmitry-Matveev: Can we instead get that operation from the Request Telemetry in HttpContext via our extension of HttpContext? Then, if it's not available (not classic ASP.NET) - check for operation in the storage we have for OperationsAPI? If still nothing, do not assign anything to operation... Or do we expect that there is a third way to set the operation and this fakeTelemetry item will help us to get to that better?

Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -0,0 +1,188 @@
...
+            if (disposing)
+            {
+                this.unregisterAction(this.CalculateStatistics);
+                this.metricManager.Dispose();
+            }
+        }
+
+        private void CalculateStatistics(object sender, FirstChanceExceptionEventArgs firstChanceExceptionArgs)
+        {
+            if (executionSyncObject == LOCKED)
```
Dmitry-Matveev: Is it just a protection from StackOverflow, so it's guaranteed that we already passed this method fully once (up until exception in our code) and won't just pass again?

Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -0,0 +1,188 @@
...
+            }
+
+            var metric = this.metricManager.CreateMetric("Exceptions Thrown", dimensions);
+
+            metric.Track(1);
+        }
+
+        private string GetDimCappedString(string value, ConcurrentBag<string> capValues)
+        {
+            if (capValues.Count > 100)
```
Dmitry-Matveev: We might want to expose this setting unless it's a hard limitation in the processing pipeline.

Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
SergeyKanzhelev: variable it thread static. So shouldn't be a problem

Src/WindowsServer/WindowsServer.Shared.Tests/FirstChanceExceptionStatisticsTelemetryModuleTest.cs
```diff
@@ -0,0 +1,222 @@
...
+                h => handler = h,
+                _ => { }))
+            {
+                module.Initialize(this.configuraiton);
+                handler.Invoke(null, new FirstChanceExceptionEventArgs(null));
+            }
+        }
+
+        [TestMethod]
+        public void FirstChanceExceptionStatisticsTelemetryModuleRecievesFirstChance()
```
SergeyKanzhelev: Forgot to delete. thx

Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
SergeyKanzhelev: WCF has it's own way, we have couple of third party Web API implementations. I understand the concern, but ultimately I want this fake telemetry to became not as fake in future and collect most of those exceptions

Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -0,0 +1,188 @@
...
+
+                // obtaining the operation name. At this stage we have no intention to send this telemetry item
+                ExceptionTelemetry fakeTelemetry = new ExceptionTelemetry(exception);
+                this.telemetryClient.Initialize(fakeTelemetry);
+
+                var operation = fakeTelemetry.Context.Operation.Name;
+
+                // obtaining failing method name by walking 1 frame up the stack
+                var failingMethod = new StackFrame(1).GetMethod();
+                var method = failingMethod.DeclaringType.FullName + "." + failingMethod.Name;
```
SergeyKanzhelev: It is more like global functions that CLR allows. Thanks

PR #256, opened 2016-12-07
transaction correlation

SergeyKanzhelev:
CC: @AlexBulankou. @nizarq

### DIFF
--- a/Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
+++ b/Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
@@ -276,6 +276,20 @@
                 {
                     webRequest.Headers.Add(RequestResponseHeaders.SourceInstrumentationKeyHeader, InstrumentationKeyHashLookupHelper.GetInstrumentationKeyHash(telemetry.Context.InstrumentationKey));
                 }
+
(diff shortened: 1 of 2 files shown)


Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
```diff
@@ -276,6 +276,20 @@ private object OnBegin(object thisObj, bool isAsyncCall)
                 {
                     webRequest.Headers.Add(RequestResponseHeaders.SourceInstrumentationKeyHeader, InstrumentationKeyHashLookupHelper.GetInstrumentationKeyHash(telemetry.Context.InstrumentationKey));
                 }
+
+                // Add the root ID
+                var rootId = telemetry.Context.Operation.Id;
+                if (!string.IsNullOrEmpty(rootId) && webRequest.Headers[RequestResponseHeaders.StandardRootIdHeader] == null)
```
nizarq: should these header additions also be override-able by some config?

SergeyKanzhelev: Created https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/257

PR #269, opened 2016-12-13
Top processes by CPU for QuickPulse

tokaplan:
This feature allows QuickPulseTelemetryModule to collect a list of top CPU consuming processes on the machine and send it along with the rest of the data for QPS to process and UX to display.


Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/NativeMethods.cs
```diff
@@ -0,0 +1,191 @@
...
+            public short wYear;
+            public short wMonth;
+            public short wDayOfWeek;
+            public short wDay;
+            public short wHour;
+            public short wMinute;
+            public short wSecond;
+            public short wMilliseconds;
+
+            public override string ToString()
```
SergeyKanzhelev: do we need it? How often will it be called. Implementation is far from ideal

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
```diff
@@ -128,9 +143,10 @@ public void Initialize(TelemetryConfiguration configuration)
                         QuickPulseEventSource.Log.ModuleIsBeingInitializedEvent(
```
SergeyKanzhelev: should this call be in the end of initialize just before the `catch` statement?

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
```diff
@@ -128,9 +143,10 @@ public void Initialize(TelemetryConfiguration configuration)
...
                             string.Format(
```
SergeyKanzhelev: The whole purpose of event source is to avoid string formatting. Please move arguments to the eventsource message

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/CategorySample.cs
```diff
@@ -0,0 +1,258 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.Implementation.QuickPulse.PerfLib
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Globalization;
+    using System.Runtime.InteropServices;
+
+    internal class CategorySample
+    {
+        public Dictionary<int, CounterDefinitionSample> CounterTable;
```
tokaplan: It's assigned in CategorySampleMock's constructor as well, which is forbidden if it's readonly.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/CategorySample.cs
```diff
@@ -0,0 +1,258 @@
...
+                int numPerfObjects = dataBlock.NumObjectTypes;
+                if (numPerfObjects == 0)
+                {
+                    this.CounterTable = new Dictionary<int, CounterDefinitionSample>();
+                    this.InstanceNameTable = new Dictionary<string, int>(StringComparer.OrdinalIgnoreCase);
+                    return;
+                }
+
+                // Need to find the right category, GetPerformanceData might return several of them.
+                NativeMethods.PERF_OBJECT_TYPE perfObject = new NativeMethods.PERF_OBJECT_TYPE();
```
tokaplan: We're using the "pseudo-registry" perf interface, which is used internally by both PDH and CLR's PerformanceCounter. The minimum amount of perf information that can be read is a single performance object (aka category) - in our case "Process". It does include all counters, but only within the object; the same thing happens when you think you read a single performance counter in CLR, for instance. No custom perf provider will have any role to play here, since it's all just the "Process" object.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/CategorySample.cs
```diff
@@ -0,0 +1,258 @@
...
+                    throw new InvalidOperationException("Category not found for categoryNameIndex " + categoryNameIndex);
+                }
+
+                int counterNumber = perfObject.NumCounters;
+                int instanceNumber = perfObject.NumInstances;
+
+                bool isMultiInstance = instanceNumber != -1;
+
+                // Move pointer forward to end of PERF_OBJECT_TYPE
+                dataRef = (IntPtr)((long)dataRef + perfObject.HeaderLength);
```
tokaplan: See comment above about why this entire folder is not unit tested.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/CategorySample.cs
```diff
@@ -0,0 +1,258 @@
...
+                }
+
+                // now set up the InstanceNameTable.  
+                if (!isMultiInstance)
+                {
+                    throw new InvalidOperationException("Single instance categories are not supported");
+                }
+
+                string[] parentInstanceNames = null;
+                this.InstanceNameTable = new Dictionary<string, int>(instanceNumber, StringComparer.OrdinalIgnoreCase);
```
tokaplan: Yes, I have tested it. With PIDs enabled, the names are not unique anymore, so we'll get the process_pid format, and won't add any #'s here.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/CategorySample.cs
```diff
@@ -0,0 +1,258 @@
...
+                    {
+                        instanceName = parentInstanceNames[perfInstance.ParentObjectInstance] + "/"
+                                       + Marshal.PtrToStringUni((IntPtr)((long)dataRef + perfInstance.NameOffset));
+                    }
+                    else
+                    {
+                        instanceName = Marshal.PtrToStringUni((IntPtr)((long)dataRef + perfInstance.NameOffset));
+                    }
+
+                    // in some cases instance names are not unique (Process), same as perfmon, so generate a unique name
```
tokaplan: CLR implementation, I'm not touching it :)

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/NativeMethods.cs
tokaplan: This is from CLR's NativeMethods.cs. I've cut the file down to only the structures/flags that we need, but I'd rather not make any changes unless necessary.

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
tokaplan: Well this is "IsBeingInitialized", so it's here at the beginning. The purpose of this event is to know that the module is at least attempting to initialize and what the initialization parameters are.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/IQuickPulseProcessProvider.cs
```diff
@@ -0,0 +1,17 @@
...
+    /// <summary>
+    /// Provider interface for Windows processes.
+    /// </summary>
+    internal interface IQuickPulseProcessProvider
+    {
+        void Initialize();
+
+        void Close();
+
+        IEnumerable<QuickPulseProcess> GetProcesses(out TimeSpan? totalTime);
```
tokaplan: I've provided the doc summary. The name "totalTime" refers to the value of the _Total instance of the processor time counter, not sure how to make it much clearer.

tokaplan:
Sergey, I've responded to your comments, please take a look when you get a chance. Let me know if you'd like to chat in person about why we're reusing all that CLR stuff as-is.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/CategorySample.cs
```diff
@@ -0,0 +1,258 @@
...
+
+    internal class CategorySample
+    {
+        public Dictionary<int, CounterDefinitionSample> CounterTable;
+
+        public Dictionary<string, int> InstanceNameTable;
+
+        private readonly PerfLib library;
+
+        public CategorySample(byte[] data, int categoryNameIndex, int counterNameIndex, PerfLib library)
```
tokaplan: This is not our code, it's copied from CLR's PerformanceCounter implementation. It does certain amount of black magic regarding parsing a byte stream with complicated structure and cross-indexed fields. Unit testing this would defeat the purpose, since constructing the byte stream would be much more error-prone than the imiplementation itself. PerformanceCounter has been around for a while, so I'm taking it as is.

SergeyKanzhelev: Add a file-level comment section that explains which file it was copied from and what version of FW you copied it from down to the file version of the file you copied it from.

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
SergeyKanzhelev: Yes, please

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
SergeyKanzhelev: ok. makes sense. Error will be handled and traced by infrastracture

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/IQuickPulsePerfLib.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.Implementation.QuickPulse
+{
+    using Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.Implementation.QuickPulse.PerfLib;
+
+    /// <summary>
+    /// Interface for the <c>PerfLib</c>.
+    /// </summary>
+    internal interface IQuickPulsePerfLib
```
SergeyKanzhelev: Instead of custom interfaces - I'd suggest to change the design:
1. Use the same method to read counters for both - standard counters and processes
2. Support '*' instance name in that interface
3. Extract top 5 and total on "query" side when creating the message to quick pulse.

This implementation looks like a band aid.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/QuickPulseTopCpuCollector.cs
```diff
@@ -0,0 +1,154 @@
...
+                    this.AccessDenied = true;
+                }
+
+                return Enumerable.Empty<Tuple<string, int>>();
+            }
+        }
+
+        public void Initialize()
+        {
+            this.InitializationFailed = false;
```
SergeyKanzhelev: this should check whether it's running on Azure Web Apps or not.

SergeyKanzhelev:
Any other reviewers? @tokaplan - do you have peer review in live metrics team?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/IQuickPulseProcessProvider.cs
dnduffy: The doc comments are not visible (yet), please push the update.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/CategorySample.cs
tokaplan: Per offline email thread - we decided we own this code as if written from scratch. Added unit tests.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/CategorySample.cs
tokaplan: Added unit tests.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/IQuickPulseProcessProvider.cs
tokaplan: Fixed.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/CategorySample.cs
dnduffy: Not if you use inheritance.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/CategorySample.cs
```diff
@@ -0,0 +1,258 @@
...
+                        }
+                    }
+
+                    dataRef = (IntPtr)((long)dataRef + perfInstance.ByteLength);
+
+                    // we only need one counter
+                    ////foreach (CounterDefinitionSample s in samples)
+                    ////{
+                    ////    s.SetInstanceValue(i, dataRef);
+                    ////}
```
dnduffy: nitpick: should probably delete the dead code.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/IQuickPulseTopCpuCollector.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// <summary>
+    /// Interface for top CPU collector.
+    /// </summary>
+    internal interface IQuickPulseTopCpuCollector
+    {
+        bool InitializationFailed { get; }
+
+        bool AccessDenied { get; }
+
+        IEnumerable<Tuple<string, int>> GetTopProcessesByCpu(int topN);
```
dnduffy: I've noticed there aren't a lot of doc comments in this code, it's already quite a long code review and the lack of comments makes it even harder for me to review.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/NativeMethods.cs
```diff
@@ -0,0 +1,191 @@
...
+
+            // this one is kind of weird. It is defined as in SDK:
+            // #ifdef _WIN64
+            //  DWORD           CounterNameTitle;
+            // #else
+            //  LPWSTR          CounterNameTitle;
+            // #endif
+            // so we can't use IntPtr here.
+
+            public int CounterNameTitlePtr = 0;
```
dnduffy: Just curious, should we be explicit with Int32 vs. Int64 instead of just int?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/PerfLib.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        private PerformanceMonitor performanceMonitor;
+
+        private PerfLib()
+        {
+        }
+
+        public static PerfLib GetPerfLib()
+        {
+            library = library ?? new PerfLib();
```
dnduffy: Rather than checking and assigning to self each time you should just pre-create it.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/PerfLib.cs
```diff
@@ -0,0 +1,51 @@
...
+        public void Initialize()
+        {
+            this.performanceMonitor = new PerformanceMonitor();
+        }
+
+        public void Close()
+        {
+            this.performanceMonitor?.Close();
+
+            library = null;
```
dnduffy: So it looks like Dispose, but it isn't, and the library can be recreated in GetPerfLib, I realize you copied this from the CLR but it's definitely not the usual pattern.  This also doesn't look thread safe if closing the performanceMonitor will cause issues for another thread trying to still use the same instance.

dnduffy: I just saw the PerformanceMonitor code below and this is definitely not thread safe.  At a minimum there would need to be more null checks but I would prefer if the design pattern can be improved upon as well.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/PerformanceMonitor.cs
```diff
@@ -0,0 +1,109 @@
...
+
+                            break;
+
+                        default:
+                            throw SharedUtils.CreateSafeWin32Exception(error);
+                    }
+                }
+                catch (InvalidCastException e)
+                {
+                    throw new InvalidOperationException("Counter data is corrupt " + this.perfDataKey, e);
```
dnduffy: Hmm, I wonder if there is a clearer exception type to use, yes the system is in an invalid state but it wasn't really the operation of GetData that was invalid nor is it likely the user's fault so InvalidOperationException seems slightly misleading.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/PerformanceMonitor.cs
```diff
@@ -0,0 +1,109 @@
...
+        [ResourceConsumption(ResourceScope.Machine, ResourceScope.Machine)]
+        private void Init()
+        {
+            try
+            {
+                this.perfDataKey = Registry.PerformanceData;
+            }
+            catch (UnauthorizedAccessException e)
+            {
+                throw new UnauthorizedAccessException("Access denied opening the performance key", e);
```
dnduffy: Why wrap the UAE with another UAE?  Wrapping and rethrowing should provide additional value which it isn't clear if that is the case here.  What is the original UAE message?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/QuickPulseProcessProvider.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// <summary>
+    /// Top CPU collector.
+    /// </summary>
+    internal sealed class QuickPulseProcessProvider : IQuickPulseProcessProvider
+    {
+        private readonly IQuickPulsePerfLib perfLib;
+
+        public QuickPulseProcessProvider(IQuickPulsePerfLib perfLib)
+        {
+            this.perfLib = perfLib ?? PerfLib.PerfLib.GetPerfLib();
```
dnduffy: PerfLib.PerfLib.GetPerfLib?

dnduffy: Why not have an empty constructor which passes PerfLib.GetPerfLib() to this constructor and have this one check for null and throw ArgumentNullException?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/QuickPulseProcessProvider.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        public void Initialize()
+        {
+            this.perfLib.Initialize();
```
dnduffy: What is the purpose of delayed initialization if you are guaranteed to have an instance set for this.perfLib in the constructor?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/QuickPulseProcessProvider.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// <summary>
+        /// Gets a collection of <see cref="QuickPulseProcess"/> objects - each corresponding to a system process and containing
+        /// information about the amount of time the process has occupied CPU cores.
+        /// </summary>
+        /// <param name="totalTime">If available, contains the value of the _Total instance of the counter, which indicates the overall
+        /// amount of time spent by CPU cores executing system processes.</param>
+        public IEnumerable<QuickPulseProcess> GetProcesses(out TimeSpan? totalTime)
+        {
+            const int CategoryIndex = 230;
+            const int CounterIndex = 6;
```
dnduffy: Why 6, why 230?  Please doc comment your magic numbers so people will know what they mean in future.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/QuickPulseTopCpuCollector.cs
```diff
@@ -0,0 +1,154 @@
...
+            this.timeProvider = timeProvider;
+            this.processProvider = processProvider;
+
+            this.InitializationFailed = false;
+            this.AccessDenied = false;
+        }
+
+        public bool InitializationFailed { get; private set; }
+
+        public bool AccessDenied { get; private set; }
```
dnduffy: Why are these public properties?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/QuickPulseTopCpuCollector.cs
```diff
@@ -0,0 +1,154 @@
...
+
+                if (this.InitializationFailed)
+                {
+                    return Enumerable.Empty<Tuple<string, int>>();
+                }
+
+                if (this.AccessDenied && now - this.lastReadAttempt < this.accessDeniedRetryInterval)
+                {
+                    return Enumerable.Empty<Tuple<string, int>>();
+                }
```
dnduffy: Some comments somewhere explaining why this code carries on running and returns empty data when its internal state is bad would be good.

dnduffy: P.S. You could merge the two conditions and only have one return empty block.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/QuickPulseTopCpuCollector.cs
```diff
@@ -0,0 +1,154 @@
...
+                {
+                    encounteredProcs.Add(process.ProcessName);
+                    
+                    TimeSpan lastObservation;
+                    if (!this.processObservations.TryGetValue(process.ProcessName, out lastObservation))
+                    {
+                        // this is the first time we're encountering this process
+                        this.processObservations.Add(process.ProcessName, process.TotalProcessorTime);
+
+                        continue;
```
dnduffy: So if this is the first time you've seen it then why don't you want to capture any information about it below?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseDataSample.cs
```diff
@@ -69,7 +69,11 @@ public QuickPulseDataSample(QuickPulseDataAccumulator accumulator, IDictionary<s
                 p => (QuickPulseDefaults.CounterOriginalStringMapping.ContainsKey(p.Value.Item1.OriginalString) ? QuickPulseDefaults.CounterOriginalStringMapping[p.Value.Item1.OriginalString] : p.Value.Item1.OriginalString), 
                 p => p.Value.Item2);
 
+            this.TopCpuData = topCpuData.ToArray();
```
dnduffy: They are both already IEnumerable<Tuple<...>> so why call ToAray()?

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
```diff
@@ -24,6 +26,8 @@ public sealed class QuickPulseTelemetryModule : ITelemetryModule, IDisposable
     {
         private const int MaxSampleStorageSize = 10;
 
+        private const int TopCpuCount = 5;
```
dnduffy: Will customers ever be able to configure a different top count they want to see?

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
```diff
@@ -406,6 +428,33 @@ private void CollectionThreadWorker(object state)
                 Thread.Sleep(timeLeftUntilNextTick > TimeSpan.Zero ? timeLeftUntilNextTick : TimeSpan.Zero);
             }
         }
+        
+
+        private void InitializeCollectionThread()
+        {
+            try
+            {
+                this.topCpuCollector.Initialize();
```
dnduffy: So if this Initialize method is actually being used to start a thread rather than to merely setup state I would rename it to reflect what it is doing such as "Start", "StartCollecting", "Run", etc.

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
```diff
@@ -406,6 +428,33 @@ private void CollectionThreadWorker(object state)
...
+                // whatever happened, don't bring the thread down
+                QuickPulseEventSource.Log.UnknownErrorEvent(e.ToInvariantString());
+            }
+        }
+
+        private void CloseCollectionThread()
+        {
+            try
+            {
+                this.topCpuCollector.Close();
```
dnduffy: So based on Initialize really being more like "Start" I would suggest that maybe this should be "Stop" or something similar.

Src/PerformanceCollector/Unit.Tests/QuickPulse/Mocks/QuickPulseProcessProviderMock.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        public Exception AlwaysThrow { get; set; } = null;
+
+        public TimeSpan? OverallTimeValue { get; set; } = null;
+
+        public void Initialize()
+        {
+            if (this.AlwaysThrow != null)
+            {
+                throw this.AlwaysThrow;
```
dnduffy: There are mocking frameworks that would give you this for free.

Src/PerformanceCollector/Unit.Tests/QuickPulse/PerfLib/CategorySampleTests.cs
```diff
@@ -0,0 +1,52 @@
...
+            string resourceName = "Unit.Tests.QuickPulse.PerfLib.PerfData.data";
+
+            Stream stream = null;
+            try
+            {
+                stream = typeof(CategorySampleTests).Assembly.GetManifestResourceStream(resourceName);
+
+                using (StreamReader reader = new StreamReader(stream, Encoding.UTF8))
+                {
+                    stream = null;
```
dnduffy: Why?

Src/PerformanceCollector/Unit.Tests/QuickPulse/PerfLib/CategorySampleTests.cs
```diff
@@ -0,0 +1,52 @@
...
+
+                    dataList.AddRange(reader.ReadToEnd().Split(',').Select(byte.Parse));
+                }
+            }
+            finally
+            {
+                stream?.Dispose();
```
dnduffy: This is where I would expect to see "stream = null;" instead.

Src/PerformanceCollector/Unit.Tests/QuickPulse/PerfLib/CategorySampleTests.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            var perfLib = PerfLib.GetPerfLib();
+
+            // ACT
+            var categorySample = new CategorySample(data, 230, 6, perfLib);
+
+            // ASSERT
+            Assert.AreEqual(28, categorySample.CounterTable.Count);
+            Assert.AreEqual(165, categorySample.InstanceNameTable.Count);
+            Assert.AreEqual(6, categorySample.CounterTable.First().Key);
```
dnduffy: All these magic numbers need comments and explanations.

Src/PerformanceCollector/Unit.Tests/QuickPulse/QuickPulseServiceClientTests.cs
```diff
@@ -647,6 +680,59 @@ public void QuickPulseServiceClientSubmitsIsWebAppToService()
...
+            var cpuData = new[] { Tuple.Create("Process1", 1), Tuple.Create("Process2", 2) };
+
+            var sample = new QuickPulseDataSample(
+                new QuickPulseDataAccumulator { StartTimestamp = now, EndTimestamp = now.AddSeconds(1) },
+                new Dictionary<string, Tuple<PerformanceCounterData, double>>(),
+                cpuData,
+                false);
+
+            // ACT
+            serviceClient.SubmitSamples(new[] { sample }, ikey);
```
dnduffy: So is this really an integration test that makes real calls rather than a unit test or is the service client mocked in some way?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/CategorySample.cs
tokaplan: C# doesn't allow modification of readonly fields in an inherited class's constructor, unfortunately.

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
tokaplan: Done.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/CategorySample.cs
tokaplan: I'm leaving this in for reference in case we need to extract more counters in the future. This is an important extension point, so I think this is useful for someone reading it.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/IQuickPulseTopCpuCollector.cs
tokaplan: There's a school of thought (that I personally subscribe to) that less comments is better, so whenever you think object names don't make sense - leave a comment, that's what PRs are for :)

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/NativeMethods.cs
tokaplan: Well int is always Int32, so unless we want to change all built-in type usages (int -> Int32, string -> String, float -> Single, etc.) everywhere in code, this should be fine.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/PerfLib.cs
tokaplan: Not sure what you mean here. If instead of lazy initialization we do a regular one - we'd still have a static instance to store the object in, right? Regardless, for our usage it doesn't really matter, we don't call this 100 times in a loop or anything, so any approach which is functionally correct should be fine.

tokaplan: We could. The reason why we're creating an instance of this here is because we don't want the higher levels to deal with this low-level detail, while at the same time we want a singleton. Plus there's a nice lazy initialization benefit.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/PerfLib.cs
tokaplan: This is indeed not thread-safe and is not supposed to be thread-safe. We only use the PerfLib singleton from a single thread.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/PerformanceMonitor.cs
tokaplan: I think the original authors followed the literal definition of InvalidOperationException, which is "the exception that is thrown when a method call is invalid for the object's current state". Looks like a good enough match to me... doesn't really matter though, the message is clear enough.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/PerformanceMonitor.cs
tokaplan: This is a registry hive call, so the original message is something along the lines of "couldn't access a registry key", which is confusing given how we're actually accessing perf data :) So this rethrow is for the sake of the clarity of the message.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/QuickPulseProcessProvider.cs
tokaplan: Because that would be an extra constructor. The desired behavior is to use the default IQuickPulsePerfLib implementation whenever nothing is passed. The interface is primarily for unit testing.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/QuickPulseProcessProvider.cs
tokaplan: The way things turned out, it is indeed not needed. But I don't see it as harmful, either. Do you?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/QuickPulseProcessProvider.cs
tokaplan: Ooops, good point. Added comments.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/QuickPulseTopCpuCollector.cs
tokaplan: This is public interface (and is used by clients as such).

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/QuickPulseTopCpuCollector.cs
tokaplan: Good point, added comments.
As for merging conditions - I typically prefer to do that, however, in this particular case it looks a bit clunky and hard to follow (it was like that originally, and I split it).

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/QuickPulseTopCpuCollector.cs
tokaplan: We are capturing the time (note process.TotalProcessorTime usage in this line). We can't do much more (since we need two data points to calculate the CPU usage), so we just add the new process to the collection - along with the original value of its time value.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseDataSample.cs
tokaplan: In this method, it is important to completely decouple the input and the output. The input comes from the accumulator (that's cleared out and reused every two seconds), while the output is stored for later usage (we'll be sending it out to the server). So any coupling between the two is catastrophic, we need to create a new object.

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
tokaplan: No (be design).

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
tokaplan: No, Initialize() doesn't start a thread. This code is already executed within the top CPU collection thread.

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
tokaplan: See above. It doesn't start a thread, this is all already within the collection thread.

Src/PerformanceCollector/Unit.Tests/QuickPulse/Mocks/QuickPulseProcessProviderMock.cs
tokaplan: Yup me neither :)

Src/PerformanceCollector/Unit.Tests/QuickPulse/PerfLib/CategorySampleTests.cs
tokaplan: Because unfortunately you guys use StyleCop.
https://msdn.microsoft.com/en-us/library/ms182334.aspx

Src/PerformanceCollector/Unit.Tests/QuickPulse/PerfLib/CategorySampleTests.cs
tokaplan: You would. You would also probably prefer to see a pair of nested usings and nothing else. But alas (see above) :)

Src/PerformanceCollector/Unit.Tests/QuickPulse/PerfLib/CategorySampleTests.cs
tokaplan: Good catch, done.

Src/PerformanceCollector/Unit.Tests/QuickPulse/QuickPulseServiceClientTests.cs
tokaplan: Everything is mocked. All of these are unit tests.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/PerfLib.cs
pharring: Nit: For lazy initialization write this:
```C#
return library ?? (library = new PerfLib());
```
Note the inline assignment on the right hand side is not executed when the value is non-null (already initialized).
Although, frankly, you could just have ```private static readonly library = new PerfLib();``` and let the runtime deal with atomic, lazy initialization (see [beforefieldinit](https://www.bing.com/search?q=beforefieldinit))

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/IQuickPulseTopCpuCollector.cs
dnduffy: There's a school of thought that I have learned to believe in from experience that says document your code so other people can read the documentation.  Have you ever tried to read code in Reflector that has no comments?  It's not fun.  Have you ever tried to look something up on MSDN and found it had no comments/documentation?  Not fun either.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/CategorySample.cs
dnduffy: public CategorySample(Dictionary<int, CounterDefinitionSample> counterTable, ...) { ... }

class CategorySampleMock : CategorySample {
    public CategorySampleMock() : base(whateverDictionaryYouWantInTheMock) { ... }
...
}

You get the idea.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/CategorySample.cs
dnduffy: Okay that's fine, maybe enhance the comment to indicate that so no one deletes it in future.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/NativeMethods.cs
dnduffy: I'm from the days where C compilers kept changing the size of "int" so I was a bit paranoid about mapping to native structs.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/PerfLib.cs
dnduffy: But it's not a proper singleton in the strict sense and so the code could be brittle if someone in future accidentally misuses it.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/PerformanceMonitor.cs
dnduffy: okay

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/QuickPulseProcessProvider.cs
dnduffy: And the extra constructor costs what?  A few bytes in the DLL.
Unit testing is great, super important, but it should not mean that API's get weird.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/QuickPulseProcessProvider.cs
dnduffy: As Martin Fowler would say, it's a design smell.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseDataSample.cs
dnduffy: That's fair, please comment that so some future editor doesn't think they can remove it and thereby accidentally break it.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/CategorySample.cs
tokaplan: That's an option, but that would mean an extra constructor in the production implementation that we don't need (and it would make zero sense semantically, too). My judgement call is that we're ok without the readonly keyword here.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/CategorySample.cs
tokaplan: Makes sense, done.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/NativeMethods.cs
tokaplan: Relax, this mapping will never change :) It would be a bit expensive to remove native types in the entire codebase... We currently have 309 occurences of "int" alone in the Web SDK :)

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/PerfLib.cs
tokaplan: Nice :) But you do need to have C++ background to perceive this as natural :)
I remember in C++ you could have multiple values inside of the round brackets, and the last one would get returned: (a, b, c)  :)

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/PerfLib.cs
tokaplan: Of course, same as Dictionary<TKey,TValue> gets corrupted if you use it in a multithreaded context. I'd rather have the simplicity of not caring :)

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/QuickPulseProcessProvider.cs
tokaplan: Reviewed again - and turns out this is not even needed anymore, we can just use whatever's passed in. Fixed.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/QuickPulseProcessProvider.cs
tokaplan: Reviewed again - and turns out the delayed initialization is in fact needed. We only need to initialize if collection actually occurred. Some (possibly most) applications will go all their lives without ever collecting, so we don't want to initialize anything in that case; in fact, we don't even start the collection thread. @SergeyKanzhelev was super glad about that way back when :)

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/QuickPulseTopCpuCollector.cs
tokaplan: Exactly correct. We don't know how much CPU it's been consuming, but we will know once the next collection occurs.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseDataSample.cs
tokaplan: Done. Added comment at the top of the method.

SergeyKanzhelev:
I don't have much time to review carefully now, I saw @dnduffy reviewed it. @dnduffy, are you ready to sign off or there are something still missing?

Src/PerformanceCollector/Unit.Tests/QuickPulse/PerfLib/CategorySampleTests.cs
dnduffy: You must have run into some weird StyleCop or FxCop bug...

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/PerfLib.cs
```diff
@@ -13,13 +16,23 @@ private PerfLib()
         {
         }
 
+        /// <summary>
+        /// Gets the performance library instance.
+        /// </summary>
+        /// <returns>The performance library instance.</returns>
         public static PerfLib GetPerfLib()
         {
             library = library ?? new PerfLib();
```
dnduffy: You didn't use the lazy load optimization Paul Harrington commented on?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Process/PerfLib/CounterDefinitionSample.cs
```diff
@@ -0,0 +1,70 @@
...
+            this.CounterType = perfCounter.CounterType;
+            this.offset = perfCounter.CounterOffset;
+            this.size = perfCounter.CounterSize;
+            this.InstanceValues = instanceNumber == -1 ? new long[1] : new long[instanceNumber];
+        }
+
+        /// <summary>
+        /// Sets the value for an instance.
+        /// </summary>
+        /// <param name="index">Index value.</param>
```
dnduffy: Some of these could be more meaningful but thank-you for at least putting something.

PR #274, opened 2016-12-14
do not increment on rethrow

SergeyKanzhelev:
Need to know whether exception was re-thrown or not

### DIFF
--- a/Src/WindowsServer/WindowsServer.Shared.Tests/FirstChanceExceptionStatisticsTelemetryModuleTest.cs
+++ b/Src/WindowsServer/WindowsServer.Shared.Tests/FirstChanceExceptionStatisticsTelemetryModuleTest.cs
@@ -3,13 +3,14 @@
     using System;
     using System.Collections.Generic;
     using System.Runtime.ExceptionServices;
+    using DataContracts;
     using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.Extensibility;
     using Microsoft.ApplicationInsights.Web.TestFramework;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
     using TestFramework;
     using Assert = Xunit.Assert;
-    
+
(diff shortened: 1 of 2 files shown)


Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -172,7 +207,14 @@ private void TrackStatistics(string type, string operation, string method)
 
             var metric = this.metricManager.CreateMetric("Exceptions Thrown", dimensions);
 
-            metric.Track(1);
+            if (wasTracked)
+            {
+                metric.Track(0);
```
cijothomas: what does Track(0) do?

SergeyKanzhelev: It will not increment the value, but will increment a base

Src/WindowsServer/WindowsServer.Shared.Tests/FirstChanceExceptionStatisticsTelemetryModuleTest.cs
```diff
@@ -241,6 +241,129 @@ public void FirstChanceExceptionStatisticsTelemetryModuleWillNotDimCapTheSameOpe
...
+                module.Initialize(this.configuration);
+
+                var exception = new Exception();
+
+                Assert.False(module.WasExceptionTracked(exception));
+
+                var wrapper = new Exception("wrapper", exception);
+
+                Assert.True(module.WasExceptionTracked(wrapper));
+                Assert.True(module.WasExceptionTracked(wrapper));
```
Dmitry-Matveev: Was the intent to check exception but not wrapper in the second case? Or the second assert is to prove that the first WasTracked didn't change it back to false?

PR #278, opened 2016-12-16
Only add the cross component correlation key if the key does not rema…

nizarq:
…in the current component.

This avoid self references for which there isn't any nice experience in the viewing experience.


nizarq:
The reason why I am merging to master is that we never locked down develop. This means some changes might have gone into develop that need to go out in beta 2.  My idea was that for beta 1, ang small further checking required can go into master and then down into develop.  This will make sure we dont have to do any cherry picking. 

Let me know if that doesn't make sense.

nizarq:
Ok taking your suggestion - submitting a pull request for develop instead. #279

PR #279, opened 2016-12-20
Only add the cross component correlation key if the key does not rema…

nizarq:
…in the current component.

This avoid self references for which there isn't any nice experience in the viewing experience.

Adds unit tests to verify additional checks.

### DIFF
--- a/Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
+++ b/Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
@@ -102,12 +102,21 @@
                 requestTelemetry.Url = context.Request.UnvalidatedGetUrl();
             }
 
+            if (string.IsNullOrEmpty(requestTelemetry.Context.InstrumentationKey))
+            {
+                // Instrumentation key is probably empty, because the context has not yet had a chance to associate the requestTelemetry to the telemetry client yet.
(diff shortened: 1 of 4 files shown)


Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -102,12 +102,21 @@ public void OnEndRequest(HttpContext context)
                 requestTelemetry.Url = context.Request.UnvalidatedGetUrl();
             }
 
+            if (string.IsNullOrEmpty(requestTelemetry.Context.InstrumentationKey))
+            {
+                // Instrumentation key is probably empty, because the context has not yet had a chance to associate the requestTelemetry to the telemetry client yet.
+                // and get they instrumentation key from all possible sources in the process. Let's do that now.
+                this.telemetryClient.Initialize(requestTelemetry);
```
Dmitry-Matveev: Does it mean that Initializer can be now called two times for the same item? Will we have some performance implications with this?

nizarq: yes it's possible but this is something that I already added for DependencyTelemetry processing on Sergey's suggestion. Need it here as well now. Unfortunately, no way around it and the impact should be small.

PR #289, opened 2017-01-10
Sergkanz/exc stat improvements

SergeyKanzhelev:
Couple fixes for exc stats

### DIFF
--- a/Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
+++ b/Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
@@ -220,13 +220,22 @@
 
         private string GetDimCappedString(string value, ConcurrentDictionary<string, int> capValues)
         {
-            if (capValues.Count > 100)
+            int tmp;
+            var result = capValues.TryGetValue(value, out tmp);
+
+            if (result)
+            {
+                return value;
+            }
+            else if (capValues.Count > 100)
             {
                 return "OtherValue";
             }
-
-            capValues.TryAdd(value, 0);
-            return value;
+            else
+            {
+                capValues.TryAdd(value, 0);
+                return value;
+            }
         }
     }
 }
--- a/Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
+++ b/Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
@@ -218,14 +218,12 @@
             }
         }
 
-        private string GetDimCappedString(string value, ConcurrentDictionary<string, int> capValues)
+        private string GetDimCappedString(string dimensionValue, ConcurrentDictionary<string, int> capValues)
         {
-            int tmp;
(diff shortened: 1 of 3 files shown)


Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -220,13 +220,22 @@ private void TrackStatistics(string type, string operation, string method, bool
 
         private string GetDimCappedString(string value, ConcurrentDictionary<string, int> capValues)
```
cijothomas: can you rename value to something like dimensionValue or something? value seem like a keyword...

Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -220,13 +220,22 @@ private void TrackStatistics(string type, string operation, string method, bool
...
         {
-            if (capValues.Count > 100)
+            int tmp;
```
cijothomas: did the unit test catch this issue? If not, consider adding a test.

dnduffy: nitpick "temp"

Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -220,13 +220,22 @@ private void TrackStatistics(string type, string operation, string method, bool
...
+            var result = capValues.TryGetValue(value, out tmp);
```
dnduffy: Is this supposed to be "bool"?

Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -220,13 +220,22 @@ private void TrackStatistics(string type, string operation, string method, bool
...
+
+            if (result)
+            {
+                return value;
+            }
+            else if (capValues.Count > 100)
```
Dmitry-Matveev: If it's a concurrent code, the capValues.Count > 100 might return true and false for the same value:
- Thread A enters the method with value "ABC", capValues.Count = 100; "ABC" is not in the dictionary
- Thread B enters the method with value "ABC", capValues.Count = 100; "ABC" is not in the dictionary
- Whoever adds it first (say "A") might create a case where both "A" and "B" passed TryGetValue() and it returned nothing. Then "A" added it and capValues.Count is now 101, so "B" does not pass the check and gets "OtherValue" instead. Is this a positive concurrency on purpose or this code is not executed in parallel?

SergeyKanzhelev: we do not need strong capping, just preventing from hitting ourselves in a foot. And I also wait for proper built-in functionality of dim capping in Metric Aggregation

Issue #286, opened 2017-01-09
Remove custom cache implementation


cijothomas:
Web SDK has implemented own cache solution as the framework provided MemoryCache was not available in Windows Phone/Store. Since WebSDK no longer need to support Phone/Store, there is no reason to continue using the custom cache implementation.

The current cache implementation has a Clock related bug as well, as mentioned here:
https://github.com/Microsoft/ApplicationInsights-dotnet/issues/271#event-865053671

Custom CacheProvider code:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared/Implementation/Operation/CacheProvider.cs

Instead of fixing the clock issue, it better to remove the cache implementation completely and depend on framework provided MemoryCache

PR #290, opened 2017-01-10
Remove custom cache used to hold dependency operations

cijothomas:
Removed custom cache provider, and instead use the framework provide cache. Framework provided cache was not used originally as it was not supported in WinRT, but now SDK no longer supports WinRT we can simply use built in cache, instead of implementing own cache. (which had some known bugs as well)
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/286


Src/DependencyCollector/Shared/Implementation/Operation/CacheProvider.cs
```diff
@@ -3,392 +3,34 @@
...
+    using System.Threading;        
     using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
 
     /// <summary>
-    /// The implementation of cache provider for Windows Phone and Windows Store as MemoryCache is not available there.
+    /// The wrapper for MemoryCache.
     /// </summary>
     /// <typeparam name="TValue">Type of items to store in the cache.</typeparam>
-    internal sealed class CacheProvider<TValue> : ICacheProvider<TValue>
+    internal class CacheProvider<TValue> : ICacheProvider<TValue>
```
dnduffy: Why keep the class at all?
Unless it is used heavily in multiple places I think we should just switch the code that uses this class to use MemoryCache directly instead and delete the rest of it.

Src/DependencyCollector/Shared/Implementation/Operation/CacheProvider.cs
```diff
@@ -435,19 +77,15 @@ public void Set(string key, TValue value)
...
+            return this.memoryCache.Remove(key) != null;
         }
 
         /// <summary>
         /// Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.
         /// </summary>
         public void Dispose()
         {
-            // there is nothing to dispose here.
+            // nothing to dispose.
```
dnduffy: Clearly this class doesn't need to be disposable.

Src/DependencyCollector/Shared/Implementation/Operation/CacheProvider.cs
cijothomas: intended to abstract away the memory cache specifics, like setting expiration policy etc. I feel its better to have it abstracted and unit tested.

Src/DependencyCollector/Shared/Implementation/Operation/CacheProvider.cs
cijothomas: Correct, will remove Disposable.

Src/DependencyCollector/Shared/Implementation/Operation/CacheProvider.cs
dnduffy: Right, except we still have all of those specifics, just in an extra wrapper class, which is not reused, and the unit tests are now testing MemoryCache instead of our code which means they should get deleted also.

cijothomas: I'll move the logic to CacheBasedOperationHolder and ensure unit tests are covering it fully.

dnduffy: Thank-you. :)

cijothomas:
After I added dispose forMemoryCache, some unit tests are failing when run together. Class:"DependencyTrackingTelemetryModuleTest" - tests from this class is failing. Functional tests are passing. I think this is because of how the above unit tests are written and somehow its clashing with MemoryCache dispose.. need to investigate before merge

Src/DependencyCollector/Shared/Implementation/DependencyTableStore.cs
```diff
@@ -35,6 +35,7 @@ internal static DependencyTableStore Instance
 
         public void Dispose()
         {
+            instance = null;
```
cijothomas: not really sure if this is the best way...

Src/DependencyCollector/Shared.Tests/Implementation/FrameworkHttpProcessingTest.cs
```diff
@@ -236,8 +236,7 @@ public void RddTestHttpProcessingFrameworkStartTimeFromGetRequestStreamAsync()
                
         #region Disposable
         public void Dispose()
-        {
-            this.httpProcessingFramework.Dispose();
+        {            
             this.configuration.Dispose();
```
SergeyKanzhelev: should you call `Dispose` in the TestUnintialize?

Src/DependencyCollector/Shared/Implementation/DependencyTableStore.cs
SergeyKanzhelev: so it it not a singleton any longer. Just instantiate it's own instance then

Src/DependencyCollector/Shared.Tests/Implementation/Operation/CacheBasedOperationHolderTests.cs
```diff
@@ -95,6 +96,19 @@ public void GetReturnsItemIfItExistsInTheTable()
...
 
+        [TestMethod]
+        public void TestItemExpiration()
+        {
+            long id = 911911;
+            this.cacheBasedOperationHolder.Store(id, this.telemetryTuple);
+            Assert.AreEqual(this.telemetryTuple, this.cacheBasedOperationHolder.Get(id));
+
+            // Sleep for 2 secs which is more than the cache expiration time.
+            Thread.Sleep(2000);
```
dnduffy: Two comments:
1) It would be good if this test explicitly setup the cache policy it expects (and restores stated when finished if needed).
2) Try to make it faster, a lot faster, sleeping for two seconds in a test is not ideal (when it's one test it doesn't seem so bad but when it is hundreds of tests and they all sleep a bit suddenly the unit tests take a long time to run).

Src/DependencyCollector/Shared/Implementation/DependencyTableStore.cs
dnduffy: There is an example of disposing of a static singleton that I found in the base SDK where it does this:
```
    if (disposing)
    {
        Interlocked.CompareExchange(ref instance, null, this);
    ...
```

Src/DependencyCollector/Shared/Implementation/FrameworkSqlEventListener.cs
```diff
@@ -47,8 +48,7 @@ private enum CompositeState
         /// Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.
         /// </summary>
         public override void Dispose()
```
dnduffy: If the only thing in here is base.Dispose() then this override can just be deleted.

Src/DependencyCollector/Shared.Tests/Implementation/Operation/CacheBasedOperationHolderTests.cs
cijothomas: Agreed. Will change. 
Regarding timing, i don't think i can go below 1sec as then it might cause random failures since the exact sleep time is not guaranteed.

Src/DependencyCollector/Shared/Implementation/FrameworkSqlEventListener.cs
cijothomas: Ok.

Src/DependencyCollector/Shared.Tests/Implementation/FrameworkHttpProcessingTest.cs
cijothomas: yes done.

Src/DependencyCollector/Shared/Implementation/DependencyTableStore.cs
cijothomas: Its singleton for a reason and I am not changing it. Dipose() will simply call the dispose of the Disposable instances it created.

PR #292, opened 2017-01-13
Fixed message to show which counters had errors.

dnduffy:


Src/PerformanceCollector/Shared/Implementation/PerformanceCollectorEventSource.cs
```diff
@@ -72,7 +72,7 @@ public void CounterParsingFailedEvent(string e, string counter, string applicati
 
         [Event(8, Keywords = Keywords.UserActionable, Level = EventLevel.Error,
             Message = @"Error collecting {0} of the configured performance counters. Please check the configuration.
-{2}")]
+{1}")]
```
cijothomas: from the code it looks like this prints the whole errors which contain the perfcounter name.. did you manually validate it? is the message readable/make sense for customer?

dnduffy: That's what it does, I accidentally broke it with the quick fix which is why we thought we needed to still add that info.

PR #296, opened 2017-01-24
Use Exception's Data dictionary instead of ConditionalWeakTable

pharring:
I noticed we are using ConditionalWeakTable as a side-channel to indicate whether an exception is tracked.

The Exception object already has a general purpose Data dictionary which can be used for this purpose and will beat ConditionalWeakTable in performance.

@SergeyKanzhelev what do you think? Did you consider this already?


cijothomas:
this PR breaks build, some style cop issues. Need to be addressed

pharring:
@cijothomas Sorry about that. I'm new to this code base and I didn't know to run Style Cop. I'll investigate. Is there an issue for the build break?

cijothomas:
@pharring its ok. You can run https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/buildRelease.cmd locally to see the build errors (buildRelease runs stylecop stuffs)
No issues are opened on build break.

PR #297, opened 2017-01-25
Fix StyleCop issues

pharring:
1. Remove unused using
2. Rename private static field with initial capital letter
3. Move private static readonly field above private static (non-readonly) field
4. Move private static method after internal method, but before private non-static method


pharring:
@cijothomas This fixes the build breaks introduced in #296

PR #298, opened 2017-01-25
Add instrumentation for SqlConnection.Open/OpenAsync

iusafaro:


Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
```diff
@@ -253,6 +251,80 @@ private object OnBegin(object thisObj)
...
+                    if (t.IsFaulted)
+                    {
+                        // track item only in case of failure
+                        exceptionObj = t.Exception != null && t.Exception.InnerException != null ? t.Exception.InnerException : t.Exception;
+                        this.OnEnd(exceptionObj, thisObj);
+                    }
+                    else
+                    {
+                        this.TelemetryTable.Remove(thisObj);
+                    }
```
Dmitry-Matveev: Do we need to wrap it with try/catch all? It looks like it's going to be executed in the context of the user code as a continuation of the user task, right?

Test/DependencyCollector/FunctionalTests/FuncTest/SqlTests.cs
```diff
@@ -336,6 +342,110 @@ public void TestSqlCommandExecuteXmlReaderFailed()
...
+                         Assert.AreEqual(1, sqlItems.Length, "Total Count of Remote Dependency items for SQL collected is wrong.");
+
+                         this.Validate(
+                             sqlItems[0],
+                             ResourceNameSQLToWrongDatabase,
+                             "Open",
+                             TimeSpan.FromSeconds(20),
+                             successFlagExpected: success,
+                             sqlErrorCodeExpected: "-1",
+                             sqlErrorMessageExpected: "A network-related or instance-specific error occurred while establishing a connection to SQL Server. The server was not found or was not accessible. Verify that the instance name is correct and that SQL Server is configured to allow remote connections. (provider: SQL Network Interfaces, error: 26 - Error Locating Server/Instance Specified)");
```
Dmitry-Matveev: This seems to be great and accurate condition but somewhat error-prone, the test failures might be induced by the updates in .NET FW rather than the updates in our code. I remember we were validating such messages in APM agent and tests were sometimes broken after update 4.0->4.5 and 4.5->4.6. Maybe, (contains) to look for some token that will always logically be present in this message?

Src/DependencyCollector/Shared/Implementation/ProfilerSqlCommandProcessing.cs
```diff
@@ -0,0 +1,86 @@
...
+        {            
+        }
+
+        /// <summary>
+        /// Gets SQL command resource name.
+        /// </summary>
+        /// <param name="thisObj">The SQL command.</param>
+        /// <remarks>Before we have clarity with SQL team around EventSource instrumentation, providing name as a concatenation of parameters.</remarks>
+        /// <returns>The resource name if possible otherwise empty string.</returns>
+        internal override string GetResourceName(object thisObj)
```
SergeyKanzhelev: Let's rename those to `GetDependencyName`. This is old and legacy name

Src/DependencyCollector/Shared/Implementation/ProfilerSqlCommandProcessing.cs
```diff
@@ -0,0 +1,86 @@
...
+        internal ProfilerSqlCommandProcessing(TelemetryConfiguration configuration, string agentVersion, ObjectInstanceBasedOperationHolder telemetryTupleHolder)
+            : base(configuration, agentVersion, telemetryTupleHolder)
+        {            
+        }
+
+        /// <summary>
+        /// Gets SQL command resource name.
+        /// </summary>
+        /// <param name="thisObj">The SQL command.</param>
+        /// <remarks>Before we have clarity with SQL team around EventSource instrumentation, providing name as a concatenation of parameters.</remarks>
```
SergeyKanzhelev: Is this remarks applicable here?

Test/DependencyCollector/FunctionalTests/FuncTest/SqlTests.cs
```diff
@@ -23,6 +24,11 @@ public class SqlTests
         private const string ResourceNameSQLToDevApm = @".\SQLEXPRESS | RDDTestDatabase";
 
         /// <summary>
+        /// Resource Name for wrong database.
+        /// </summary>
+        private const string ResourceNameSQLToWrongDatabase = @"invalid\SQLEXPRESS | RDDTestDatabase";
```
SergeyKanzhelev: Have you tested auth errors in SQL.Open? DNS errors typically thrown before any attempt to open connection will be made so the flow of the code execution for different errors will be different

Src/DependencyCollector/Shared/Implementation/ProfilerSqlCommandProcessing.cs
```diff
@@ -0,0 +1,86 @@
+﻿namespace Microsoft.ApplicationInsights.DependencyCollector.Implementation
+{
+    using System.Data;
+    using System.Data.SqlClient;
+    using Microsoft.ApplicationInsights.DependencyCollector.Implementation.Operation;
+    using Microsoft.ApplicationInsights.Extensibility;
+
+    /// <summary>
+    /// Concrete class with all processing logic to generate RDD data from the calls backs
```
cijothomas: replace RDD with Dependency wherever you see :)

Src/DependencyCollector/Shared/Implementation/ProfilerSqlConnectionProcessing.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <param name="thisObj">The SQL connection.</param>
+        /// <returns>Returns predefined command text.</returns>
+        internal override string GetCommandName(object thisObj)
+        {            
+            return SqlConnectionCommandText;
+        }
+
+        private string GetResourceNameInternal(object thisObj)
+        {
+            string resource = string.Empty;
```
cijothomas: Could refactor a bit and the move the common logic to the base class. Both SQLCommandProcessor and SQLConnectionProcessor has duplicated the logic to extract some name from SqlConnection object. That could be moved to base class.

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
```diff
@@ -92,6 +92,24 @@ public object OnEndForOneParameter(object context, object returnValue, object th
         }
 
         /// <summary>
+        /// On end async callback for methods with 1 parameter.
+        /// </summary>
+        public object OnEndAsyncForOneParameter(object context, object returnValue, object thisObj)
```
cijothomas: Who is calling this? Decorator only adds callback for Begin and OnException right?

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
```diff
@@ -101,6 +119,24 @@ public object OnEndForTwoParameters(object context, object returnValue, object t
...
+        /// On end async callback for methods with 2 parameter.
+        /// </summary>
+        public object OnEndAsyncForTwoParameters(object context, object returnValue, object thisObj, object parameter1)
+        {
+            this.OnEndAsync(returnValue, thisObj);
+            return returnValue;
+        }
+
+        /// <summary>
+        /// On end async callback for methods with 2 parameter.
```
cijothomas: comment is misleading. This is callback for OnException

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
```diff
@@ -253,6 +251,80 @@ private object OnBegin(object thisObj)
...
+                DependencyCollectorEventSource.Log.CallbackError(thisObj == null ? 0 : thisObj.GetHashCode(), "OnEndAsyncSql", ex);
+            }
+        }
+
+        /// <summary>
+        ///  Common helper for all EndAsync Callbacks that should send data only in the case of exception happened.
+        /// </summary>
+        /// <param name="taskObj">Returned task by the async method.</param>
+        /// <param name="thisObj">This object.</param>
+        private void OnEndExceptionAsync(object taskObj, object thisObj)
```
cijothomas: suggestion: change the method name to OnEndAsync to avoid confusing name  - we dont know if exception is really thrown yet.

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
```diff
@@ -253,6 +251,80 @@ private object OnBegin(object thisObj)
...
+
+                DependencyCollectorEventSource.Log.EndAsyncExceptionCallbackCalled(thisObj.GetHashCode().ToString(CultureInfo.InvariantCulture));
+
+                task.ContinueWith(t =>
+                {
+                    Exception exceptionObj = null;
+                    if (t.IsFaulted)
+                    {
+                        // track item only in case of failure
+                        exceptionObj = t.Exception != null && t.Exception.InnerException != null ? t.Exception.InnerException : t.Exception;
```
cijothomas: if IsFaulted is true, then you don't need to check if t.Exception is not null - it will be non-null always as per documentation:
https://msdn.microsoft.com/en-us/library/system.threading.tasks.task.isfaulted(v=vs.110).aspx

Src/DependencyCollector/Shared/Implementation/ProfilerSqlConnectionProcessing.cs
```diff
@@ -0,0 +1,72 @@
+﻿namespace Microsoft.ApplicationInsights.DependencyCollector.Implementation
+{
+    using System;
+    using System.Data.SqlClient;
+    using System.Threading.Tasks;
```
iusafaro: For casting to Task

---
In reply to: [97904986](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r97904986) [](ancestors = 97904986)

Src/DependencyCollector/Shared/Implementation/ProfilerSqlConnectionProcessing.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        /// <summary>
+        /// Gets SQL connection resource target name.
+        /// </summary>
+        /// <param name="thisObj">The SQL connection.</param>
+        /// <returns>The resource target name if possible otherwise empty string.</returns>
+        internal override string GetResourceTarget(object thisObj)
+        {
+            return this.GetResourceNameInternal(thisObj);
```
iusafaro: in getcommandname command text i.e. "Open" should be appended - fixed. thanks!

---
In reply to: [97905118](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r97905118) [](ancestors = 97905118)

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
```diff
@@ -253,6 +251,80 @@ private object OnBegin(object thisObj)
...
+                }
+
+                DependencyCollectorEventSource.Log.EndAsyncCallbackCalled(thisObj.GetHashCode().ToString(CultureInfo.InvariantCulture));
+
+                task.ContinueWith(t =>
+                {
+                    Exception exceptionObj = null;
+                    if (t.IsFaulted)
+                    {
+                        exceptionObj = t.Exception != null && t.Exception.InnerException != null ? t.Exception.InnerException : t.Exception;
```
iusafaro: yep. refactored.

---
In reply to: [97905656](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r97905656) [](ancestors = 97905656)

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
```diff
@@ -253,6 +251,80 @@ private object OnBegin(object thisObj)
         }
 
         /// <summary>
+        ///  Common helper for all EndAsync Callbacks.
+        /// </summary>
+        /// <param name="taskObj">Returned task by the async method.</param>
+        /// <param name="thisObj">This object.</param>
+        private void OnEndAsync(object taskObj, object thisObj)
```
iusafaro: no, that's our callbacks and we have an access to the return value i.e. Task only in OnEnd

---
In reply to: [97906094](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r97906094) [](ancestors = 97906094)

Src/DependencyCollector/Shared.Tests/Implementation/ProfilerSqlProcessingTest.cs
```diff
@@ -361,6 +366,280 @@ public void OnEndLogsWarningWhenObjectPassedToEndWithoutCorrespondingBeginAsync(
...
+            Assert.AreEqual(ExpectedTarget, actualTargetName);
+
+            string actualCommandText = this.sqlConnectionProcessingProfiler.GetCommandName(connection);
+            Assert.AreEqual(ExpectedCommandTextForSqlConnection, actualCommandText);
+        }
+
+        [TestMethod]
+        public void OnEndAsyncSendsCorrectTelemetry_1ArgumentOverride()
+        {
+            var command = GetSqlCommandTestForStoredProc();
```
dnduffy: nitpick: I have no idea what the type is here.

Src/DependencyCollector/Shared/Implementation/ProfilerSqlCommandProcessing.cs
iusafaro: left from old code - probably not? removed.

---
In reply to: [98011375](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r98011375) [](ancestors = 98011375)

Test/DependencyCollector/FunctionalTests/TestApps/Aspx451/ExternalCalls.aspx.cs
```diff
@@ -20,6 +20,11 @@ public partial class ExternalCalls : System.Web.UI.Page
         private const string ConnectionString = @"Data Source=.\SQLEXPRESS;Initial Catalog=RDDTestDatabase;Integrated Security=True";
 
         /// <summary>
+        /// Invalid connection string to database.
+        /// </summary> 
+        private const string InvalidConnectionString = @"Data Source=invalid\SQLEXPRESS;Initial Catalog=RDDTestDatabase;Integrated Security=True";
```
cijothomas: all the scenarios for validating Open() failure is using a SQLServer which do not exist. Can we add separate scenarios like:
1. A SQL Server which dont exist.
2. SQL Serverinstance exists, but the database name dont exist.
3. Uses incorrect credentials to simulate failure.
4. Add blocking firewall rule to simulate failures.
(dont need all , but atleast 2 or maybe 3 different ways of simulating failure)

Src/DependencyCollector/Shared.Tests/Implementation/ProfilerSqlProcessingTest.cs
```diff
@@ -361,6 +366,280 @@ public void OnEndLogsWarningWhenObjectPassedToEndWithoutCorrespondingBeginAsync(
...
+
+            var returnTaskPassed = Task.Factory.StartNew(() => { });
+
+            var context = this.sqlCommandProcessingProfiler.OnBeginForOneParameter(command);
+            var objectReturned = this.sqlCommandProcessingProfiler.OnEndAsyncForOneParameter(context, returnTaskPassed, command);
+
+            stopwatch.Stop();
+
+            // wait for OnEnd async completion
+            Thread.Sleep(50);
```
dnduffy: Is there something we can get a signal of completion from instead of sleeping?  Sometimes sleep dependent tests start failing when hardware changes or the framework changes.

Src/DependencyCollector/Shared.Tests/Implementation/ProfilerSqlProcessingTest.cs
```diff
@@ -361,6 +366,280 @@ public void OnEndLogsWarningWhenObjectPassedToEndWithoutCorrespondingBeginAsync(
...
+                    Assert.Fail("Should not be throwing unhandled exceptions.");
+                }
+
+                var message = listener.Messages.First(item => item.EventId == 14);
+                Assert.IsNotNull(message);
+            }
+        }
+
+        [TestMethod]
+        public void OnEndExceptionAsyncNotSendsTelemetryIfSuccess_1ArgumentOverride()
```
dnduffy: nitpick: Grammar - "...DoesNotSend..."
Here and below with similar test method names.

Src/DependencyCollector/Shared/Implementation/ProfilerSqlCommandProcessing.cs
dnduffy: nitpick: "calls" -> "call"

Src/DependencyCollector/Shared/Implementation/ProfilerSqlConnectionProcessing.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// </summary>
+        internal ProfilerSqlConnectionProcessing(TelemetryConfiguration configuration, string agentVersion, ObjectInstanceBasedOperationHolder telemetryTupleHolder)
+            : base(configuration, agentVersion, telemetryTupleHolder)
+        {
+        }              
+
+        /// <summary>
+        /// Gets SQL connection resource name.
+        /// </summary>
+        /// <param name="thisObj">The SQL connection.</param>
```
dnduffy: If it is a SQL connection then please rename the variable to reflect what it is; "thisObj" is not a meaningful name.

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
dnduffy: If I understand correctly the code uses different parameter signature versions depending on the number of parameters so the appropriate end handler is wired up.

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
dnduffy: There is a flag you can pass to ContinueWith so that it only runs when it is faulted, would that work for this scenario?

Src/DependencyCollector/Shared/Implementation/ProfilerSqlCommandProcessing.cs
iusafaro: renamed

---
In reply to: [98011264](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r98011264) [](ancestors = 98011264)

Src/DependencyCollector/Shared/Implementation/ProfilerSqlCommandProcessing.cs
iusafaro: replaced in several places - at least in the code that i touched

---
In reply to: [98049580](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r98049580) [](ancestors = 98049580)

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
iusafaro: makes sense. updated.

---
In reply to: [98071623](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r98071623) [](ancestors = 98071623)

Src/DependencyCollector/Shared/Implementation/ProfilerSqlConnectionProcessing.cs
iusafaro: i'd like to leave as is if no objections - string manipulations are done slightly bit different in different places and let's have them being optimized as much as possible.

---
In reply to: [98055552](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r98055552) [](ancestors = 98055552)

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
iusafaro: Is not used now, but will be when we'll update SQL async methods instrumentation.

---
In reply to: [98060621](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r98060621) [](ancestors = 98060621)

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
iusafaro: no, that's OnEnd that sends event only in case of exception.

---
In reply to: [98065128](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r98065128) [](ancestors = 98065128)

Src/DependencyCollector/Shared.Tests/Implementation/ProfilerSqlProcessingTest.cs
iusafaro: method name is self descriptive, and it's test.

---
In reply to: [98072442](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r98072442) [](ancestors = 98072442)

Src/DependencyCollector/Shared.Tests/Implementation/ProfilerSqlProcessingTest.cs
iusafaro: updated

---
In reply to: [98075481](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r98075481) [](ancestors = 98075481)

Src/DependencyCollector/Shared/Implementation/ProfilerSqlCommandProcessing.cs
iusafaro: updated comments

---
In reply to: [98076447](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r98076447) [](ancestors = 98076447)

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
iusafaro: yes, i thought about - it's definitely improvement, but from the other side we'll keep thisObj object in the table and here i;m removing it in any case even if task is not faulted. let me know what you think about this.

---
In reply to: [98079670](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r98079670) [](ancestors = 98079670)

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
iusafaro: yes

---
In reply to: [98078491](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r98078491) [](ancestors = 98078491)

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
iusafaro: there's OnEndAsync already, let's leave as is? comment is pretty self explanatory.

---
In reply to: [98070645](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r98070645) [](ancestors = 98070645)

Src/DependencyCollector/Shared.Tests/Implementation/ProfilerSqlProcessingTest.cs
iusafaro: nope, as we don't return resulted continuation task, but only have initial one in return.

---
In reply to: [98073573](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r98073573) [](ancestors = 98073573)

Src/DependencyCollector/Shared/Implementation/ProfilerSqlConnectionProcessing.cs
iusafaro: i'll leave as is - for me it's more meaningful as i understand that it's exactly 'this' object that i got from the instrumentation

---
In reply to: [98077223](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r98077223) [](ancestors = 98077223)

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
iusafaro: i've added checks for thisObj==null - that's the only place that can throw an exception in this.TelemetryTable.Remove(), other places are safe.

---
In reply to: [97908795](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/298#discussion_r97908795) [](ancestors = 97908795)

Src/DependencyCollector/Shared.Tests/Implementation/ProfilerSqlProcessingTest.cs
dnduffy: I think the method name is great, no complaints about the method name.
I just don't know what "command" is, or what "context" is, etc.  There is no intellisense in GitHub.

Test/DependencyCollector/FunctionalTests/FuncTest/SqlTests.cs
iusafaro: Yes, have tested different cases manually. Added 'invalid account' case in functional tests.

SergeyKanzhelev: I cannot find it in PR. Maybe just set up connection string with sql creds

Test/DependencyCollector/FunctionalTests/TestApps/Aspx451/ExternalCalls.aspx.cs
iusafaro: I've added incorrect credentials case. Do you think that we really need to test many different scenarios in functional tests? We covered instrumentation and information retrieval for two diff cases (agree that one was not enough), but it's not the goal to cover all possible exceptions from Open...

SergeyKanzhelev: Agree. If we will find something else - we will add later

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
cijothomas: both this method and the one right above it has exact same comment but does something different. Hence I asked to modify the comment to reflect what the function is doing.

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
cijothomas: "OnEndExceptionAsync" gives the impression that this is called on exception but it is not. Its called OnTaskEnd (?) and we determine inside if task ended nicely or threw ex.
(I agree comment is self explanatory)

Test/DependencyCollector/FunctionalTests/TestApps/Aspx451/ExternalCalls.aspx.cs
cijothomas: Invalid account credentials is good - we connect to sql and fail to open. (the original one was failing to even find a sql instance..hence i suggested adding more scenario)

Test/DependencyCollector/FunctionalTests/FuncTest/SqlTests.cs
cijothomas: @SergeyKanzhelev its already added. InvalidAccountConnectionString

Src/DependencyCollector/Shared/Implementation/ProfilerSqlConnectionProcessing.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        private string GetResourceNameInternal(object thisObj)
+        {
+            string resource = string.Empty;
+
+            SqlConnection connection = thisObj as SqlConnection;            
+            if (connection != null)
+            {
+                resource = string.Join(" | ", connection.DataSource, connection.Database);
+            }
```
cijothomas: string.Join(" | ", command.Connection.DataSource, command.Connection.Database); - this is the logic which is common for command and connection. You can have a method in base class which takes Connection as argument and return this name.

Src/DependencyCollector/Shared/Implementation/ProfilerSqlConnectionProcessing.cs
cijothomas: string.Join(" | ", command.Connection.DataSource, command.Connection.Database); - this is the logic which is common for command and connection. You can have a method in base class which takes Connection as argument and return this name.

cijothomas:
@iusafaro For some reason my comments disappeared it seems!.

Here's the only one i had left:

string.Join(" | ", command.Connection.DataSource, command.Connection.Database); - this is the logic which is common for command and connection processing. You can have a method in base class which takes Connection as argument and return this name.

Issue #237, opened 2016-11-29
WebApp PerfCounters - exception handling is not consistent with Standard PerfCounter


cijothomas:
1. The messages logged when 1st time a counter is attempted to be registered is different for WebApp vs Standard.

2. After initial registration, any counter marked as 'bad' will be re-attempted once in 5 min. If this read fails, StandardPerfCollector reports error 'CounterRegistrationFailed', but WebApp reports 'UnknownErrrorEvent', and 'FirstReadFailed'. This would make very difficult to debug if things go bad for a customer.

kartang:
The reason we used different registration messages is to distinguish between the total usage of standard counters vs web app counters. The same is applied for failures as well. We have an additional event for failed counter reads/registration, because, there is a scope to create a user actionable event if the target counter is not in the supported list of web app counters (which is different from failed registration/read).

cijothomas:
@kartang to distinguish between usage of standard vs webapp, we had sdk version pre-fix.

cijothomas:
After initial registration, any counter marked as 'bad' will be re-attempted once in 5 min. If this read fails, StandardPerfCollector reports error 'CounterRegistrationFailed', but WebApp reports 'UnknownErrrorEvent', and 'FirstReadFailed'. This would make very difficult to debug if things go bad for a customer.

-This statement was incorrect, both WebApp and Standard handles this similar way. PerformanceCounterFirstReadFailed is whats reported by both.

PR #299, opened 2017-01-27
Modified WebApp counter to behave same as Standard counter.

cijothomas:
(minor changes in logging error messages only)

Also fix:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/237

### DIFF
--- a/Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/WebAppPerformanceCollector.cs
+++ b/Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/WebAppPerformanceCollector.cs
@@ -95,6 +95,11 @@
                 if (pc != null)
                 {
                     this.RegisterPerformanceCounter(perfCounter, this.GetCounterReportAsName(perfCounter, reportAs), pc.CategoryName, pc.CounterName, pc.InstanceName, useInstancePlaceHolder, false);
+                }
+                else
+                {
+                    // Even if validation failed, we might still be able to collect perf counter in WebApp.
+                    this.RegisterPerformanceCounter(perfCounter, this.GetCounterReportAsName(perfCounter, reportAs), string.Empty, perfCounter, string.Empty, useInstancePlaceHolder, false);
                 }                
             }
             catch (Exception e)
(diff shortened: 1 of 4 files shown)


Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/WebAppPerformanceCollector.cs
```diff
@@ -86,28 +86,16 @@ public void RefreshCounters()
...
+                bool useInstancePlaceHolder = false;                
+                var pc = PerformanceCounterUtility.CreateAndValidateCounter(perfCounter, null, null, out useInstancePlaceHolder, out error);
+                
                 if (pc != null)
                 {
                     this.RegisterPerformanceCounter(perfCounter, this.GetCounterReportAsName(perfCounter, reportAs), pc.CategoryName, pc.CounterName, pc.InstanceName, useInstancePlaceHolder, false);
-                }
-                else
-                {
-                    this.RegisterPerformanceCounter(perfCounter, this.GetCounterReportAsName(perfCounter, reportAs), string.Empty, perfCounter, string.Empty, useInstancePlaceHolder, false);
```
kartang: Why is this condition ignored? I believe we can still register the performance counter for web apps if the validation fails. We might just be missing Instance name and Category name?

cijothomas: I didn't knew that webapp perf counter works even if validation fails.. But if that's the case, why bother validating? Or should we add a separate validator for webappperf counter?

kartang: We do the validation to get the category name and instance name. If available, we pass them as parameters, and if not, we use empty strings.

cijothomas: got it. I will bring back the original behavior and add a comment explaining this.

cijothomas:
@kartang can you take a look now?

kartang:
:shipit:

PR #300, opened 2017-01-28
Remove duration validation from aync tests as it's not reliable

iusafaro:


Src/DependencyCollector/Shared.Tests/Implementation/ProfilerSqlProcessingTest.cs
```diff
@@ -659,7 +659,11 @@ public void OnEndExceptionAsyncLogsWarningWhenNullTaskPassedToEndMethods()
             Assert.AreEqual(expectedSuccess, remoteDependencyTelemetryActual.Success, "Success in the sent telemetry is wrong");
             Assert.AreEqual(expectedResultCode, remoteDependencyTelemetryActual.ResultCode, "ResultCode in the sent telemetry is wrong");
 
-            Assert.IsTrue(remoteDependencyTelemetryActual.Duration.TotalMilliseconds <= timeBetweenBeginEndInMs + (async ? 2 : 1), "Incorrect duration. Collected " + remoteDependencyTelemetryActual.Duration.TotalMilliseconds + " should be less than max " + timeBetweenBeginEndInMs);
+            if (!async)
+            {
+                Assert.IsTrue(remoteDependencyTelemetryActual.Duration.TotalMilliseconds <= timeBetweenBeginEndInMs + 1, "Incorrect duration. Collected " + remoteDependencyTelemetryActual.Duration.TotalMilliseconds + " should be less than max " + timeBetweenBeginEndInMs);
```
SergeyKanzhelev: is it only applicable to `OpenAsync`? Why it worked before?

iusafaro: Race condition - it's complicated to write correct test that behaves independent from time spent on tasks initialization. All tests pass when running individually, but sometimes fail when running all of them together...

PR #304, opened 2017-02-01
Unit test fix for QuickPulse

tokaplan:


Src/PerformanceCollector/Unit.Tests/QuickPulse/QuickPulseDataSampleTests.cs
```diff
@@ -11,10 +11,12 @@
     [TestClass]
     public class QuickPulseDataSampleTests
     {
+        private readonly DateTimeOffset now = new DateTimeOffset(2017, 1, 1, 0, 0, 0, 0, TimeSpan.Zero);
```
dnduffy: nitpick: I suggest calling it fakeNow so no one gets confused in future thinking it is really "now".

tokaplan: It is "now" for the unit test :) The time is synthetic.

Issue #168, opened 2016-10-04
RoleInstance for Cloud Services sometimes populated with RoleName_IN_0 and sometimes with RD003FF...


zakimaksyutov:
RoleInstance for Cloud Services sometimes populated with RoleName_IN_0 and sometimes with RD003FF...

FrictionFree:
Note: We should maintain consistency between how the .net sdk logs role instance names and how WAD2AI reports the names of the same instances. Otherwise, we get logs and perf counters coming from the same nodes but appearing as it the nodes are different. Currently there is no consistency.

SergeyKanzhelev:
The reason of inconsistency is that 
1. there is no reliable way to detect that application is running in Cloud Service
2. correctly load the same version of `ServiceRuntime.dll` as customer will attempt to load later

So SDK rely on customer code to load service runtime before calling our initializers. And in most cases it's true as the code will look like [this](https://github.com/Microsoft/ApplicationInsights-Home/blob/5f9cb502bf115777c1bbe4d6c3143901216a35cd/Samples/AzureEmailService/MvcWebRole/Global.asax.cs#L27):

``` csharp
TelemetryConfiguration.Active.InstrumentationKey = RoleEnvironment.GetConfigurationSettingValue("APPINSIGHTS_INSTRUMENTATIONKEY"); 
```

The idea of fixing this issue was to attempt to load Service Runtime into a separate AppDomain and then unload this domain. This approach should minimize an interference with the customer code.

PR #305, opened 2017-02-02
AzureRoleEnvironmentTelemetryInitializer modified to load necessary assembly and read properties like RoleInstance name in a deterministic way

cijothomas:
Fix: https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/168

Current logic:
Lazily wait to see customer app code has loaded ServiceRunTime assembly, if yes use reflection and read Azure properites like RoleInstance name etc.

New logic:
During initialize itself, ServiceRunTime assembly is loaded into a new temporary AppDomain, and all required properties are read and saved. This temp AppDomain is then unloaded. 
This ensures
1. No interference with customer application code.
2. Deterministic behavior.

(From measurements it is found that this new logic takes around 300milliseconds to do the new AppDomain, load assembly etc..)


Src/WindowsServer/WindowsServer.Shared/Implementation/AssemblyLoader.cs
```diff
@@ -0,0 +1,81 @@
...
+        public string[] VersionsToAttempt;
+
+        public bool ReadAndPopulateContextInformation(ref string roleName, ref string roleInstanceId)
+        {
+            Assembly loadedAssembly = null;
+            try
+            {
+                // As this is executed inside a separate AppDomain, it is safe to load assemblies here without interfering with user code.                
+                loadedAssembly = this.AttemptToLoadAssembly(this.AssemblyName, this.Culture, this.PublicKeyToken, this.VersionsToAttempt);
+                if (loadedAssembly != null)
```
cijothomas: Instead of writing the logic of 'what to do with the loaded assembly' here, it could be accepted as a function parameter, so as to improve testability...

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureRoleEnvironmentContextReader.cs
```diff
@@ -45,6 +48,27 @@ public static IAzureRoleEnvironmentContextReader Instance
...
+
+                if (string.IsNullOrEmpty(AzureRoleEnvironmentContextReader.PublicKeyToken) == true)
+                {
+                    AzureRoleEnvironmentContextReader.PublicKeyToken = "31bf3856ad364e35";
+                }
+
+                if (AzureRoleEnvironmentContextReader.VersionsToAttempt == null || AzureRoleEnvironmentContextReader.VersionsToAttempt.Length == 0)
+                {
+                    // This list should be explicity updated when 3.* versions of Microsoft.WindowsAzure.ServiceRuntime are released after confirming they don't have breaking API changes.
+                    AzureRoleEnvironmentContextReader.VersionsToAttempt = new string[] { "2.7.0.0", "2.6.0.0", "2.5.0.0", "2.4.0.0", "2.3.0.0", "2.2.0.0", "2.1.0.0", "2.8.0.0", "2.9.0.0" };
```
cijothomas: need to see if this list is correct...

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureRoleEnvironmentContextReader.cs
```diff
@@ -62,25 +86,76 @@ internal set
...
+                // Load the RemoteWorker assembly to the new domain            
+                tempDomainToLoadAssembly.Load(typeof(AssemblyLoader).Assembly.FullName);
+
+                // Any method invoked on this object will be executed in the newly created AppDomain.
+                AssemblyLoader remoteWorker = (AssemblyLoader)tempDomainToLoadAssembly.CreateInstanceAndUnwrap(typeof(AssemblyLoader).Assembly.FullName, typeof(AssemblyLoader).FullName);
+                remoteWorker.AssemblyName = AzureRoleEnvironmentContextReader.AssemblyName;
+                remoteWorker.Culture = AzureRoleEnvironmentContextReader.Culture;
+                remoteWorker.PublicKeyToken = AzureRoleEnvironmentContextReader.PublicKeyToken;
+                remoteWorker.VersionsToAttempt = AzureRoleEnvironmentContextReader.VersionsToAttempt;
+                bool success = remoteWorker.ReadAndPopulateContextInformation(ref this.roleName, ref this.roleInstanceName);
```
SergeyKanzhelev: you need to check whether it's emulator somewhere. In Emulator we may need to implement different behavior. Depending on what emulator returns as RoleName and RoleInstance we may want to not use provided values

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureRoleEnvironmentContextReader.cs
```diff
@@ -62,25 +86,76 @@ internal set
...
+            finally
+            {                
+                try
+                {
+                    if (tempDomainToLoadAssembly != null)
                     {
-                        this.roleName = role.Name;
-                    }
+                        AppDomain.Unload(tempDomainToLoadAssembly);
+                        long endTimeInTicks = Stopwatch.GetTimestamp();
```
SergeyKanzhelev: what do you plan to use timeout for?

Src/WindowsServer/WindowsServer.Shared/Implementation/AssemblyLoader.cs
```diff
@@ -0,0 +1,81 @@
...
+    // AssemblyLoader contains logic to load Microsoft.WindowsAzure.ServiceRuntime assembly and read properties using reflection
+    // Inherits MarshalByRefObject so that methods of this class can be executed remotely in separate AppDomain.
+    internal class AssemblyLoader : MarshalByRefObject
+    {
+        public string AssemblyName;
+        public string Culture;
+        public string PublicKeyToken;
+        public string[] VersionsToAttempt;
+
+        public bool ReadAndPopulateContextInformation(ref string roleName, ref string roleInstanceId)
```
SergeyKanzhelev: should they be `out`, not `ref`? It will ensure you set at least some value to those strings

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureRoleEnvironmentContextReader.cs
SergeyKanzhelev: will you?

SergeyKanzhelev: Here is the list I believe: https://azure.microsoft.com/en-us/downloads/archive-net-downloads/

SergeyKanzhelev: Maybe you can have a list verion of DLL -> version of SDK.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureRoleEnvironmentContextReader.cs
cijothomas: This is the RoleInstance value returned when running under Emulator : deployment29(59).AzureCloudService2.WebRole1_IN_0
For RoleName it returned WebRole1.

I think its Okay to just use it as it is - people would be using separate ikeys for their tests (emulator ) vs prod.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureRoleEnvironmentContextReader.cs
cijothomas: If you meant the time measurement, its not used for any functionality. Its just used in logging messages to see how much it took for the initialize to complete.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureRoleEnvironmentContextReader.cs
```diff
@@ -62,25 +86,76 @@ internal set
...
+
+                // Create a new AppDomain
+                tempDomainToLoadAssembly = AppDomain.CreateDomain(tempDomainName, null, domaininfo);
+
+                // Load the RemoteWorker assembly to the new domain            
+                tempDomainToLoadAssembly.Load(typeof(AssemblyLoader).Assembly.FullName);
+
+                // Any method invoked on this object will be executed in the newly created AppDomain.
+                AssemblyLoader remoteWorker = (AssemblyLoader)tempDomainToLoadAssembly.CreateInstanceAndUnwrap(typeof(AssemblyLoader).Assembly.FullName, typeof(AssemblyLoader).FullName);
+                remoteWorker.AssemblyName = AzureRoleEnvironmentContextReader.AssemblyName;
```
cijothomas: Along with domain, I made assembly details also as parameters so as to unit test.

Src/WindowsServer/WindowsServer.Shared/Implementation/AssemblyLoader.cs
```diff
@@ -0,0 +1,81 @@
+﻿namespace Microsoft.ApplicationInsights.WindowsServer.Implementation
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Globalization;
+    using System.Reflection;
+    using System.Text;
+
+    // AssemblyLoader contains logic to load Microsoft.WindowsAzure.ServiceRuntime assembly and read properties using reflection
```
cijothomas: Added summary tag

Src/WindowsServer/WindowsServer.Shared/Implementation/AssemblyLoader.cs
cijothomas: they are both instance variables, and are assigned string.Empty by default.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureRoleEnvironmentContextReader.cs
```diff
@@ -62,25 +86,76 @@ internal set
         /// </summary>
         internal static string BaseDirectory { get; set; }
 
+        internal static string AssemblyName { get; set; }
```
cijothomas: We don't know what the fully qualified name of assembly will be yet - so whats the advantage of wrapping the string inside AssemblyName class?

SergeyKanzhelev: Advantage is a single class instead of multiple properties

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureRoleEnvironmentContextReader.cs
SergeyKanzhelev: Will it only return it under emulator? Or it will return it for a regular ASP.NET app as well?

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureRoleEnvironmentContextReader.cs
SergeyKanzhelev: do we need it then? Let's not do anything we do not use

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureRoleEnvironmentContextReader.cs
```diff
@@ -62,25 +86,76 @@ internal set
...
         /// <summary>
         /// Initializes the current reader with respect to its environment.
         /// </summary>
         public void Initialize()
         {
-            ServiceRuntime serviceRuntime = new ServiceRuntime();
-            RoleEnvironment roleEnvironment = serviceRuntime.GetRoleEnvironment(AzureRoleEnvironmentContextReader.BaseDirectory);
+            AppDomain tempDomainToLoadAssembly = null;
+            string tempDomainName = "AppInsightsDomain-" + Guid.NewGuid().ToString().Substring(0, 6);
+            long beginTimeInTicks = Stopwatch.GetTimestamp();
```
cijothomas: Correct. Will change.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureRoleEnvironmentContextReader.cs
cijothomas: only under emulator. For regular runs , IsAvailable() returns false indicating not running on Azure, and we don't populate role instance/rolename from this initializer

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureRoleEnvironmentContextReader.cs
cijothomas: ok got it.

cijothomas:
TODO: Add a check to see if this is Azure WebSite at the beginning - if yes, just return without trying anything futher.

Src/WindowsServer/WindowsServer.Shared/Implementation/AssemblyLoader.cs
cijothomas: will change to out to enforce this.

Src/WindowsServer/WindowsServer.Shared.Tests/AzureRoleEnvironmentTelemetryInitializerTest.cs
```diff
@@ -1,50 +1,32 @@
...
-            Assert.Equal(expectedRoleInstanceName, telemetryItem.Context.Cloud.RoleInstance);
-            Assert.Equal(expectedRoleInstanceName, telemetryItem.Context.GetInternalContext().NodeName);
-        }
-
+    {        
         [TestMethod]
         public void AzureRoleEnvironmentTelemetryInitializerDoesNotOverrideRoleName()
         {
-            var telemetryItem = new EventTelemetry();
+            var telemetryItem = new EventTelemetry();
```
dnduffy: nitpick: seems like there are a bunch of spaces on the end of all these lines as the only change, I think VS might have an option to strip trailing whitespace but I don't recall.

Src/WindowsServer/WindowsServer.Shared/Implementation/AssemblyLoader.cs
```diff
@@ -0,0 +1,90 @@
...
+            }
+            catch (Exception ex)
+            {
+                WindowsServerEventSource.Log.TroubleshootingMessageEvent("Unknown error occured attempting to populate Azure Context, Exception: " + ex.ToString());
+            }
+
+            return loadedAssembly != null;
+        }
+
+        private Assembly AttemptToLoadAssembly(string assemblyName, string culture, string publicKeyToken, string[] versionsToAttempt)
```
SergeyKanzhelev: I'd recommend again to use AssemblyName class. It knows how to format itself (you can pass the class itself to the `Assembly.Load`. One you'll switch to use of that class - this entire method will become redundant I believe

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureRoleEnvironmentContextReader.cs
```diff
@@ -62,25 +65,97 @@ internal set
...
+
         /// <summary>
         /// Initializes the current reader with respect to its environment.
         /// </summary>
         public void Initialize()
         {
-            ServiceRuntime serviceRuntime = new ServiceRuntime();
-            RoleEnvironment roleEnvironment = serviceRuntime.GetRoleEnvironment(AzureRoleEnvironmentContextReader.BaseDirectory);
+            AppDomain tempDomainToLoadAssembly = null;
+            string tempDomainName = "AppInsightsDomain-" + Guid.NewGuid().ToString().Substring(0, 6);
```
SergeyKanzhelev: forgot to ask - why `substring(0,6)`? Did you hit some limitations in appdomain name length?

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureRoleEnvironmentContextReader.cs
```diff
@@ -62,25 +65,97 @@ internal set
...
                 {
-                    this.roleInstanceName = roleEnvironment.CurrentRoleInstance.Id;
-                    Role role = roleInstance.Role;
-                    if (role != null)
+                    if (tempDomainToLoadAssembly != null)
                     {
-                        this.roleName = role.Name;
-                    }
+                        AppDomain.Unload(tempDomainToLoadAssembly);                                                
+                        WindowsServerEventSource.Log.TroubleshootingMessageEvent(tempDomainName + " AppDomain  Unloaded.");
```
SergeyKanzhelev: It is a bad pattern using the EventSource. Single method and string formatting inside is bad practice. Please split into mmultiple methods. Or at least have a `TempraryAppDomainTroubleshooint` method that accepts the domain name as an argument

Src/WindowsServer/WindowsServer.Shared/Implementation/RuntimeBindingObject.cs
```diff
@@ -51,6 +58,15 @@ protected internal object TargetObject
         }
 
         /// <summary>
+        /// Gets or sets the target object.
+        /// </summary>
+        protected internal Assembly LoadedAssembly
```
SergeyKanzhelev: `TargetAssembly`? Assembly class implies that it is already loaded

Src/WindowsServer/WindowsServer.Shared/Implementation/AssemblyLoader.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            Assembly loadedAssembly = null;
+            try
+            {
+                // As this is executed inside a separate AppDomain, it is safe to load assemblies here without interfering with user code.                
+                loadedAssembly = this.AttemptToLoadAssembly(this.AssemblyName, this.Culture, this.PublicKeyToken, this.VersionsToAttempt);
+                if (loadedAssembly != null)
+                {
+                    ServiceRuntime serviceRuntime = new ServiceRuntime();
+                    RoleEnvironment roleEnvironment = serviceRuntime.GetRoleEnvironment(loadedAssembly);
```
SergeyKanzhelev: should you pass assembly class to `ServiceRuntime` constructor, not to `GetRoleEnvironment` logically?

Src/WindowsServer/WindowsServer.Shared/Implementation/RuntimeBindingObject.cs
cijothomas: typo from copy paste. Corrected.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureRoleEnvironmentContextReader.cs
cijothomas: Agreed.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureRoleEnvironmentContextReader.cs
cijothomas: I wasn't sure if appdomain name has any length restriction so I played safe. But it looks like full Guid length can be used - will remove substringing.

Src/WindowsServer/WindowsServer.Shared/Implementation/AssemblyLoader.cs
cijothomas: Yes that seems more correct. Will modify.

Src/WindowsServer/WindowsServer.Shared.Tests/AzureRoleEnvironmentTelemetryInitializerTest.cs
```diff
@@ -69,33 +47,83 @@ public void AzureRoleEnvironmentTelemetryInitializerDoesNotOverrideRoleInstance(
...
+        [Description("Validates that requested DLL was loaded into separate AppDomain and not to the current domain.")]
+        public void AzureRoleEnvironmentTelemetryInitializerLoadDllToSeparateAppDomain()
+        {
+            // A random dll which is not already loaded to the current AppDomain but dropped into bin folder.
+            string dllPath = Path.Combine(AppDomain.CurrentDomain.SetupInformation.ApplicationBase, "Newtonsoft.Json.dll");
+                        
+            try
+            {
+                // Publish the dll to GAC to give a chance for  AzureRoleEnvironmentTelemetryInitializer to load it to a new AppDomain
+                new System.EnterpriseServices.Internal.Publish().GacInstall(dllPath);
```
dnduffy: Do we have to GAC it in order to sandbox it?

Src/WindowsServer/WindowsServer.Shared/Implementation/AssemblyLoader.cs
cijothomas: modified .

Src/WindowsServer/WindowsServer.Shared.Tests/AzureRoleEnvironmentTelemetryInitializerTest.cs
```diff
@@ -69,33 +47,83 @@ public void AzureRoleEnvironmentTelemetryInitializerDoesNotOverrideRoleInstance(
...
+                {
+                    const long AllKeyword = -1;
+                    listener.EnableEvents(WindowsServerEventSource.Log, EventLevel.Verbose, (EventKeywords)AllKeyword);
+
+                    // Create initializer - this will internally create separate appdomain and load assemblies into it.
+                    AzureRoleEnvironmentContextReader.BaseDirectory = AppDomain.CurrentDomain.SetupInformation.ApplicationBase;
+                    AzureRoleEnvironmentContextReader.AssemblyName = "Newtonsoft.Json";
+                    AzureRoleEnvironmentContextReader.Culture = "neutral";
+                    AzureRoleEnvironmentContextReader.PublicKeyToken = "30ad4fe6b2a6aeed";
+                    AzureRoleEnvironmentContextReader.VersionsToAttempt = new string[] { "2.7.0.0", "8.0.0.0" };
```
dnduffy: 9?

Src/WindowsServer/WindowsServer.Shared.Tests/WindowsServer.Shared.Tests.projitems
```diff
@@ -21,10 +21,14 @@
...
     <Compile Include="$(MSBuildThisFileDirectory)DeviceTelemetryInitializerTest.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Implementation\ServiceRuntimeTests.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)Implementation\TypeHelpersTests.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Implementation\WindowsServerEventSourceTests.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)FirstChanceExceptionStatisticsTelemetryModuleTest.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)UnhandledExceptionTelemetryModuleTest.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)UnobservedExceptionTelemetryModuleTest.cs" />
   </ItemGroup>
+  <ItemGroup>
+    <Content Include="$(MSBuildThisFileDirectory)Newtonsoft.Json.dll">
```
dnduffy: What is the significance of this DLL?  Why are we using this 3rd party lib for testing?

Src/WindowsServer/WindowsServer.Shared.Tests/AzureRoleEnvironmentTelemetryInitializerTest.cs
cijothomas: AzureServiceRuntimeAssemblyLoader has the ability to load assembly from GAC only now. So test is putting a dll in the GAC for testing.

Src/WindowsServer/WindowsServer.Shared.Tests/WindowsServer.Shared.Tests.projitems
cijothomas: nothing spl about this dll - testing the ability to load any given DLL from GAC to a new app domain - so i simply used a random (newtonsoft.json) dll.

Src/WindowsServer/WindowsServer.Shared.Tests/AzureRoleEnvironmentTelemetryInitializerTest.cs
SergeyKanzhelev: Mark test as "Inconclusive" if failed to add to GAC. I never run this solutions "As Administrator"

Src/WindowsServer/WindowsServer.Shared/Implementation/WindowsServerEventSource.cs
```diff
@@ -135,6 +135,90 @@ public void UnobservedTaskExceptionThrewUnhandledException(string exception, str
...
+        public void AzureRoleEnvironmentContextReaderAppDomainTroubleshoot(string domainName, string msg, string applicationName = "dummy")
+        {
+            this.WriteEvent(22, domainName, msg, this.ApplicationName);
+        }
+
+        [Event(
+           23,
+           Message = "AzureRoleEnvironmentContextReader initialization took {0} msec.",
+           Level = EventLevel.Informational)]
+        public void AzureRoleEnvironmentContextReaderInitializationDuration(long durationMsec, string applicationName = "dummy")
```
SergeyKanzhelev: `dummy` is a bad name. policheck may not like it

cijothomas: replaced with unknown.

Src/WindowsServer/WindowsServer.Shared.Tests/AzureRoleEnvironmentTelemetryInitializerTest.cs
cijothomas: @SergeyKanzhelev No Inconcluseive option in the Assert class for XUnit

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureServiceRuntimeAssemblyLoader.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        public AzureServiceRuntimeAssemblyLoader()
+        {
+            this.AssemblyNameToLoad.Name = "Microsoft.WindowsAzure.ServiceRuntime";
+            this.AssemblyNameToLoad.CultureInfo = CultureInfo.InvariantCulture;
+            this.AssemblyNameToLoad.SetPublicKeyToken(new byte[] { 49, 191, 56, 86, 173, 54, 78, 53 });
+            this.VersionsToAttempt = new Version[] { new Version("2.7.0.0"), new Version("2.6.0.0"), new Version("2.5.0.0") };
+        }
+
+        public virtual bool ReadAndPopulateContextInformation(out string roleName, out string roleInstanceId)
```
dnduffy: nitpick: Please comment new code.

Src/WindowsServer/WindowsServer.Shared/AzureRoleEnvironmentTelemetryInitializer.cs
```diff
@@ -30,22 +48,41 @@ public AzureRoleEnvironmentTelemetryInitializer()
...
             {
-                var name = LazyInitializer.EnsureInitialized(ref this.roleInstanceName, AzureRoleEnvironmentContextReader.Instance.GetRoleInstanceName);
-                telemetry.Context.GetInternalContext().NodeName = name;
+                try
+                {
+                    isAzureWebApp = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable(WebSiteEnvironmentVariable));
+                }
+                catch (Exception ex)
+                {
+                    WindowsServerEventSource.Log.AccessingEnvironmentVariableFailedWarning(WebSiteEnvironmentVariable, ex.ToString());
```
dnduffy: I think we should switch to PlatformImplementation.AccessingEnvironmentVariableFailedWarning and make it public.

PR #306, opened 2017-02-02
Putting instance name, machine name, and stream id into ping's header…

tokaplan:
…s for QPS's convenience. Allows QPS not to read the content of every Ping request.

### DIFF
--- a/Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseConstants.cs
+++ b/Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseConstants.cs
@@ -14,5 +14,20 @@
         /// Transmission time header.
         /// </summary>
         internal const string XMsQpsTransmissionTimeHeaderName = "x-ms-qps-transmission-time";
+
+        /// <summary>
+        /// Instance name header.
+        /// </summary>
+        internal const string XMsQpsInstanceNameHeaderName = "x-ms-qps-instance-name";
+
+        /// <summary>
+        /// Stream id header.
+        /// </summary>
+        internal const string XMsQpsStreamIdHeaderName = "x-ms-qps-stream-id";
+
+        /// <summary>
+        /// Machine name header.
+        /// </summary>
+        internal const string XMsQpsMachineNameHeaderName = "x-ms-qps-machine-name";
     }
 }
--- a/Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
+++ b/Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
@@ -51,7 +51,7 @@
         public bool? Ping(string instrumentationKey, DateTimeOffset timestamp)
         {
             var path = string.Format(CultureInfo.InvariantCulture, "ping?ikey={0}", Uri.EscapeUriString(instrumentationKey));
-            HttpWebResponse response = this.SendRequest(WebRequestMethods.Http.Post, path, stream => this.WritePingData(timestamp, stream));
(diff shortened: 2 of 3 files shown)


Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
```diff
@@ -204,7 +205,7 @@ private void WriteSamples(IEnumerable<QuickPulseDataSample> samples, string inst
             this.serializerDataPointArray.WriteObject(stream, monitoringPoints.ToArray());
         }
 
-        private HttpWebResponse SendRequest(string httpVerb, string path, Action<Stream> onWriteBody)
+        private HttpWebResponse SendRequest(string httpVerb, string path, bool includeHeaders, Action<Stream> onWriteBody)
```
zakimaksyutov: Do we need optimize and not send this info for posts? Don't think it is a huge overhead. I'd rather simplify the logic here and send them always.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseConstants.cs
```diff
@@ -14,5 +14,20 @@ internal class QuickPulseConstants
...
+
+        /// <summary>
+        /// Instance name header.
+        /// </summary>
+        internal const string XMsQpsInstanceNameHeaderName = "x-ms-qps-instance-name";
+
+        /// <summary>
+        /// Stream id header.
+        /// </summary>
+        internal const string XMsQpsStreamIdHeaderName = "x-ms-qps-stream-id";
```
SergeyKanzhelev: what is stream id?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseServiceClient.cs
tokaplan: Well we read Post's body anyway, so I figured it's not that complicated :)

Src/PerformanceCollector/Unit.Tests/QuickPulse/QuickPulseServiceClientTests.cs
```diff
@@ -23,7 +23,7 @@ public sealed class QuickPulseServiceClientTests : IDisposable
 
         private readonly List<Tuple<DateTimeOffset, MonitoringDataPoint>> samples = new List<Tuple<DateTimeOffset, MonitoringDataPoint>>();
 
-        private readonly List<Tuple<DateTimeOffset, MonitoringDataPoint>> pings = new List<Tuple<DateTimeOffset, MonitoringDataPoint>>();
+        private readonly List<Tuple<PingHeaders, MonitoringDataPoint>> pings = new List<Tuple<PingHeaders, MonitoringDataPoint>>();
```
SergeyKanzhelev: Are those headers changes much? Could they be set once in constructor or initialization of sender?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseConstants.cs
tokaplan: A GUID that uniquely identifies a QuickPulse client. Even when there are multiple application instances hosted on the same machine, stream id will be unique for each of them.

Src/PerformanceCollector/Unit.Tests/QuickPulse/QuickPulseServiceClientTests.cs
tokaplan: This is unit test, convenience is king :)

PR #308, opened 2017-02-03
Fix a randomly failing FunctionalTest

cijothomas:
Catch and trace exception if killing IISExpress process fails instead of failing test as a whole with unhandled exception.

(Perfcounter module already does this and has been running fine. Web func tests did not handle this exception and has been failing randomnly)

### DIFF
--- a/Test/Web/FunctionalTests/FunctionalTests/IisExpress/IISExpress.cs
+++ b/Test/Web/FunctionalTests/FunctionalTests/IisExpress/IISExpress.cs
@@ -94,7 +94,14 @@
             if (false == this.hostProcess.WaitForExit(processExitWaitTimeout))
             {
                 Trace.TraceWarning("iisexpress.exe process hasn't exited during expected time, terminating!");
-                this.hostProcess.Kill();
+                try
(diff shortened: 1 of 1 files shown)


Test/Web/FunctionalTests/FunctionalTests/IisExpress/IISExpress.cs
```diff
@@ -94,7 +94,14 @@ public void Stop()
             if (false == this.hostProcess.WaitForExit(processExitWaitTimeout))
             {
                 Trace.TraceWarning("iisexpress.exe process hasn't exited during expected time, terminating!");
-                this.hostProcess.Kill();
+                try
+                {
+                    this.hostProcess.Kill();
+                }
+                catch (Win32Exception e)
```
cijothomas: If the process is already in the middle is exiting, it is guaranteed to throw a Security Exception.
From MSDN:
"_If the call to the Kill method is made while the process is currently terminating, a Win32Exception is thrown for Access Denied_."

Instead of increasing our timeout further, its better to catch

Issue #248, opened 2016-12-02
No way to know which environment generated event in Azure Web Sites


SergeyKanzhelev:
In beta4 we used full name with the "slot" name as instance name:

![image](https://cloud.githubusercontent.com/assets/9950081/20853062/5d9dffb6-b89e-11e6-9c22-8bd9c60eeda5.png)

In beta6 we are using hash code of an instance and kept role name to be the role name of the "main" site:

![image](https://cloud.githubusercontent.com/assets/9950081/20853081/81319c1c-b89e-11e6-9ec2-157efd921abe.png)

So now there is no way to know which slot dependency was sent from. Consider adding a custom property or change the role name property

cijothomas:
We might need to use %WEBSITE_HOSTNAME% for Rolename instead of WEBSITE_SITE_NAME, as the former appends the 'slot' name to the sitename.
Following output from production vs dev slot of my webapp

> echo %WEBSITE_HOSTNAME%
D:\home\site\wwwroot
cijo22beta6-dev.azurewebsites.net

> echo %WEBSITE_HOSTNAME%
D:\home\site\wwwroot
cijo22beta6.azurewebsites.net

cijothomas:
%WEBSITE_HOSTNAME% combined with WEBSITE_INSTANCE_ID would be needed. Because HostName alone will not be able to distinguish between instances, and instance_id can't distinguish slots. So a combination should be used.

dnduffy:
Fixed: https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/250

SergeyKanzhelev:
Ok, the fix was applied to send slot as a nodeName. However nodeName is not exposed neither in UI nor in analytics. So it wasn't really fixed. Consider sending it as a role name instead of a host name

Also `SCM_instance_id` is not used any more. What's the reason? Using instance ID as Azure Web Apps understand it can help correlate telemetry with the azure Web Site telemetry. You can even send a ping to the specific instance of Azure Web Site using arr cookies. Computer name doesn't look like a meaningful choice for Azure Web Apps

SergeyKanzhelev:
@iusafaro can you please comment on this?

iusafaro:
[...]

PR #314, opened 2017-02-06
Issue #248 Azure Web App - update role name to contain the slot

iusafaro:


Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
```diff
@@ -13,13 +13,13 @@
...
     {
-        /// <summary>Azure Web App name corresponding to the resource name.</summary>
-        private const string WebAppNameEnvironmentVariable = "WEBSITE_SITE_NAME";
-
         /// <summary>Azure Web App Hostname. This will include the deployment slot, but will be same across instances of same slot.</summary>
         private const string WebAppHostNameEnvironmentVariable = "WEBSITE_HOSTNAME";
 
-        private string roleInstanceName;
+        /// <summary>Predefined suffix for Azure Web App Hostname.</summary>
+        private const string WebAppSuffix = ".azurewebsites.net";
```
dnduffy: Does this still work if they use a custom domain?
I'm not sure if the default Azure one still exists, or is at least usable, when users/companies use their own URL.
Assuming it is available, it may not be what they are expecting to see or search for in the portal/data.

dnduffy: Oh I see below, it's only relevant when this default name is in use, otherwise the full domain name will remain.

Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
```diff
@@ -44,14 +44,20 @@ public void Initialize(ITelemetry telemetry)
...
+                string name = LazyInitializer.EnsureInitialized(ref this.nodeName, this.GetNodeName);                
                 telemetry.Context.GetInternalContext().NodeName = name;
             }
         }
 
         private string GetRoleName()
         {
-            return Environment.GetEnvironmentVariable(WebAppNameEnvironmentVariable) ?? string.Empty;
+            var result = this.GetNodeName();
+            if (result.ToLowerInvariant().EndsWith(WebAppSuffix))
```
pharring: Nit: EndsWith has an overload which takes a StringComparison. So you could do:

```result.EndsWith(webAppSuffix, StringComparison.InvariantCultureIgnoreCase)```

dnduffy: Thanks, our build caught this and we've fixed it.

PR #329, opened 2017-02-17
Replace a test dll with its signed version to avoid StrongName verifi…

cijothomas:
…cation failed.


SergeyKanzhelev:
Can you get it from NuGet? Let's not check in binaries into repo

cijothomas:
@SergeyKanzhelev we just need a dll file for testing. No advantage in using nuget to obtain a .dll for testing which is not expected to change. I prefer the simplicity of just checking-in a dll in repo.

SergeyKanzhelev:
@cijothomas no, please do not check in binaries.

cijothomas:
I understand its a bad idea to check-in large dlls into a git repo - but this is a 20KB file, for which we dont care for any git feature like history etc. since the test code does not care whats inside the dll. It simply loads the dll and unloads.

I don't see any problem having a 20KB test binary in the repository. If you have a strong reason please let me know.

SergeyKanzhelev:
Yes, I do have a strong reason. It is not hard to take a binary from the NuGet and I do not see a reason to take a shortcut here

dnduffy:
I think getting it from NuGet for the test would be a whole bunch of extra code and obscurity that isn't worth it.

PR #332, opened 2017-02-17
Dependency Functional test fixes

cijothomas:
Modify dependency functions to use a url which wont fail at DNS resolution stage for simulating failed case scenarios. This will ensure result code is collected.
Modified validation logic to validate result code
Added new test to simulate failure even before reaching server by using non existent DNS and validate that result code is not populated.


cijothomas:
The binaryfile from my previous PR is part of this PR by accident (when i created branch).
Won't merge until the solution is agreed.

SergeyKanzhelev:
can we set result code to "Dns Failure" in case of profiler? Maybe based on exception type?

Test/DependencyCollector/FunctionalTests/FW40Shared/HttpHelper40.cs
```diff
@@ -29,9 +29,14 @@ namespace FW40Shared
     public class HttpHelper40
     {
         /// <summary>
-        /// Invalid Hostname to trigger exception being thrown
+        /// Invalid endpoint to trigger exception being thrown
         /// </summary>
-        private const string InvalidHostName = "http://www.zzkaodkoakdahdjghejajdnad.com";
+        private const string InvalidHostName = "http://google.com/404";
```
SergeyKanzhelev: this is valid host name. Why variable called `InvalidHostName`?

Test/DependencyCollector/FunctionalTests/FW45Shared/HttpHelper45.cs
```diff
@@ -51,7 +51,7 @@ public class HttpHelper45
         {
             for (int i = 0; i < count; i++)
             {
-                Uri ourUri = new Uri("http://www.zzkaodkoakdahdjghejajdnad.com");
+                Uri ourUri = new Uri("http://google.com/404");
```
SergeyKanzhelev: you just defined a constant for it

Test/DependencyCollector/FunctionalTests/FuncTest/Helpers/DeploymentAndValidationTools.cs
```diff
@@ -197,11 +197,17 @@ public static void CleanUp()
         public static void Validate(
             TelemetryItem<RemoteDependencyData> itemToValidate,
             TimeSpan accessTimeMax,
-            bool successFlagExpected)
+            bool successFlagExpected,
+            string resultCodeExpected = "DontCheck")
```
SergeyKanzhelev: can you just pass `string.Empty`?

Test/DependencyCollector/FunctionalTests/FW45Shared/HttpHelper45.cs
cijothomas: repaced with constant.

Test/DependencyCollector/FunctionalTests/FW40Shared/HttpHelper40.cs
cijothomas: renamed to avoid the confusion

Test/DependencyCollector/FunctionalTests/FuncTest/Helpers/DeploymentAndValidationTools.cs
cijothomas: string.Empty will be the value for clientside exception, so can't just pass that as default.

cijothomas:
@SergeyKanzhelev Yes its a nice improvement to populate resultcode to indicate if the DependencyFailure was due to DNSResolution failure - i'll open separate issue to add that feature. 

**This PR is purely modify the functional tests to check for result code and NO product change.**

cijothomas:
Open issue to improve profile dependency collection for clientside errors.
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/335

Test/DependencyCollector/FunctionalTests/FuncTest/Helpers/DeploymentAndValidationTools.cs
SergeyKanzhelev: Can you clarify this? Do we have tests that do not check the result code? If so - maybe it's a good idea to modify every test?

Test/DependencyCollector/FunctionalTests/FW40Shared/HttpHelper40.cs
```diff
@@ -29,9 +29,14 @@ namespace FW40Shared
     public class HttpHelper40
     {
         /// <summary>
-        /// Invalid Hostname to trigger exception being thrown
+        /// Invalid endpoint to trigger exception being thrown
         /// </summary>
-        private const string InvalidHostName = "http://www.zzkaodkoakdahdjghejajdnad.com";
+        public const string UrlWhichThrowException = "http://google.com/404";
```
dnduffy: nitpick: Throws

PR #333, opened 2017-02-18
Add a .NET Core DependencyCollector implementation

daschult:


Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,171 @@
...
+
+        /// <summary>
+        /// Standard parent Id header.
+        /// </summary>
+        internal const string StandardParentIdHeader = "x-ms-request-id";
+
+        /// <summary>
+        /// Standard root id header.
+        /// </summary>
+        internal const string StandardRootIdHeader = "x-ms-request-root-id";
```
cijothomas: is it possible to get these constants from https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/Common/RequestResponseHeaders.cs instead of re-declaring?

Test/DependencyCollector/FunctionalTests/FuncTest/HttpTests.cs
```diff
@@ -153,212 +154,355 @@ public void MyTestCleanup()
...
         #endregion 451
-                          
+
+        #region Core
+
+        private const string AspxCoreTestAppFolder = "..\\TestApps\\ASPXCore\\App\\";
+
+        private static void EnsureNetCoreInstalled()
+        {
+            //if (!RegistryCheck.IsNetCoreInstalled)
```
cijothomas: this should be uncommented right?

NuGet.config
```diff
@@ -16,7 +16,9 @@
         <add key="repositoryPath" value="..\packages" />
     </config>
   <packageSources>
+    <add key="LocallyBuiltApplicationInsightsPackages" value="..\bin\Debug\NuGet" />
```
dnduffy: I am not sure that this change should get checked in.

Src/DependencyCollector/Microsoft.AI.DependencyCollector.Tests/DependencyCollectorDiagnosticListenerTests.cs
```diff
@@ -0,0 +1,251 @@
...
+            Assert.AreEqual(true, telemetry.Success);
+
+            Assert.AreEqual(InstrumentationKeyHashLookupHelper.GetInstrumentationKeyHash(instrumentationKey), GetRequestHeaderValues(request, DependencyCollectorDiagnosticListener.SourceInstrumentationKeyHeader).SingleOrDefault());
+            Assert.IsFalse(GetRequestHeaderValues(request, DependencyCollectorDiagnosticListener.StandardRootIdHeader).Any());
+            Assert.IsTrue(GetRequestHeaderValues(request, DependencyCollectorDiagnosticListener.StandardParentIdHeader).Any());
+
+            Assert.AreEqual(0, sentTelemetry.Count);
+        }
+
+        private static IEnumerable<string> GetRequestHeaderValues(HttpRequestMessage request, string headerName)
```
dnduffy: You might need to StyleCop the file so the build will pass.

Src/DependencyCollector/Microsoft.AI.DependencyCollector.Tests/DependencyCollectorDiagnosticListenerTests.cs
```diff
@@ -0,0 +1,251 @@
...
+            Assert.AreEqual("www.example.com", telemetry.Target);
+            Assert.AreEqual("404", telemetry.ResultCode);
+            Assert.AreEqual(false, telemetry.Success);
+        }
+
+        [TestMethod]
+        public void OnResponseWithSuccessfulResponseEventWithMatchingRequestAndDifferentTargetInstrumentationKeyHashHeader()
+        {
+            Guid loggingRequestId = Guid.NewGuid();
+            HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Post, "https://www.example.com");
```
dnduffy: nitpick: it might be nice to make constants for the repeated magic strings (e.g. "https://www.example.com").

NuGet.config
daschult: This is actually required to get the AspxCore test application to reference the locally built DependencyCollector NuGet package. I think this won't be necessary after you guys move to VS 2017, but I'm not positive.

Src/DependencyCollector/Microsoft.AI.DependencyCollector.Tests/InstrumentationKeyHashLookupHelperTests.cs
```diff
@@ -0,0 +1,46 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
```
dnduffy: There is no License.txt file in the repo.
Our NuGet packages link to here: [http://go.microsoft.com/fwlink/?LinkID=510709](http://go.microsoft.com/fwlink/?LinkID=510709)

Src/DependencyCollector/Microsoft.AI.DependencyCollector.Tests/DependencyCollectorDiagnosticListenerTests.cs
daschult: Done.

Src/DependencyCollector/Microsoft.AI.DependencyCollector.Tests/InstrumentationKeyHashLookupHelperTests.cs
daschult: Remove.

Src/DependencyCollector/Microsoft.AI.DependencyCollector.Tests/DependencyCollectorDiagnosticListenerTests.cs
daschult: I've built using buildRelease.cmd, which sets StyleCopEnabled to True, so I believe I've gotten this to pass StyleCop.

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
daschult: Unfortunately, no. RequestResponseHeaders is declared internal, which means that its members aren't available to me here, so I have to re-declare these header names.

Test/DependencyCollector/FunctionalTests/FuncTest/HttpTests.cs
daschult: Done.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,189 @@
...
+        /// <summary>
+        /// Standard root id header.
+        /// </summary>
+        internal const string StandardRootIdHeader = "x-ms-request-root-id";
+
+        private readonly ApplicationInsightsUrlFilter applicationInsightsUrlFilter;
+        private readonly TelemetryClient client;
+        private readonly ConcurrentDictionary<Guid, DependencyTelemetry> pendingTelemetry = new ConcurrentDictionary<Guid, DependencyTelemetry>();
+
+        internal DependencyCollectorDiagnosticListener(TelemetryConfiguration configuration)
```
dnduffy: Doc comments, StyleCop, etc.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,189 @@
...
+
+            // Since dependencySource is no longer set, sdk version is prepended with information which can identify whether RDD was collected by profiler/framework
+            // For directly using TrackDependency(), version will be simply what is set by core
+            string versionStr = typeof(DependencyCollectorDiagnosticListener).GetTypeInfo().Assembly.GetCustomAttributes()
+                    .OfType<AssemblyFileVersionAttribute>()
+                    .First()
+                    .Version;
+
+            Version version = new Version(versionStr);
+            this.client.Context.GetInternalContext().SdkVersion = "rddf" + version.ToString(3) + "-" + version.Revision;
```
dnduffy: Isn't there a helper for this already?

daschult: There is, but the implementation wasn't compatible with .NETStandard. I had to write a slightly different implementation here.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,189 @@
...
+                value.SubscribeWithAdapter(this);
+            }
+        }
+
+        public void OnCompleted()
+        {
+        }
+
+        public void OnError(Exception error)
+        {
```
dnduffy: Shouldn't we log the error?

daschult: Every example with DiagnosticListener I've seen doesn't do anything with this method. This method is called if there is an error adding the DiagnosticListener to the list of other DiagnosticListeners, not if there is an error with an HTTP request.

NuGet.config
dnduffy: If that test app is in the same solution then can you do a project reference instead of referencing the dropped package?

Src/DependencyCollector/Microsoft.AI.DependencyCollector.Tests/DependencyCollectorDiagnosticListenerTests.cs
dnduffy: In my experience StyleCop complains about static and non static method order and public and private method order and you've got both together so maybe some StyleCop settings are turned off and aren't in the default state?

Src/DependencyCollector/Microsoft.AI.DependencyCollector.Tests/project.json
```diff
@@ -0,0 +1,30 @@
...
+    "MSTest.TestFramework": "1.0.8-rc2",
+    "Microsoft.AI.DependencyCollector": {
+      "target": "project",
+      "version": "2.3.0-beta2"
+    }
+  },
+
+  "frameworks": {
+    "netcoreapp1.0": {
+      "imports": "dnxcore50"
```
dnduffy: Is dnxcore still needed?  I thought it was obsolete?

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
dnduffy: Can the existing one be tweaked to work for all?  If we change the version reporting scheme it would be good to not have to try to find multiple locations.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
dnduffy: Are the errors from adding listeners logged somewhere else or just "lost"?
I would think we should write the error to the event source so we have some record of the failure and something to investigate.

Test/DependencyCollector/FunctionalTests/FuncTest/Helpers/IISTestWebApplication.cs
```diff
@@ -0,0 +1,69 @@
...
+
+                if (Directory.Exists(this.AppFolder))
+                {
+                    ACLTools.GetEveryoneAccessToPath(this.AppFolder);
+                }
+            }
+            catch (Exception ex)
+            {
+                Trace.TraceError("Exception occured while attempting to deploy application {0}: {1}. Tests will not continue as they are guaranteed to fail.", this.AppName, ex);
+                throw ex;
```
dnduffy: Just "throw;" to preserve the stack trace.

Test/DependencyCollector/FunctionalTests/FuncTest/HttpTests.cs
```diff
@@ -153,212 +155,245 @@ public void MyTestCleanup()
...
+        #region Core
+
+        private static void EnsureDotNetCoreInstalled()
+        {
+            string error = "";
+            DotNetCoreProcess process = new DotNetCoreProcess("--version")
+                .RedirectStandardErrorTo((string errorMessage) => error += errorMessage)
+                .Run();
+
+            if (process.ExitCode.Value != 0 || !string.IsNullOrEmpty(error))
```
dnduffy: Would it be good to also check for a minimum version? (e.g. 1.0.1 or 1.1.0 etc.)

Test/DependencyCollector/FunctionalTests/TestApps/AspxCore/Controllers/ExternalCallsController.cs
```diff
@@ -0,0 +1,127 @@
...
+
+            return $"<HTML><BODY>{title}<BR/>{response}</BODY></HTML>";
+        }
+
+        /// <summary>
+        /// Make sync http GET calls
+        /// </summary>        
+        /// <param name="count">no of GET calls to be made</param>        
+        /// <param name="hostname">the GET call will be made to http://www.hostname.com</param>        
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Usage", "CA2202:Do not dispose objects multiple times", Justification = "Reviewed manually")]
```
dnduffy: Where is this rule being triggered?  I don't see any using statements or explicit Dispose calls...

Test/DependencyCollector/FunctionalTests/TestApps/AspxCore/Startup.cs
```diff
@@ -0,0 +1,54 @@
...
+            services.AddMvc();
+        }
+
+        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline
+        public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
+        {
+            loggerFactory.AddConsole(Configuration.GetSection("Logging"));
+            loggerFactory.AddDebug();
+
+            app.UseApplicationInsightsRequestTelemetry();
```
dnduffy: Obsolete, remove.

Test/DependencyCollector/FunctionalTests/TestApps/AspxCore/Startup.cs
```diff
@@ -0,0 +1,54 @@
...
+
+            app.UseApplicationInsightsExceptionTelemetry();
```
dnduffy: Also obsolete and should be removed.

NuGet.config
daschult: Yes. I think the project folders may need to be in the same parent folder, but I'm not positive about that.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
daschult: Is there a way to turn off StyleCop for my unit tests? I don't think its necessary to write XML documentation for each of my methods (your other unit tests certainly don't).

Src/DependencyCollector/Microsoft.AI.DependencyCollector.Tests/DependencyCollectorDiagnosticListenerTests.cs
daschult: I fixed all of the build warnings (for both debug and release).

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
daschult: I added comments anyways. Easier than digging around trying to figure out how to disable those warnings.

Test/DependencyCollector/FunctionalTests/FuncTest/Helpers/IISTestWebApplication.cs
daschult: Done.

Test/DependencyCollector/FunctionalTests/TestApps/AspxCore/Controllers/ExternalCallsController.cs
daschult: Removed.

Test/DependencyCollector/FunctionalTests/TestApps/AspxCore/Startup.cs
daschult: Removed.

Test/DependencyCollector/FunctionalTests/TestApps/AspxCore/Startup.cs
daschult: Removed.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
daschult: I learned that you can provide build constants in the project.json file for conditional compilation. I've defined NETCORE and then added conditions in the code to use .NET Core specific code in certain places. That helped me to not duplicate other functionality elsewhere (InstrumentationKeyHashLookupHelper, for example).

Test/DependencyCollector/FunctionalTests/TestApps/AspxCore/Startup.cs
daschult: I removed them, and then my functional tests started failing. The instrumentation key wasn't getting initialized into my telemetry objects. Since this is just a test application, I'd suggest that I keep these in here.

Test/DependencyCollector/FunctionalTests/TestApps/AspxCore/Startup.cs
daschult: I removed them, and then my functional tests started failing. The instrumentation key wasn't getting initialized into my telemetry objects. Since this is just a test application, I'd suggest that I keep these in here.

Src/DependencyCollector/Microsoft.AI.DependencyCollector.Tests/project.json
daschult: This was the default that the tool created when I created a new ASP.NET project. I tried to remove it, but that created a build failure. It looks like that allows me to run .NET Core apps on the .NET Core framework.

Test/DependencyCollector/FunctionalTests/FuncTest/HttpTests.cs
daschult: Done.

daschult:
@dnduffy Please take another look. I've responded to all of your comments.

NuGet.config
dnduffy: The solution should have a parent folder with all the child projects in it, that's how it is currently setup.

Src/DependencyCollector/Microsoft.AI.DependencyCollector.Tests/project.json
dnduffy: If you just want .NET Core then I think you want the following:
  "frameworks": {
    "netcoreapp1.0": {
      "dependencies": {
        "Microsoft.NETCore.App": {
          "type": "platform",
          "version": "1.1.0"
        }
      }
    }
  }

Test/DependencyCollector/FunctionalTests/TestApps/AspxCore/Startup.cs
dnduffy: Using the one for exception tracking will give you duplicate exceptions being reported.
The other one is literally an empty method call that does nothing, neither one can possibly have caused what you describe unless you are not using 2.0.0 of the SDK.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
dnduffy: Thank-you :)

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
dnduffy: EventSource is available in .NET Core, we use it extensively in each of the SDK's including the .NET Core SDK.
Each SDK has its own class for messages, the one in the dependency collector project is DependencyCollectorEventSource.

dnduffy: You can see examples in DependencyTrackingTelemetryModule:
e.g. ```DependencyCollectorEventSource.Log.RemoteDependencyModuleError(exc.ToInvariantString(), Environment.Version.ToString());```

NuGet.config
daschult: That's how most solutions/projects are set up, but this one isn't set up that way. The DependencyCollector has two solutions files associated with it (.../Src/Microsoft.ApplicationInsights.Web.sln and .../Test/DependencyCollector/FunctionalTests.sln). The Src solution is the one that contains the implementation and creates the DependencyCollector NuGet package. The Test solution is the one that consumes the NuGet package, and since the folder structure is so different, it can't reference the Src DependencyCollector project.

Src/DependencyCollector/Microsoft.AI.DependencyCollector.Tests/project.json
daschult: If I use the snippet you've put here instead of the "frameworks" section I have in the project.json, then I get restore failures.

Test/DependencyCollector/FunctionalTests/TestApps/AspxCore/Startup.cs
daschult: I was able to remove the first one (app.UseApplicationInsightsRequestTelemetry()) and the functional tests still pass. However, as soon as I remove the second one (app.UseApplicationInsightsExceptionTelemetry()), then my .NET Core tests start failing.
What should the test application be using instead?

daschult:
@dnduffy Please take another look. I've responded to all of your comments.

Src/DependencyCollector/Microsoft.AI.DependencyCollector.Tests/project.json
dnduffy: You have to do a restore from the command line once due to a VS "bug", is that what you are referring to?

Test/DependencyCollector/FunctionalTests/TestApps/AspxCore/Startup.cs
dnduffy: @pakrym Can you provide some insight on why Dan needs this old call for his unit tests to pass?

pakrym: Hosting diagnostic listener should handle exception code path, check if exception gets to it and how it is getting traced.

Src/DependencyCollector/Microsoft.AI.DependencyCollector.Tests/project.json
daschult: No, I always do my restoring and building from the command line. The restore error I got was from a command line build.

PR #339, opened 2017-02-27
Use appId instead of hashed ikey for cross component correlation.

nizarq:
Once the breeze service is available with a query endpoint for appId, we would need to replace the current placeholder stub to use that.


Src/Common/CorelationIdLookupHelper.cs
```diff
@@ -0,0 +1,80 @@
...
+    using System.Globalization;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// A store for instrumentation App Ids. This makes sure we don't query the public endpoint to find an app Id for the same Ikey more than once.
+    /// </summary>
+    internal static class CorelationIdLookupHelper
+    {
+        /// <summary>
+        /// Max number of component hashes to cache.
```
OsvaldoRosado: Nit: Out of date comment

Src/Common/CorelationIdLookupHelper.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// </summary>
+        private const int MAXSIZE = 100;
+
+        private const string CorrelationIdFormat = "aid-v1:{0}";
+
+        private static ConcurrentDictionary<string, string> knownAppIds = new ConcurrentDictionary<string, string>();
+
+        /// <summary>
+        /// Retrieves the hash for a given instrumentation key.
```
OsvaldoRosado: Nit: out of date comment

Src/Common/CorelationIdLookupHelper.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// <param name="correlationId">AppId corresponding to the provided instrumentation key</param>
+        /// <returns>true if correlationId was successfully retrieved; false otherwise.</returns>
+        public static bool TryGetAppId(string instrumentationKey, out string correlationId)
+        {
+            if (string.IsNullOrEmpty(instrumentationKey))
+            {
+                throw new ArgumentNullException("instrumentationKey");
+            }
+
+            var found = knownAppIds.TryGetValue(instrumentationKey, out correlationId);
```
OsvaldoRosado: Why do we need to know more than our own AppId? What is this cache for?

Src/Common/CorelationIdLookupHelper.cs
```diff
@@ -0,0 +1,80 @@
...
+                }
+
+                correlationId = string.Format(CorrelationIdFormat, appId, CultureInfo.InvariantCulture);
+                knownAppIds[instrumentationKey] = correlationId;
+            }
+            return true;
+        }
+
+        /// <summary>
+        /// Computes the SHA256 hash for a given value and returns it in the form of a base64 encoded string.
```
OsvaldoRosado: Nit: out of date comment

Src/Common/RequestResponseHeaders.cs
```diff
@@ -8,12 +8,12 @@ internal static class RequestResponseHeaders
         /// <summary>
         /// Source instrumentation header that is added by an application while making http requests and retrieved by the other application when processing incoming requests.
         /// </summary>
-        public const string SourceInstrumentationKeyHeader = "x-ms-request-source-ikey";
+        public const string SourceAppIdHeader = "x-ms-request-source-appid";
```
OsvaldoRosado: Do we break anything by changing these header names?

Src/Web/Web.Shared.Net.Tests/OperationCorrelationTelemetryInitializerTests.cs
```diff
@@ -135,7 +135,7 @@ public void InitializeDoesNotAddSourceFieldForRequestForSameComponent()
             string hashedIkey = "o05HMrc4Og8W1Jyy60JPDPxxQy3bOKyuaj6HudZHTjE=";
```
OsvaldoRosado: This shouldn't be a hash anymore

Src/Web/Web.Shared.Net/OperationCorrelationTelemetryInitializer.cs
```diff
@@ -90,16 +90,23 @@ public OperationCorrelationTelemetryInitializer()
...
             if (string.IsNullOrEmpty(requestTelemetry.Source) && currentRequest.Headers != null)
             {
-                string sourceIkey = currentRequest.Headers[RequestResponseHeaders.SourceInstrumentationKeyHeader];
+                string sourceAppId = currentRequest.Headers[RequestResponseHeaders.SourceAppIdHeader];
+
+                string myAppId = string.Empty;
+                bool foundMyAppId = false;
+                if (!string.IsNullOrEmpty(requestTelemetry.Context.InstrumentationKey))
+                {
+                    foundMyAppId = CorelationIdLookupHelper.TryGetAppId(requestTelemetry.Context.InstrumentationKey, out myAppId);
```
OsvaldoRosado: Out of curiosity. Why this syntax with an output parameter vs returning the id or null if not found?

Src/Common/CorelationIdLookupHelper.cs
nizarq: fixed

Src/Common/CorelationIdLookupHelper.cs
nizarq: fixed

Src/Common/CorelationIdLookupHelper.cs
nizarq: There are many ways an instrumentation key might be set. One of these ways is using a custom telemetry initializer. Hence the right way to ascertain instrumentation key is from the telemetry object before it is sent. It might be different from what's in the config. 

This cache saves the app ids against instrumentation keys (can be more than one) and makes sure we don't always call the breeze endpoint.

Src/Common/CorelationIdLookupHelper.cs
nizarq: fixed

Src/Common/RequestResponseHeaders.cs
nizarq: Don't think so. Even if we use the same header name to now hold app id where previously we were holding hashed instrumentation key, two application - one using beta and the other full sdk would have broken cross component correlation. Since one will be sending ikey hash in this header and the other will be expecting an app id. 

It is infact, better that we used different header names, so that this situation may never occur

Src/Web/Web.Shared.Net.Tests/OperationCorrelationTelemetryInitializerTests.cs
nizarq: Yes - tests need to be modified. I need to mock the service you've written to use in tests. They are intentionally left broken for now. I do not indent to merge this change without first putting in the change to use your service; at which point I'll update tests as well.

Src/Web/Web.Shared.Net/OperationCorrelationTelemetryInitializer.cs
nizarq: I like to be more explicit in my interfaces. Basically in principle I want to distinguish between failing to call the service vs. the service returning null / empty string as a value for app id.

Src/Common/CorelationIdLookupHelper.cs
```diff
@@ -0,0 +1,80 @@
...
+
+                string appId;
+                try
+                {
+                    // Ideally, we wouldn't do .Result and keep propogating the task up.
+                    // But in this case - it is useless, because we are eventually getting called by things like profiler callbacks which are not async;
+                    // If we don't block now, at some point up the call stack where we are hooking into an outgoing request / response, we will have to. Might as well do it here.
+                    appId = FetchAppIdFromService(instrumentationKey.ToLowerInvariant()).Result;
+                }
+                catch(AggregateException)
```
OsvaldoRosado: We'll probably need to handle more here. Transient network errors should return false and not store to cache, but breeze saying a profile doesn't exist (404) should probably get cached

Src/Common/CorelationIdLookupHelper.cs
```diff
@@ -0,0 +1,80 @@
...
+                    knownAppIds.Clear();
+                }
+
+                string appId;
+                try
+                {
+                    // Ideally, we wouldn't do .Result and keep propogating the task up.
+                    // But in this case - it is useless, because we are eventually getting called by things like profiler callbacks which are not async;
+                    // If we don't block now, at some point up the call stack where we are hooking into an outgoing request / response, we will have to. Might as well do it here.
+                    appId = FetchAppIdFromService(instrumentationKey.ToLowerInvariant()).Result;
```
Dmitry-Matveev: If this can be part of some synchronous path in customer code, e.g. the profiler callback that you mention, I would suggest to change a paradigm a little bit: Retrieve AppId asynchronously, and do not send correlation ID for the first couple times the dependency is called. This seems to be a better way off for enterprise level apps, rather than block their calls on http response from AI Endpoint. It looks like if we have a call to AI Endpoint on any customer code path then we are introducing a failure mode/perf degradation point.

Src/Common/CorelationIdLookupHelper.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// Computes the SHA256 hash for a given value and returns it in the form of a base64 encoded string.
+        /// </summary>
+        /// <param name="value">Value for which the hash is to be computed.</param>
+        /// <returns>Base64 encoded hash string.</returns>
+        private static async Task<string> FetchAppIdFromService(string value)
+        {
+            // Stub - needs to be replaced once the actual service is available.
+            TaskCompletionSource<string> mockAsyncTask = new TaskCompletionSource<string>();
+            mockAsyncTask.SetResult("mockAppId");
+
```
brahmnes: To be implemented?

nizarq: yes - this is to be implemented. Working on it right now. Was waiting for Osvaldo's when I first wrote it.

Src/Common/CorelationIdLookupHelper.cs
Dmitry-Matveev: Simply put, I'd prefer to lose couple associated items in favor of avoiding any synchronous HTTP calls within customer code execution path.

Src/Common/Common.projitems
```diff
@@ -9,7 +9,8 @@
     <Import_RootNamespace>Microsoft.ApplicationInsights.Common</Import_RootNamespace>
   </PropertyGroup>
   <ItemGroup>
+    <Compile Include="$(MSBuildThisFileDirectory)CommonEventSource.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)RequestResponseHeaders.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)InstrumentationKeyHashLookupHelper.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)CorelationIdLookupHelper.cs" />
```
OsvaldoRosado: Typo in name

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        /// <summary>
+        /// Retrieves the correlation id corresponding to a given instrumentation key.
+        /// </summary>
+        /// <param name="instrumentationKey">Instrumentation key string.</param>
+        /// <param name="breezeEndpointAddress">Breeze endpoint to query against.</param>
+        /// <param name="correlationId">AppId corresponding to the provided instrumentation key</param>
+        /// <returns>true if correlationId was successfully retrieved; false otherwise.</returns>
+        public static bool TryGetXComponentCorrelationId(string instrumentationKey, string breezeEndpointAddress, out string correlationId)
```
OsvaldoRosado: The argument name `breezeEndpointAddress` (and the other instances of Breeze in argument names in this class) seem like internal knowledge that doesn't belong in the SDK. In my opinion, just `endpointAddress` would be better

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -0,0 +1,158 @@
...
+            if (found)
+            {
+                return true;
+            }
+            else
+            {
+
+                if (breezeEndpointAddress == null)
+                {
+                    throw new ArgumentNullException("breezeEndpointAddress");
```
OsvaldoRosado: Is this caught somewhere? The SDK shouldn't throw into user code

Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
```diff
@@ -385,13 +391,17 @@ private void OnEnd(object exception, object thisObj, object returnValue)
...
-                                    telemetry.Target += " | " + targetIkeyHash;
+                                    // We only add the cross component correlation key if the key does not remain the current component.
+                                    if (!string.IsNullOrEmpty(targetAppId) && targetAppId != myAppId)
+                                    {
+                                        telemetry.Type = RemoteDependencyConstants.AI;
+                                        telemetry.Target += " | " + targetAppId;
+                                    }
                                 }
                             }
                         }
```
OsvaldoRosado: Should the try catch this is wrapped in handle more than ObjectDisposedException? There's an outer catch over everything but it's not clear to me that it is appropriate to jump out of this logic and skip setting resultcode & success as well as not run EndTracking on exceptions from correlation id.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -109,6 +123,28 @@ public void OnEndRequest(HttpContext context)
...
 
+            if (string.IsNullOrEmpty(requestTelemetry.Source) && context.Request.Headers != null)
+            {
+                string sourceAppId = context.Request.Headers[RequestResponseHeaders.SourceAppIdHeader];
+
+                string myAppId = string.Empty;
+                bool foundMyAppId = false;
+                if (!string.IsNullOrEmpty(requestTelemetry.Context.InstrumentationKey))
+                {
+                    foundMyAppId = CorrelationIdLookupHelper.TryGetXComponentCorrelationId(requestTelemetry.Context.InstrumentationKey, EffectiveProfileQueryEndpoint, out myAppId);
```
OsvaldoRosado: This can throw, right?

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -132,9 +168,14 @@ public void AddTargetHashForResponseHeader(HttpContext context)
             }
 
             if (!string.IsNullOrEmpty(requestTelemetry.Context.InstrumentationKey)
-                && context.Response.Headers[RequestResponseHeaders.TargetInstrumentationKeyHeader] == null)
+                && context.Response.Headers[RequestResponseHeaders.TargetAppIdHeader] == null)
             {
-                context.Response.Headers[RequestResponseHeaders.TargetInstrumentationKeyHeader] = InstrumentationKeyHashLookupHelper.GetInstrumentationKeyHash(requestTelemetry.Context.InstrumentationKey);
+                string appId;
+
+                if (CorrelationIdLookupHelper.TryGetXComponentCorrelationId(requestTelemetry.Context.InstrumentationKey, EffectiveProfileQueryEndpoint, out appId))
```
OsvaldoRosado: See above comment about this throwing

Src/Common/CommonEventSource.cs
```diff
@@ -0,0 +1,71 @@
...
+    using System.Globalization;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+#if NET40
+    using Microsoft.Diagnostics.Tracing;
+#endif
+
+    /// <summary>
+    /// ETW EventSource tracing class.
+    /// </summary>
+    [EventSource(Name = "Microsoft-ApplicationInsights-Extensibility-Common")]
```
SergeyKanzhelev: `Common` is a bad name for eventsource. Can you name is cross components correlation instead

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -0,0 +1,158 @@
...
+    using System.Collections.Concurrent;
+    using System.Globalization;
+    using System.IO;
+    using System.Net;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// A store for instrumentation App Ids. This makes sure we don't query the public endpoint to find an app Id for the same Ikey more than once.
+    /// </summary>
+    internal static class CorrelationIdLookupHelper
```
SergeyKanzhelev: why do you want this class to be static? Why wouldn't modules have an instace of it that they can pre-initialize?

Src/Common/CorrelationIdLookupHelper.cs
SergeyKanzhelev: Styling - use `nameof(breezeEndpointAddress)` instead of string =)

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -0,0 +1,158 @@
...
+        /// </summary>
+        private const int MAXSIZE = 100;
+
+        private const int GET_APP_ID_TIMEOUT = 2000; // milliseconds
+
+        private const string CORRELATION_ID_FORMAT = "aid-v1:{0}";
+
+        private const string APPID_QUERY_API_RELATIVE_URI_FORMAT = "api/profiles/{0}/appId";
+
+        private static ConcurrentDictionary<string, string> knownCorrelationIds = new ConcurrentDictionary<string, string>();
```
SergeyKanzhelev: provide a configuration experience to pre-populate this dictionary by setting property on corresponding telemetry module. If customer wants to run it once in startup task or during the build.

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -0,0 +1,158 @@
...
+                                CommonEventSource.Log.FetchAppIdFailed(ae.Flatten().InnerException.ToString());
+                            }
+                        });
+
+                        return false;
+                    }
+                }
+                catch(AggregateException ae)
+                {
+                    CommonEventSource.Log.FetchAppIdFailed(ae.Flatten().InnerException.ToString());
```
SergeyKanzhelev: use extension method that prints exception in invariant culture.

Src/Common/RequestResponseHeaders.cs
```diff
@@ -8,12 +8,12 @@ internal static class RequestResponseHeaders
         /// <summary>
         /// Source instrumentation header that is added by an application while making http requests and retrieved by the other application when processing incoming requests.
         /// </summary>
-        public const string SourceInstrumentationKeyHeader = "x-ms-request-source-ikey";
```
SergeyKanzhelev: can you pls talk to @lmolkova to use the header that will go to RFC?

Src/Common/CorrelationIdLookupHelper.cs
nizarq: Actually this shouldn't throw anyways because the method signature is Try... Updated.

Src/Common/CorrelationIdLookupHelper.cs
nizarq: fixed.

Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
nizarq: The try method shouldn't throw. I've updated the implementation. With that, we shouldn't need to change anything.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
nizarq: Try method cannot throw.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
nizarq: same here.

Src/Common/CommonEventSource.cs
nizarq: resolved.

Src/Common/CorrelationIdLookupHelper.cs
nizarq: throw is gone. Irrelevant.

Src/Common/CorrelationIdLookupHelper.cs
nizarq: advanced feature. Don't really need it

Src/Common/CorrelationIdLookupHelper.cs
nizarq: It's static because I wanted to share the dictionary (mapping between instrumentation key and app ID) across everything. 
Besides, the dictionary, there aren't any instance specific properties. 

One can argue whether or not we should make the endpoint an instance specific property. Perhaps - we can and avoid  some computation - but then again the current code is optimized such that it will only worry about the endpoint if the instrumentation key is not found in the current dictionary entries - so not too worried about that either.

SergeyKanzhelev: You do not share as you compile this code twice - once per module. Not worth it

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -0,0 +1,158 @@
...
+                    knownCorrelationIds.Clear();
+                }
+
+                try
+                {
+                    // Todo: When this fails, say in the vortex endpoint case, ProfileQueryEndpont is not provided, it may perpetually keep failing.
+                    // We can possibly make it more robust by having an exponential backoff on making a call to prod endpoint. Or store failure and never query again.
+                    // Is that worth the effort?
+                    
+                    // We wait for 2 seconds to retrieve the appId. If retrieved during that time, we return success setting the correlation id.
```
Dmitry-Matveev: Why do we still wait for even 2 seconds here?

Even though the current response is expected to be under 500 ms, it may still screw up the operation API statistics if it takes 100 ms without SDK code on this path.
Imagine 9 calls by 10 ms and 1 call with 500 ms leading to an average of 59 ms - 6 times slower than without AI SDK.
Looking at the statistics before/after I tried AI I would think of turning AI off due to 6x slow down on my fast dependency calls.

This comes in addition to the execution of the http call in the dependency collection context that (in case or insufficient protection on our side) can lead to another dependency collected (for our call) and unnecessary stacks (if we do not get it from the first time and our own call that's collected as dependency will initiate another call to AI Endpoint).

Also, the execution of the external communication in the user code context does not sound like a good idea in general. Can we make this code entirely asynchronous and then take the value if it was already populated?

Src/Common/RequestResponseHeaders.cs
nizarq: Per @lmolkova the only headers affected by her work are Parent and Root Id headers. The source and target headers have no story. She suggests no change is required here.

Src/Common/CorrelationIdLookupHelper.cs
nizarq: I've reduced the time out to 0 ms. At this point this should pretty much be entirely async.

I've not eliminated the timeout totally - instead have allowed the test hook to provide a non-zero timeout so that tests can work without having to make the calls twice or having to resort to pre-populating the dictionary - which may result in not hitting some correlation id formatting code.

However, the communication has to be in the user code context - because the instrumentation key required to get the app id is not known until the send telemetry call.

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -17,7 +17,10 @@ internal class CorrelationIdLookupHelper
         /// </summary>
         private const int MAXSIZE = 100;
 
-        private const int GET_APP_ID_TIMEOUT = 2000; // milliseconds
+        // For now we have decided to go with not waiting to retrieve the app Id, instead we just cache it on retrieval.
+        // This means the initial few attempts to get correlation id might fail and the inital telemetry sent might be missing such data.
```
pharring: nit: typo 'inital' -> 'initial'

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -0,0 +1,178 @@
...
+        private ConcurrentDictionary<string, string> knownCorrelationIds = new ConcurrentDictionary<string, string>();
+
+        private Func<string, Task<string>> provideAppId;
+
+        /// <summary>
+        /// This constructor is mostly used by the test classes to provide an override for fetching appId logic
+        /// </summary>
+        /// <param name="appIdProviderMethod">The delegate to be called to fetch the appId</param>
+        /// <param name="expectedResponseTime">Wait time for the provided method to return the app Id.</param>
+        public CorrelationIdLookupHelper(Func<string, Task<string>> appIdProviderMethod, int expectedResponseTime = 20)
```
Dmitry-Matveev: It looks like default wait time is still non-zero, 20 in this particular case.

Src/Common/CrossComponentCorrelationEventSource.cs
```diff
@@ -0,0 +1,71 @@
...
+    using System.Globalization;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+#if NET40
+    using Microsoft.Diagnostics.Tracing;
+#endif
+
+    /// <summary>
+    /// ETW EventSource tracing class.
+    /// </summary>
+    [EventSource(Name = "Microsoft-ApplicationInsights-Extensibility-Common")]
```
SergeyKanzhelev: please change the name of event source as well...

Src/Common/CorrelationIdLookupHelper.cs
nizarq: This constructor is preety much a test hook. Is used by test classes only. The product uses the other constructor. 

I can however, change this constructor as well to have a 0 default value and let the test classes explicitly pass in a 20 value

nizarq: In fact - it just so happens that my mock tasks are created - completed. So I don't even need a timeout. I'm getting rid of the timeout input parameter on the constructor.

Src/Common/RequestResponseHeaders.cs
SergeyKanzhelev: I'd then use more generic names at least. The idea is that target and source just grab the value of this header and pass it on... So when in fture it will be `appid-v2:` we will not need to change the logic of every SDK, we can gradually switch everything to v2

Also let's talk how we will include `roleInstance` in this header later. We shouldn't change a header name for it

nizarq: Makes sense. I guess we can go with **x-ms-request-source** and **x-ms-response-target**. That will make it generic. So the header content can change to include whatever it needs to include about the request source and dependency target in future.

Also I suggest instead of using `appid-v1:` as the versioning string, we use `corr_id-v1:` or simply `id-v1:` That also removes any indication that this is appId. 

So today we will have `corr_id-v1: {app Id here}`, tomorrow we may have `corr_id: { base64 encoded json containing appId and roleInstance}`

Let me know if you agree.

SergeyKanzhelev: I have a meeting with @lmolkova tomorrow around switching to `Request-ID` instead of `x-ms-request-id`. So maybe for consistency we will need to drop `x-ms-` here as well. Also I like `request-source`, `response-target` doesn't sound good. `request-target` makes more sense when pronouncing it in a context. Saying response target it's unclear whether it is where response originated from or to...

Alternatively we can do `request-context` and `response-context` headers similar to `correlation-context`. We can follow the format of `appId=XXX, roleInstance = YYY`.

When do we need to finalize on this schedule wise?

nizarq: Removing 'x-' because it's a deprecated practice sounds good. But Request-ID or request-target sound like very common terms. I'm afraid a lot of people might be using those terms as headers already for orthogonal scenarios. I still see value in having a namespace like prefix e.g., `ms-request-source` or more appropriately `ai-request-source` so that we make it less likely to unintentionally conflict with an existing user header.

Scheduling wise, this change is targeted for 2.3 sdk release. Having said that, if there are no other issues on this PR and I can get a sign off - I would suggest we check-in the rest of the change, leaving this as it is and we can modify the header names in a subsequent change even all the way up to the very last stretch of the release.

Src/Common/WebHeaderCollectionExtensions.cs
```diff
@@ -0,0 +1,104 @@
...
+    internal static class WebHeaderCollectionExtensions
+    {
+        /// <summary>
+        /// For the given header collection, for a given header of name-value type, find the value of a particular key.
+        /// </summary>
+        /// <param name="headers">Header collection.</param>
+        /// <param name="headerName">Name of the header in the collection.</param>
+        /// <param name="keyName">Desired key of the key-value list.</param>
+        /// <returns>Value against the given parameters.</returns>
+        public static string GetNameValueHeaderValue(this NameValueCollection headers, string headerName, string keyName)
```
nizarq: We could have used something like [NameValeHeaderValue](https://msdn.microsoft.com/en-us/library/system.net.http.headers.namevalueheadervalue(v=vs.110).aspx) to process multi value header but that would have meant adding an additional nuget package dependency. Since we are sensitive to adding new nuget package dependencies - I chose to write the functionality myself.

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -0,0 +1,176 @@
...
+                            catch (AggregateException ae)
+                            {
+                                CrossComponentCorrelationEventSource.Log.FetchAppIdFailed(ae.Flatten().InnerException.ToInvariantString());
+                            }
+                        });
+
+                        return false;
+                    }
+                }
+                catch (AggregateException ae)
```
Dmitry-Matveev: May it only throw AggregateException from the above code? May concurrent Dictionary operations throw some other type of the exceptions? We might want to catch all here.

Src/Common/WebHeaderCollectionExtensions.cs
Dmitry-Matveev: Are those methods safe and keep the header collection in the same state it was before? I remember issue years ago when one could've touched headers and some flag was set off, so the next attempt to access those headers (from user code) failed. New method of reading headers was introduced in .NET since then, not sure if it applies here, though..

SergeyKanzhelev:
1. Will you change names as we agreed?
2. You said you will implement some sort of throttling when endpoint is down or do not have information for the profile so we will not start an http call on every http dependency and request processing. I cannot find it.
3. New requirement: please mark your http request with `SdkOpernation.Enter` like [here](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/PerformanceCollector/Shared/PerformanceCollectorModule.cs#L235) so exceptions thrown by http client will be marked as internal for Exceptions Statistics feature

Src/Common/WebHeaderCollectionExtensions.cs
nizarq: Well I'm not using the NameValueHeaderValue methods - so can't comment on whether they are safe or not and lack the necessary historical context that you are mentioning.

nizarq:
@SergeyKanzhelev - what name changes are you talking about. I already changed header name to "Request-Context" with "appId" as the key. I can make the key name "correlationId" or "source" and "target" if that's preferable. But Can't tell from your comment what you are talking about. 

As for throttling, I was more concerned when we had a wait time on failure. We don't anymore - so it won't affect performance of the user calls. That makes it optional. I (or someone else) can do it as a feature enhancement in the future. I can create a separate issue for that. Don't think that's a blocker.

SergeyKanzhelev:
I'm sorry, I was looking at the wrong thing for the header. 

Please create an issue for throttling

nizarq:
Issue for throttling created: https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/352

PR #353, opened 2017-03-10
Marking get app id http request so that any exceptions thrown by http…

nizarq:
… client will be marked as internal for exception statistics feature.


nizarq:
@SergeyKanzhelev - I had to make this change in a separate pull request because it relied on other changes in develop that were not in my branch.

Issue #335, opened 2017-02-20
Dependency Collection improvement for clientside failures


cijothomas:
This is applicable to Profiler scenarios only:
For client side errors like DNS resolution failures, DependencyModule marks dependency as Failed, with an empty ResultCode. Since we have access to the Exception object, its definitely possible to identify a client side error like DNS Resolution failure, and populate ResultCode to indicate the same.

PR #362, opened 2017-03-14
support non-server result codes

SergeyKanzhelev:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/335

### DIFF
--- a/Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
+++ b/Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
@@ -390,6 +390,7 @@
                     if (responseObj != null)
                     {
                         int statusCode = -1;
+
                         try
                         {
                             statusCode = (int)responseObj.StatusCode;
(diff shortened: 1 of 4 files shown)


Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
```diff
@@ -387,25 +385,33 @@ private void OnEnd(object exception, object thisObj, object returnValue)
...
+                        {
+                            // ObjectDisposedException is expected here in the following sequence: httpWebRequest.GetResponse().Dispose() -> httpWebRequest.GetResponse()
+                            // on the second call to GetResponse() we cannot determine the statusCode.
                         }
-#endif
+                    }
+                    else if (exception != null)
+                    {
+                        var webException = exception as WebException;
+                        telemetry.ResultCode = webException.Status.ToString();
```
dnduffy: To be safe I think you need to check that webException is not null.
e.g. webException?.Status.ToString();

SergeyKanzhelev: thanks!

PR #363, opened 2017-03-14
Add DependencyCollector implementation for .NET Core

daschult:


Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -119,22 +122,36 @@ public bool TryGetXComponentCorrelationId(string instrumentationKey, out string
                             {
                                 this.GenerateCorrelationIdAndAddToDictionary(instrumentationKey, appId.Result);
                             }
+#if !NETCORE
                             catch (Exception ex)
                             {
                                 CrossComponentCorrelationEventSource.Log.FetchAppIdFailed(this.GetExceptionDetailString(ex));
                             }
+#else
+                            catch (Exception)
```
nizarq: How would the user go about debugging if say the endpoint to get app ID is down. Shouldn't we log this somewhere even for .net core?

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -119,22 +122,36 @@ public bool TryGetXComponentCorrelationId(string instrumentationKey, out string
...
                 {
                     CrossComponentCorrelationEventSource.Log.FetchAppIdFailed(this.GetExceptionDetailString(ex));
 
                     correlationId = string.Empty;
                     return false;
                 }
+#else
+                catch (Exception)
+                {
+                    correlationId = string.Empty;
```
nizarq: same here

Src/Common/CorrelationIdLookupHelper.cs
daschult: The CrossComponentCorrelationEventSource is not available for .NETStandard1.6. What other logging options are available?

Src/Common/CorrelationIdLookupHelper.cs
daschult: Same above.

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -152,24 +169,31 @@ private void GenerateCorrelationIdAndAddToDictionary(string ikey, string appId)
...
-                    using (StreamReader reader = new StreamReader(response.GetResponseStream()))
-                    {
-                        return await reader.ReadToEndAsync();
-                    }
+                    return await reader.ReadToEndAsync();
+                }
+#else
+                using (HttpClient client = new HttpClient())
+                {
+                    return await client.GetStringAsync(appIdEndpoint);
```
nizarq: .ConfigureAwait(false). Probably not very important with 0 wait time for first call (as things are today) - but if the timout in the caller was non zero - you would want to get to that timer which this async call was happening. Hence encouraging the framework to queue continuation on a separate thread is a good idea.

daschult: Good catch. I'll do that.

daschult: Done.

Src/Common/CorrelationIdLookupHelper.cs
daschult: (I agree completely with what you're saying. I just don't know what other options are available.)

nizarq: I haven't spent any amount of time in the .net core code base for AI. Perhaps @osvaldorosado can comment

dnduffy: Use an eventsource that is available or change the project target.

Src/DependencyCollector/Microsoft.AI.DependencyCollector.Tests/MockCorrelationIdLookupHelper.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace Microsoft.ApplicationInsights.DependencyCollector
+{
+    using Microsoft.ApplicationInsights.Common;
+    using System.Collections.Generic;
+
+    internal class MockCorrelationIdLookupHelper : ICorrelationIdLookupHelper
+    {
+        private readonly Dictionary<string, string> instrumentationKeyToCorrelationIdMap = new Dictionary<string, string>();
+
+        public void SetCorrelationId(string instrumentationKey, string correlationId)
```
nizarq: Niggle - you could also just have a constructor that takes the full content of the dictionary in one go and sets things up. Most likely when you are setting up a test - you already know all possible values. It is arguably cleaner to avoid creation and population being separate steps.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,190 @@
...
+                DependencyTelemetry telemetry = new DependencyTelemetry();
+                this.client.Initialize(telemetry);
+                telemetry.Start();
+                telemetry.Name = resourceName;
+                telemetry.Target = requestUri.Host;
+                telemetry.Type = "Http";
+                telemetry.Data = requestUri.OriginalString;
+                this.pendingTelemetry.TryAdd(loggingRequestId, telemetry);
+
+                if (!request.Headers.Contains(RequestResponseHeaders.RequestContextSourceKey))
```
nizarq: can request.Headers ever be null?
Also this check is totally wrong. Use request.Headers.GetNameValueHeaderValue(RequestResponseHeaders.RequestContextHeader, RequestResponseHeaders.RequestContextSourceKey) == null. 

See .net implemenation.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,190 @@
...
+                {
+                    string sourceApplicationId;
+                    if (this.correlationIdLookupHelper.TryGetXComponentCorrelationId(telemetry.Context.InstrumentationKey, out sourceApplicationId))
```
nizarq: what if no instrumentation key is set?

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,190 @@
...
+                    {
+                        request.Headers.Add(RequestResponseHeaders.RequestContextSourceKey, sourceApplicationId);
```
nizarq: use request.Headers.SetNameValueHeaderValue(RequestResponseHeaders.RequestContextHeader, RequestResponseHeaders.RequestContextSourceKey, appId);

See .net implementation for details.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,190 @@
...
+        /// </summary>
+        [DiagnosticName("System.Net.Http.Response")]
+        public void OnResponse(HttpResponseMessage response, Guid loggingRequestId)
+        {
+            if (response != null)
+            {
+                DependencyTelemetry telemetry;
+                if (this.pendingTelemetry.TryRemove(loggingRequestId, out telemetry))
+                {
+                    if (response.Headers.Contains(RequestResponseHeaders.RequestContextTargetKey))
```
nizarq: same as above.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,190 @@
...
+                    {
+                        string targetApplicationId = response.Headers.GetValues(RequestResponseHeaders.RequestContextTargetKey).SingleOrDefault();
+                        if (!string.IsNullOrEmpty(targetApplicationId))
+                        {
+                            // We only add the cross component correlation key if the key does not represent the current component.
+                            string sourceApplicationId;
+                            if (this.correlationIdLookupHelper.TryGetXComponentCorrelationId(telemetry.Context.InstrumentationKey, out sourceApplicationId) &&
```
nizarq: can instrumentation key be null or empty?

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,190 @@
...
+                                targetApplicationId != sourceApplicationId)
+                            {
+                                telemetry.Type = "Application Insights";
```
nizarq: "Application Insights" has changed to Http (tracked component) - see .net implemenation. I realize it's not part of your change but it's a simple string change in a single place - please go ahead and make the change.

Src/DependencyCollector/Microsoft.AI.DependencyCollector.Tests/MockCorrelationIdLookupHelper.cs
daschult: Done.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
daschult: ProfilerHttpProcessing.cs (the .NET implementation) you are using WebRequest and WebHeadersCollection. I'm using a completely different set of classes here (HttpRequestMessage and HttpRequestHeaders). That method isn't available to me, since the extension method is written for NameValueCollection, and HttpHeaders doesn't inherit from that class.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
daschult: I added a check to make sure that the instrumentation key isn't null or empty.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
daschult: I added a check to make sure that the instrumentation key isn't null or empty.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
daschult: Fixed (I changed this implementation to use the RemoteDependencyConstants values).

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
daschult: I wrote a function that checks if the provided header value exists using just the GetValues() function instead of Contains().

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
daschult: Same as above.

Src/Common/CorrelationIdLookupHelper.cs
dnduffy: P.S. I mentioned this in the last pull request, EventSource is supported in .NET Core.

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -152,24 +169,31 @@ private void GenerateCorrelationIdAndAddToDictionary(string ikey, string appId)
         {
             try
             {
+#if !NETCORE
                 SdkInternalOperationsMonitor.Enter();
-
+#endif
                 Uri appIdEndpoint = this.GetAppIdEndPointUri(instrumentationKey);
-
+#if !NETCORE
```
dnduffy: This was try for PCL, but is it still true for .NET Core?

daschult: In DependencyCollector.NetCore's project.json file I define a buildOption for NETCORE, so it allows me to use NETCORE as a preprocessor definition.

daschult:
I'm going to create a new pull request against the 2.4.0-beta1 branch, per Sergey Kanzhelev's request.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/project.json
```diff
@@ -0,0 +1,37 @@
+{
+  "version": "2.3.0-beta2",
```
dnduffy: Is there any way to pick this up from the targets or props file where it is defined now so we don't have to remember to maintain it in two places?
If not, please update to 2.3.0-beta3 since that just shipped and update the OneNote instructions for incrementing the version number for a release to include this.

dirs.proj
```diff
@@ -1,27 +1,52 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project=".\GlobalStaticVersion.props" />
```
dnduffy: Where does this come from now?

PR #366, opened 2017-03-15
Add .NETStandard DependencyCollector implementation

daschult:
Attempt number 3!


Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,235 @@
...
+
+                HttpRequestHeaders requestHeaders = request.Headers;
+                if (requestHeaders != null)
+                {
+                    if (!string.IsNullOrEmpty(telemetry.Context.InstrumentationKey) && !ContainsRequestContextKeyValue(requestHeaders, RequestResponseHeaders.RequestContextSourceKey))
+                    {
+                        string sourceApplicationId;
+                        if (this.correlationIdLookupHelper.TryGetXComponentCorrelationId(telemetry.Context.InstrumentationKey, out sourceApplicationId))
+                        {
+                            SetRequestContextKeyValue(requestHeaders, RequestResponseHeaders.RequestContextSourceKey, sourceApplicationId);
```
nizarq: what is the exception handling story here. Getting or setting request context key values or other calls here might throw. We shouldn't let the exception bubble up to the user. If we failed to fetch and add the correlation header, we should still let the user's calls carry on.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,235 @@
...
+        /// <summary>
+        /// Diagnostic event handler method for 'System.Net.Http.Response' event. Even in the case of an exception, this will still be called.
+        /// See https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/src/System/Net/Http/DiagnosticsHandler.cs#L71 for more information.
+        /// </summary>
+        [DiagnosticName("System.Net.Http.Response")]
+        public void OnResponse(HttpResponseMessage response, Guid loggingRequestId)
+        {
+            if (response != null)
+            {
+                DependencyTelemetry telemetry;
```
nizarq: Have some uber exception handling so that anything that goes wrong in this OnResponse callback doesn't fail user's application. The user should still be able to live without dependency data going to AI.

NuGet.config
```diff
@@ -16,7 +16,10 @@
         <add key="repositoryPath" value="..\packages" />
     </config>
   <packageSources>
+    <add key="LocallyBuiltApplicationInsightsDebugPackages" value="..\bin\Debug\NuGet" />
```
dnduffy: Unfortunately it is intentional, we hope to be able to remove it with the new project format soon.

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -152,24 +155,31 @@ private void GenerateCorrelationIdAndAddToDictionary(string ikey, string appId)
...
-
+#endif
                 Uri appIdEndpoint = this.GetAppIdEndPointUri(instrumentationKey);
-
+#if !NETCORE
                 WebRequest request = WebRequest.Create(appIdEndpoint);
                 request.Method = "GET";
 
                 using (HttpWebResponse response = (HttpWebResponse)await request.GetResponseAsync().ConfigureAwait(false))
+                using (StreamReader reader = new StreamReader(response.GetResponseStream()))
```
dnduffy: It is correct the way it is, successive usings can stack and don't indent again if they are sharing a single set of braces (which these are).

Src/Common/WebHeaderCollectionExtensions.cs
```diff
@@ -29,26 +27,9 @@ public static string GetNameValueHeaderValue(this NameValueCollection headers, s
...
-
-                    if (keyValue.Length == 2 && keyValue[0].Trim() == keyName)
-                    {
-                        return keyValue[1].Trim();
-                    }
-                }
-            }
-
-            return null;
+            return HeadersExtensions.GetHeaderKeyValue(headers[headerName]?.Split(','), keyName);
```
dnduffy: Only if you build with an older compiler as well, should be fine.

Src/Common/HeadersExtensions.cs
```diff
@@ -0,0 +1,107 @@
+﻿namespace Microsoft.ApplicationInsights.Common
+{
+    using System.Collections.Generic;
+    using System.Globalization;
+
+    /// <summary>
+    /// Generic functions that can be used to get and set Http headers.
+    /// </summary>
+    public static class HeadersExtensions
```
daschult: Done.

Src/Common/HeadersExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// provided key value. If the initial header value strings don't contain the key name,
+        /// then the key name/value pair should be added to the comma-separated list and returned.
+        /// </summary>
+        /// <param name="headerValues">The existing header values that the key/value pair should be added to.</param>
+        /// <param name="keyName">The name of the key to add.</param>
+        /// <param name="keyValue">The value of the key to add.</param>
+        /// <returns>The result of setting the provided key name/value pair into the provided headerValues.</returns>
+        public static string SetHeaderKeyValue(IEnumerable<string> headerValues, string keyName, string keyValue)
+        {
+            string result = string.Empty;
```
daschult: I optimized this function a bit, which includes using a StringBuilder.

Src/Common/HeadersExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+            return result;
+        }
+
+        /// <summary>
+        /// Append the provided toAppend string after the provided value string.
+        /// </summary>
+        /// <param name="value">The string to append toAppend to.</param>
+        /// <param name="toAppend">The string to append after value.</param>
+        /// <returns>The result of appending toAppend to value with a comma a space separating them.</returns>
+        private static string Append(string value, string toAppend)
```
daschult: I simplified this when I optimized this class.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,235 @@
...
+                string httpMethod = request.Method.Method;
+                Uri requestUri = request.RequestUri;
+                string resourceName = requestUri.AbsolutePath;
+                if (!string.IsNullOrEmpty(httpMethod))
+                {
+                    resourceName = httpMethod + " " + resourceName;
+                }
+
+                DependencyTelemetry telemetry = new DependencyTelemetry();
+                this.client.Initialize(telemetry);
```
daschult: This call is really about setting the instrumentation key in the DependencyTelemetry object. This is the same thing that we do in the other DependencyCollector implementations.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,235 @@
...
+        {
+            IEnumerable<string> result;
+            if (headers == null || !headers.TryGetValues(headerName, out result))
+            {
+                result = Enumerable.Empty<string>();
+            }
+            return result;
+        }
+
+        internal static string GetHeaderKeyValue(HttpHeaders headers, string headerName, string keyName)
```
daschult: These are all internal so that I can run them in my tests.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
daschult: I've added a try-catch block to both OnRequest() and OnResponse() that catches all exceptions and then logs them to the CrossComponentCorrelationEventSource.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
daschult: I've added a try-catch block to both OnRequest() and OnResponse() that catches all exceptions and then logs them to the CrossComponentCorrelationEventSource.

Src/DependencyCollector/Microsoft.AI.DependencyCollector.Tests/MockCorrelationIdLookupHelper.cs
```diff
@@ -0,0 +1,22 @@
...
+        public MockCorrelationIdLookupHelper(Dictionary<string, string> instrumentationKeyToCorrelationIdMap)
+        {
+            this.instrumentationKeyToCorrelationIdMap = instrumentationKeyToCorrelationIdMap;
+        }
+
+        public bool TryGetXComponentCorrelationId(string instrumentationKey, out string correlationId)
+        {
+            bool result = this.instrumentationKeyToCorrelationIdMap.ContainsKey(instrumentationKey);
+            correlationId = result ? this.instrumentationKeyToCorrelationIdMap[instrumentationKey] : string.Empty;
+            return result;
```
daschult: Done.

Src/Common/CrossComponentCorrelationEventSource.cs
```diff
@@ -55,13 +57,27 @@ public void GetHeaderFailed(string exception, string appDomainName = "Incorrect"
...
+        }
+
         [NonEvent]
         private string GetApplicationName()
         {
             string name;
             try
             {
+#if NETCORE
+                name = typeof(CrossComponentCorrelationEventSource).GetTypeInfo().Assembly.GetName().FullName;
```
dnduffy: Is this supposed to be our DLL name or the end customer's application name?

Src/Common/HeadersUtilities.cs
```diff
@@ -0,0 +1,139 @@
...
+                            last = lastSearchIndex;
+                            break;
+                        }
+                    }
+                    break;
+                }
+            }
+
+            return firstNonWhitespaceIndex == -1 ? null : value.Substring(firstNonWhitespaceIndex, last - firstNonWhitespaceIndex + 1);
+        }
```
dnduffy: Why not use the built in .NET string functions IndexOf and LastIndexOf?

daschult: Because those functions only take a character or a string as the search parameter. I'm using a function here.

Src/Common/CrossComponentCorrelationEventSource.cs
daschult: It's supposed to be the equivalent of AppDomain.CurrentDomain.FriendlyName. .NETStandard doesn't support AppDomain, so this seemed to be the closest I could find, but it sounds like it's still not right. Do you have any suggestions?

Src/DependencyCollector/Microsoft.AI.DependencyCollector/project.json
```diff
@@ -0,0 +1,40 @@
+{
+  "version": "2.3.0-beta2",
```
dnduffy: Is there any way to pick this up from the targets or props file where it is defined now so we don't have to remember to maintain it in two places?
If not, please update to 2.3.0-beta3 since that just shipped and update the OneNote instructions for incrementing the version number for a release to include this.

dirs.proj
```diff
@@ -1,27 +1,52 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project=".\GlobalStaticVersion.props" />
```
dnduffy: Where does this come from now?

Src/Common/CrossComponentCorrelationEventSource.cs
dnduffy: We could try GetEntryAssembly which is supposed to exist in .NET Standard 1.5.

Src/Common/HeadersUtilities.cs
dnduffy: No worries, I realized later that the checks would get more complicated to do what I was thinking and not simpler as intended.

dirs.proj
daschult: This comes from Global.props. One of the warnings during build was that GlobalStaticVersion.props was being imported multiple times, and removing this reference solved it.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/project.json
daschult: Once these projects get migrated over to VS 2017, then we'll be able to reference a properties file for this reference. Until then, it needs to be hardcoded here.
I'll update it to 2.3.0-beta2.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,252 @@
...
+                    if (requestHeaders != null)
+                    {
+                        if (!string.IsNullOrEmpty(telemetry.Context.InstrumentationKey) && !ContainsRequestContextKeyValue(requestHeaders, RequestResponseHeaders.RequestContextSourceKey))
+                        {
+                            string sourceApplicationId;
+                            if (this.correlationIdLookupHelper.TryGetXComponentCorrelationId(telemetry.Context.InstrumentationKey, out sourceApplicationId))
+                            {
+                                SetRequestContextKeyValue(requestHeaders, RequestResponseHeaders.RequestContextSourceKey, sourceApplicationId);
+                            }
+                        }
```
nizarq: What if the exception is thrown somewhere while trying to add cross component correlation. Why would that not allow root id / partent id headers to be set. There is room for improving some localized exception handling as well.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,252 @@
...
+                        if (!string.IsNullOrEmpty(targetApplicationId) && !string.IsNullOrEmpty(telemetry.Context.InstrumentationKey))
+                        {
+                            // We only add the cross component correlation key if the key does not represent the current component.
+                            string sourceApplicationId;
+                            if (this.correlationIdLookupHelper.TryGetXComponentCorrelationId(telemetry.Context.InstrumentationKey, out sourceApplicationId) &&
+                                targetApplicationId != sourceApplicationId)
+                            {
+                                telemetry.Type = RemoteDependencyConstants.AI;
+                                telemetry.Target += " | " + targetApplicationId;
+                            }
```
nizarq: Same here - if an exception happens trying to set cross component correlation, we can pretend as if there is no cross component correlation but still try and send the telemetry.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
nizarq: Why can't they be in a utility class then? Don't you already have a HeadersUtilities class?

daschult: Since they're only used in this class, I don't see why it would help to move them to a different class. Why not leave them where they're used?

daschult: And these header functions are specific to HttpHeaders, which are for the .NET Core impelementation.

Src/Common/CrossComponentCorrelationEventSource.cs
daschult: Changed to Assembly.GetEntryAssembly().FullName.

Src/Common/HeadersUtilities.cs
```diff
@@ -0,0 +1,139 @@
...
+{
+    using System.Collections.Generic;
+    using System.Globalization;
+    using System.Linq;
+    using System.Text;
+
+    /// <summary>
+    /// Generic functions that can be used to get and set Http headers.
+    /// </summary>
+    public static class HeadersUtilities
```
daschult: Done.

Src/Common/HeadersUtilities.cs
```diff
@@ -0,0 +1,139 @@
...
+    {
+        /// <summary>
+        /// Get the key value from the provided HttpHeader value that is set up as a comma-separated list of key value pairs. Each key value pair is formatted like (key)=(value).
+        /// </summary>
+        /// <param name="headerValue">The header values that may contain key name/value pairs.</param>
```
daschult: Done.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
nizarq: My concern is DependencyCollectorDiagnosticsListener.GetHeaderKeyValue doesn't make any sense. It is a good thing to strive to test everything but that shouldn't result in classes that are harder to make sense of. So for example, if they are only used by this class one way to test this functionality would be to test these through the classes internal / public methods. If you tests are telling you to make something internal when they aren't truly internal - that's a good indication that there is less cohesion in the class - that's where my suggestion about fixing it by moving it to a utility class comes.

Src/Common/HeadersUtilities.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// name/value pairs with the provided keyName and keyValue. If the initial header value
+        /// strings contains the key name, then the original key value should be replaced with the
+        /// provided key value. If the initial header value strings don't contain the key name,
+        /// then the key name/value pair should be added to the comma-separated list and returned.
+        /// </summary>
+        /// <param name="headerValues">The existing header values that the key/value pair should be added to.</param>
+        /// <param name="keyName">The name of the key to add.</param>
+        /// <param name="keyValue">The value of the key to add.</param>
+        /// <returns>The result of setting the provided key name/value pair into the provided headerValues.</returns>
+        public static string SetHeaderKeyValue(IEnumerable<string> headerValues, string keyName, string keyValue)
```
daschult: HttpHeaders returns header values as an IEnumerable<string>. If this function took a comma-separated string of values, then we'd have to join the header values just to split them apart again. Instead, I've changed this function to take an IEnumerable<string> and to return an IEnumerable<string>.

Src/Common/HeadersUtilities.cs
```diff
@@ -0,0 +1,139 @@
...
+                        else if (name != keyName)
+                        {
+                            string value = TrimSubstring(keyValuePair, equalsSignIndex + 1, keyValuePair.Length);
+                            if (value == null)
+                            {
+                                value = string.Empty;
+                            }
+                            Append(result, CreateKeyValuePair(name, value));
+                        }
+                        else if (!found)
```
daschult: That changes the order, and I was trying to not modify your tests. If the order really doesn't matter, then yes, I can do that.

daschult:
@dnduffy Can someone on your end please make your build machines build this pull request? I'd like to know sooner rather than later if I have any build machine-specific problems.

Src/Common/HeadersUtilities.cs
```diff
@@ -44,65 +44,18 @@ public static string GetHeaderKeyValue(IEnumerable<string> headerValue, string k
         /// <param name="keyName">The name of the key to add.</param>
         /// <param name="keyValue">The value of the key to add.</param>
         /// <returns>The result of setting the provided key name/value pair into the provided headerValues.</returns>
-        public static string SetHeaderKeyValue(IEnumerable<string> headerValues, string keyName, string keyValue)
+        public static IEnumerable<string> SetHeaderKeyValue(IEnumerable<string> headerValues, string keyName, string keyValue)
```
nizarq: Update the header description

Src/Common/WebHeaderCollectionExtensions.cs
```diff
@@ -49,7 +49,7 @@ public static void SetNameValueHeaderValue(this NameValueCollection headers, str
                 throw new ArgumentNullException(nameof(keyName));
             }
 
-            headers[headerName] = HeadersUtilities.SetHeaderKeyValue(headers[headerName]?.Split(','), keyName, value);
+            headers[headerName] = string.Join(", ", HeadersUtilities.SetHeaderKeyValue(headers[headerName]?.Split(','), keyName, value));
```
nizarq: Consider revising the name SetHeaderKeyValue. This makes for a weird read where you set and join and set again?
Don't have a good suggestion but may be UpdateHeaderWithKeyValue.

Src/Common/CrossComponentCorrelationEventSource.cs
```diff
@@ -27,6 +27,17 @@ private CrossComponentCorrelationEventSource()
 
         public string ApplicationName { [NonEvent]get; [NonEvent]private set; }
 
+        public static string GetExceptionDetailString(Exception ex)
```
daschult: Done.

Src/Common/HeadersUtilities.cs
daschult: Done.

Src/Common/WebHeaderCollectionExtensions.cs
daschult: Done.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,223 @@
...
+
+        private readonly ApplicationInsightsUrlFilter applicationInsightsUrlFilter;
+        private readonly TelemetryClient client;
+        private readonly ICorrelationIdLookupHelper correlationIdLookupHelper;
+        private readonly ConcurrentDictionary<Guid, DependencyTelemetry> pendingTelemetry = new ConcurrentDictionary<Guid, DependencyTelemetry>();
+
+        internal DependencyCollectorDiagnosticListener(TelemetryConfiguration configuration, ICorrelationIdLookupHelper correlationIdLookupHelper = null)
+        {
+            this.client = new TelemetryClient(configuration);
+            this.client.Context.GetInternalContext().SdkVersion = SdkVersionUtils.GetSdkVersion("rddf");
```
cijothomas: we need to use something other than 'rddf' to distinguish between dependencies collected via EventSource instrumentation.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,223 @@
...
+                        }
+                        catch (Exception e)
+                        {
+                            CrossComponentCorrelationEventSource.Log.UnknownError(ExceptionUtilities.GetExceptionDetailString(e));
+                        }
+
+                        // Add the root ID
+                        string rootId = telemetry.Context.Operation.Id;
+                        if (!string.IsNullOrEmpty(rootId) && !requestHeaders.Contains(RequestResponseHeaders.StandardRootIdHeader))
+                        {
```
cijothomas: why not use HttpHeadersUtilities method as before(line 138) for consistency?

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,223 @@
...
+                            CrossComponentCorrelationEventSource.Log.UnknownError(ExceptionUtilities.GetExceptionDetailString(e));
+                        }
+
+                        int statusCode = (int)response.StatusCode;
+                        telemetry.ResultCode = (0 < statusCode) ? statusCode.ToString(CultureInfo.InvariantCulture) : string.Empty;
+                        telemetry.Success = (0 < statusCode) && (statusCode < 400);
+
+                        telemetry.Stop();
+                        this.client.Track(telemetry);
+                    }
```
cijothomas: please add Verbose level logging for all 'else' cases to help debugging.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/project.json
```diff
@@ -0,0 +1,41 @@
...
+        "../../Common/ExceptionUtilities.cs",
+        "../../Common/ICorrelationIdLookupHelper.cs",
+        "../../Common/RequestResponseHeaders.cs",
+        "../../Common/HeadersUtilities.cs",
+        "../../Web/Web.Net45/Implementation/SdkVersionUtils.cs"
+      ]
+    }
+  },
+  "dependencies": {
+    "Microsoft.ApplicationInsights": "2.3.0-beta3",
```
cijothomas: this should be 2.4.0-beta1 right?

Test/DependencyCollector/FunctionalTests/FuncTest/HttpTests.cs
```diff
@@ -1,16 +1,18 @@
...
     using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using System.Diagnostics;
 
     /// <summary>
-    /// Tests RDD Functionality for a ASP.NET WebApplication in DOTNET 4.5.1 and DOTNET 4.6
+    /// Tests RDD Functionality for a ASP.NET WebApplication in DOTNET 4.5.1, DOTNET 4.6,
+    /// and DOTNET Core.
     /// ASPX451 refers to the test application throughout the functional test context.
-    /// The same app is used for testsing 4.5.1 and 4.6 scenarios.
+    /// The same app is used for testing 4.5.1 4.6, and Core scenarios.
```
cijothomas: For core, its not the same app. Update comment to reflect it.

Test/DependencyCollector/FunctionalTests/FuncTest/Helpers/DotNetCoreTestWebApplication.cs
```diff
@@ -0,0 +1,83 @@
...
+                    }
+                }
+            }
+        }
+
+        internal override void Remove()
+        {
+            if (process != null)
+            {
+                process.Kill();
```
cijothomas: We need to wait for exit after issuing kill right?

Test/DependencyCollector/FunctionalTests/FuncTest/Helpers/DotNetCoreProcess.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            return this;
+        }
+
+        /// <summary>
+        /// Block until this started process has exited.
+        /// </summary>
+        public void WaitForExit()
+        {
+            process.WaitForExit();
```
cijothomas: This can potentially be infinite wait.

Src/DependencyCollector/Microsoft.AI.DependencyCollector/DependencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,223 @@
...
+                    if (requestHeaders != null)
+                    {
+                        try
+                        {
+                            if (!string.IsNullOrEmpty(telemetry.Context.InstrumentationKey) && !HttpHeadersUtilities.ContainsRequestContextKeyValue(requestHeaders, RequestResponseHeaders.RequestContextSourceKey))
+                            {
+                                string sourceApplicationId;
+                                if (this.correlationIdLookupHelper.TryGetXComponentCorrelationId(telemetry.Context.InstrumentationKey, out sourceApplicationId))
+                                {
+                                    HttpHeadersUtilities.SetRequestContextKeyValue(requestHeaders, RequestResponseHeaders.RequestContextSourceKey, sourceApplicationId);
```
cijothomas: where is the logic to exclude adding headers for user configured domains?

Src/DependencyCollector/NetCore.Tests/DependencyCollector.NetCore.Tests.csproj
```diff
@@ -0,0 +1,27 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp1.0</TargetFramework>
```
lmolkova: why not netstandard 1.3?

Issue #257, opened 2016-12-08
Allow to override names of the correlation headers in outgoing http calls


SergeyKanzhelev:
Need an agreement on naming for the fields SourceInstrumentationKeyHeader and StandardRootIdHeader or allow to override those

SergeyKanzhelev:
/cc: @lmolkova to address with the correlation improvements

PR #368, opened 2017-03-16
[WIP] Dependency correlation part1: Mimic correlation protocol

lmolkova:
This change improves dependency correlation in AppInsights.Web
1. it aligns AppInsights.Web with [HTTP protocol](https://github.com/lmolkova/correlation/blob/master/http_protocol_proposal_v1.md) that is adopted by System.Net.Http (.NET Core) and currently being implemented in AspNetCore.
2. It uses execution context stored for the operation in Base SDK

This is a **work in progress** and besides feedback and comments you may have for the implementation, we need to have an agreement on following:
1. **Id generation**: it is a part of big feature implemented in .NET: [Activity API](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/ActivityUserGuide.md). I suggest to change AI base SDK to use Activity where possible (it's available starting from .NET 4.6) and mimic Id generation in older versions. Id becomes hierarchical with guaranteed randomness. Root behavior does not change.
It requires to bring System.Diagnostic.DiagnosticSource to base SDK that may be a problem for users who don't need correlation and use base SDK with powershell. I'll try to address this concern 
2. **Execution context** could be lost between HttpModule.BeginRequest and application code because of ASP.NET native/managed code switches. The solution from ASP.NET team is to use `OnPreRequestHandlerExecute` event to restore from HttpContext.Current any context if it was lost.
However there is no (and should not be) public API to restore CallContext/AsyncLocal set in `StartOperation`. **I suggest add `[InternalsVisisbleTo]` for AppInsgihts.Web to the base SDK lib**. Let me know if you have better suggestions.

Things that are not done yet and will be done
1. Unit/Functional tests for new headers
2. Correlation-Context implementation
3. **AppInsights.Web target and implementation for .NET4.6** ASP.NET classic team will provide their own HttpModule with DiagnosticSource/Activity support and will take care about AsyncLocal flow. AppInsights.Web should use it and do request tracking. The module is being implemented now.

### DIFF
--- a/Src/Common/AppInsightsActivity.cs
+++ b/Src/Common/AppInsightsActivity.cs
@@ -1,0 +1,64 @@
+﻿namespace Microsoft.ApplicationInsights.Common
+{
+    using System;
+    using System.Diagnostics;
+    using System.Threading;
+
+    //this is a temporary solution that mimics System.DiagnosticSource,Activity and Correlation HTTP protocol:
+    //https://github.com/lmolkova/correlation/blob/master/http_protocol_proposal_v1.md
+    //It does not implement 
+    // - Request-Id length limitation and overflow
+    internal class AppInsightsActivity
+    {
+        internal static string GetRootId(string requestId)
+        {
+            Debug.Assert(!string.IsNullOrEmpty(requestId));
+            if (requestId[0] == '|')
+            {
+                var rootEnd = requestId.IndexOf('.');
+                var rootId = requestId.Substring(1, rootEnd - 1);
+                return rootId;
+            }
+            return requestId;
+        }
+
+
+        internal static string GenerateNewId()
+        {
+            if (_machinePrefix == null)
+                Interlocked.CompareExchange(ref _machinePrefix, Environment.MachineName + "-" + ((int)Stopwatch.GetTimestamp()).ToString("x"), null);
+            return '|' + _machinePrefix + '-' + Interlocked.Increment(ref _currentOperationNum).ToString("x") + '.';
+        }
+
+        internal static string GenerateRequestId(string parentRequestId, string telemetryId)
+        {
+            if (parentRequestId != null)
+            {
+                var childRequestId = parentRequestId[0] != '|' ? '|' + parentRequestId : parentRequestId;
+                if (childRequestId[childRequestId.Length - 1] != '.')
+                    childRequestId += '.';
+
+                return GenerateChildTelemetryId(childRequestId, telemetryId, '_');
+            }
+            return GenerateNewId();
+        }
+
+        internal static string GenerateDependencyId(string parentRequestId, string telemetryId)
+        {
+            if (parentRequestId != null)
+                return GenerateChildTelemetryId(parentRequestId, telemetryId, '.');
+
+            return GenerateNewId();
+        }
+
+        private static string _machinePrefix;
+        private static long _currentOperationNum = BitConverter.ToUInt32(Guid.NewGuid().ToByteArray(), 12);
+
+        private static string GenerateChildTelemetryId(string parentId, string telemetryId, char delimiter)
+        {
+            Debug.Assert(!string.IsNullOrEmpty(parentId));
+            Debug.Assert(!string.IsNullOrEmpty(telemetryId));
+            return parentId + telemetryId + delimiter;
+        }
+    }
+}
--- a/Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
+++ b/Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
@@ -278,7 +278,6 @@
                 bool isCustomCreated = false;
 
                 var telemetry = ClientServerDependencyTracker.BeginTracking(this.telemetryClient);
-
                 telemetry.Name = resourceName;
                 telemetry.Target = DependencyTargetNameHelper.GetDependencyTargetName(url);
                 telemetry.Type = RemoteDependencyConstants.HTTP;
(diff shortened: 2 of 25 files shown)


SergeyKanzhelev:
Unrelated - do you think we can move https://github.com/lmolkova/correlation/blob/master/http_protocol_proposal_v1.md to microsoft repo so it looks more official? I can grant permissions to http://github.com/Microsoft/ApplicationInsights-Home to you

SergeyKanzhelev:
Please rebase to 2.4.0-beta1 as it will not go to 2.3 SDK

Src/Common/AppInsightsActivity.cs
```diff
@@ -0,0 +1,64 @@
...
+                return rootId;
+            }
+            return requestId;
+        }
+
+
+        internal static string GenerateNewId()
+        {
+            if (_machinePrefix == null)
+                Interlocked.CompareExchange(ref _machinePrefix, Environment.MachineName + "-" + ((int)Stopwatch.GetTimestamp()).ToString("x"), null);
```
SergeyKanzhelev: machine name can be long. Maybe take a hash of it?

lmolkova:
> @lmolkova we should preserve the scenario when a single dll is sufficient to report telemetry. Having said that I realized that we can take a reference anyway. We just need to ensure that code will function (gratefully degrade) when diagnostics source dll cannot be found.
>
> Do you think it is doable?
I do not think it's doable.
If the DiagnosticSource assembly is referenced by AppInsgiths, it's not possible to use AppInsights assembly without it.

If AppInsights don't reference DiagnosticSource assembly, there is probably no good way to require AppInsgihts and it's users to use it.
And if we dynamically load DiagnosticSource assembly, we would have to use reflection to work with Activity which is not an option.

How bad it would be to require users that use Microsoft.AppInsights.dll to copy DiagnosticSource dll to the folder where they run it? How many customers are affected by it?
I believe there is no assumption that an assembly could be used without it's dependencies. And requirement to keep AppInsgihts base SDK to depend on framework assemblies only is quite restrictive.

------------------
However we may consider following plan:

* AppInsights Base SDK does not do ID generation and is not aware about Activity.
* AppInsights.Web 4.0/4.5 generates Id on it's own and leverages CallContext in base SDK
* AppInsights.Web 4.6 works with Activity and don't need AppInsights Base to have it's own AsyncLocal
* AppInsights.Core works with Activity and don't need AppInsights Base to have it's own AsyncLocal.
* The assumption is that all other users of base SDK gradually move to Activity

This way AppInsights Base SDK may eventually drop `OperationContextForAsyncLocal` .
The problem that we never know if we can safely drop `OperationContextForAsyncLocal`  and in the mean time (that is probably forever) we will have duplicated implementation done in different parts of the AppInsights that is harder to maintain and bad in terms of performance.

Src/Common/AppInsightsActivity.cs
```diff
@@ -0,0 +1,64 @@
...
+        internal static string GenerateDependencyId(string parentRequestId, string telemetryId)
+        {
+            if (parentRequestId != null)
+                return GenerateChildTelemetryId(parentRequestId, telemetryId, '.');
+
+            return GenerateNewId();
+        }
+
+        private static string _machinePrefix;
+        private static long _currentOperationNum = BitConverter.ToUInt32(Guid.NewGuid().ToByteArray(), 12);
```
SergeyKanzhelev: in base SDK we are using custom ID generation that presumably faster than `guid.NewGuid`, We can take it or keep using guid - I'm fine either way

Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
```diff
@@ -323,9 +322,12 @@ private object OnBegin(object thisObj, bool isAsyncCall)
 
                     // Add the parent ID
                     var parentId = telemetry.Id;
-                    if (!string.IsNullOrEmpty(parentId) && webRequest.Headers[RequestResponseHeaders.StandardParentIdHeader] == null)
+                    if (!string.IsNullOrEmpty(parentId))
                     {
-                        webRequest.Headers.Add(RequestResponseHeaders.StandardParentIdHeader, parentId);
+                        if (webRequest.Headers[RequestResponseHeaders.StandardParentIdHeader] == null)
```
SergeyKanzhelev: oh, you adding both? We at least need a setting to only use new settings. Or better - allow to override the list of headers as suggested here https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/257

Src/Common/AppInsightsActivity.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    using System;
+    using System.Diagnostics;
+    using System.Threading;
+
+    //this is a temporary solution that mimics System.DiagnosticSource,Activity and Correlation HTTP protocol:
+    //https://github.com/lmolkova/correlation/blob/master/http_protocol_proposal_v1.md
+    //It does not implement 
+    // - Request-Id length limitation and overflow
+    internal class AppInsightsActivity
```
SergeyKanzhelev: If you need this class in Base SDK - send PR against 2.4.0-beta1 base and I can create a signed build for you once that PR is approved. Or you can keep this code here for now...

Need to think of a public surface of this class though.

lmolkova:
> Unrelated - do you think we can move https://github.com/lmolkova/correlation/blob/master/http_protocol_proposal_v1.md to microsoft repo so it looks more official? I can grant permissions to http://github.com/Microsoft/ApplicationInsights-Home to you

We definitely need to move it somewhere. We considered to store it in [corefx](https://github.com/dotnet/corefx) repo. Since it's implemented in .NET, it's ok to store it there. 
I'll do that in next couple of weeks (I'd like to do some minor changes first) and send PR to update http://github.com/Microsoft/ApplicationInsights-Home README (or whatever the good place is) with the link

Src/Common/AppInsightsActivity.cs
lmolkova: thanks, I'll use your weak random

Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
lmolkova: what if downstream service has lower version of AppInsights? If we care about backward compatibility, we should have both

Src/Common/AppInsightsActivity.cs
lmolkova: well the point of machine name is to give a bit more info to human being that is looking into telemetry Id. 
In Azure, machine name looks like `RD0003FF21406D` that is not long.

However I agree that machine name in a huge cluster may not be that useful.

Src/Common/AppInsightsActivity.cs
lmolkova: It makes sense to have it in Base SDK for 4.0/4.5 if we also have Activity for 4.6 there. Let's decide on it first.

lmolkova:
I've created a new PC for it #374, closing this one

Issue #359, opened 2017-03-14
[Exception Statistics] Threshold on # of exceptions/sec


SergeyKanzhelev:
Performance measurements shows significant overhead of exceptions statistics feature in case of "exceptions storm". 

We will improve performance of the feature, but for the stable 2.3 we need to implement a stopgap solution to degrade amount of telemetry collected when we see too many exceptions being tracked per second.

PR #370, opened 2017-03-17
Auto throttle for exception processing

kstevenham:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/359

### DIFF
--- a/Src/WindowsServer/WindowsServer.Shared.Tests/FirstChanceExceptionStatisticsTelemetryModuleTest.cs
+++ b/Src/WindowsServer/WindowsServer.Shared.Tests/FirstChanceExceptionStatisticsTelemetryModuleTest.cs
@@ -352,6 +352,66 @@
         }
 
         [TestMethod]
+        public void FirstChanceExceptionStatisticsTelemetryExceptionsAreThrottled()
+        {
+            var metrics = new List<KeyValuePair<Metric, double>>();
+            this.configuration.MetricProcessors.Add(new StubMetricProcessor()
+            {
+                OnTrack = (m, v) =>
+                {
+                    metrics.Add(new KeyValuePair<Metric, double>(m, v));
+                }
+            });
+
+            this.configuration.TelemetryInitializers.Add(new StubTelemetryInitializer()
+            {
+                OnInitialize = (item) =>
+                {
+                    item.Context.Operation.Name = "operationName";
+                }
+            });
+
+            using (var module = new FirstChanceExceptionStatisticsTelemetryModule())
+            {
+                module.Initialize(this.configuration);
+
+                module.TargetMovingAverage = 50;
+
+                for (int i = 0; i < 200; i++)
+                {
+                    try
+                    {
+                        // FirstChanceExceptionStatisticsTelemetryModule will process this exception
+                        throw new Exception("test");
+                    }
+                    catch (Exception exc)
+                    {
+                        // code to prevent profiler optimizations
+                        Assert.Equal("test", exc.Message);
+                    }
+                }
+            }
+
+            int countProcessed = 0;
+            int countThrottled = 0;
+
+            foreach (KeyValuePair<Metric, double> items in metrics)
+            {
+                if (items.Key.Dimensions.Count == 1)
+                {
+                    countThrottled++;
+                }
+                else
+                {
+                    countProcessed++;
+                }
+            }
+
+            Assert.Equal(166, countProcessed);
+            Assert.Equal(34, countThrottled);
+        }
+
+        [TestMethod]
         public void FirstChanceExceptionStatisticsTelemetryModuleDoNotIncrementOnRethrow()
         {
             var metrics = new List<KeyValuePair<Metric, double>>();
@@ -378,7 +438,7 @@
                     {
                         // this assert is neede to avoid code optimization
                         Assert.Equal("test", ex.Message);
-                        throw new Exception("new exception");
+                        throw;
                     }
                 }
                 catch (Exception exc)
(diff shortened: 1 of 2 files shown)


Src/WindowsServer/WindowsServer.Shared.Tests/FirstChanceExceptionStatisticsTelemetryModuleTest.cs
```diff
@@ -341,14 +453,9 @@ public void FirstChanceExceptionStatisticsTelemetryModuleDoNotIncrementOnRethrow
...
 
-            Assert.Equal(2, this.items.Count);
-
-            Assert.Equal(1, ((MetricTelemetry)this.items[0]).Count);
-            Assert.Equal(1, ((MetricTelemetry)this.items[1]).Count);
-
-            // One of them should be 0 as re-thorwn, another - one
-            Assert.Equal(0, Math.Min(((MetricTelemetry)this.items[0]).Sum, ((MetricTelemetry)this.items[1]).Sum), 15);
-            Assert.Equal(1, Math.Max(((MetricTelemetry)this.items[0]).Sum, ((MetricTelemetry)this.items[1]).Sum), 15);
+            Assert.Equal(1, this.items.Count);
```
Dmitry-Matveev: It looks like change in the dim capping also affected Rethrow counts - was it intentional and is it indeed expected?

Src/WindowsServer/WindowsServer.Shared.Tests/FirstChanceExceptionStatisticsTelemetryModuleTest.cs
```diff
@@ -300,6 +300,118 @@ public void FirstChanceExceptionStatisticsTelemetryModuleWillNotDimCapTheSameOpe
...
+                {
+                    countThrottled++;
+                }
+                else
+                {
+                    countProcessed++;
+                }
+            }
+
+            Assert.Equal(166, countProcessed);
```
Dmitry-Matveev: These values (166 & 34) do not look entirely obvious from the context of the test, is there a quick explanation we can add here to make this test easier to maintain for new contributors if any?

Src/WindowsServer/WindowsServer.Shared.Tests/FirstChanceExceptionStatisticsTelemetryModuleTest.cs
dnduffy: It looks like what used to be multiple items became properties on the same item (seems good to me).

kstevenham:
For item 2, what would be preferable, keeping the literal values and just add a comment or getting rid of the literal values and adding the formula code to the test?

Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -77,14 +98,14 @@ public void Initialize(TelemetryConfiguration configuration)
...
-                        this.isInitialized = true;
-
                         this.telemetryClient = new TelemetryClient(configuration);
                         this.telemetryClient.Context.GetInternalContext().SdkVersion = SdkVersionUtils.GetSdkVersion("exstat:");
 
                         this.metricManager = new MetricManager(this.telemetryClient);
 
                         this.registerAction(this.CalculateStatistics);
+
+                        this.isInitialized = true;
```
dnduffy: Won't moving this down increase the risk of re-entry by another thread (not that it locks so it's racing either way)?

Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -248,21 +321,100 @@ private void TrackStatistics(string type, string operation, string method, bool
...
+                    cache3.RwLock.ExitWriteLock();
+                }
+
+                cache2.RwLock.ExitWriteLock();
+            }
+
+            cache1.RwLock.ExitWriteLock();
+        }
+
+        internal class HashCache<T> : IDisposable
```
dnduffy: Something in the back of my mind is nagging me that this might already exist somewhere in the .NET framework...

Src/WindowsServer/WindowsServer.Shared.Tests/FirstChanceExceptionStatisticsTelemetryModuleTest.cs
Dmitry-Matveev: I would think that either will suffice. To avoid chancing numbers in the test in the future we can put a formula here. On the other hand, the formula in the test will be testing the identical formula in the code. So, if both have the bug - it might go undetected. From such a perspective, it looks like we can go with the simple comment here.

Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
kstevenham: From my testing, this is less overhead and more performant than using Concurrent Dictionary

---
In reply to: [107244751](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/370#discussion_r107244751) [](ancestors = 107244751)

Src/WindowsServer/WindowsServer.Shared.Tests/FirstChanceExceptionStatisticsTelemetryModuleTest.cs
kstevenham: Actually Sergey must have made the change to the code that affected this but didn't change the test case. Or there is some .Net bug? Either way, Sergey and I reviewed this and decided to adjust the test case.

---
In reply to: [107207496](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/370#discussion_r107207496) [](ancestors = 107207496)

Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
kstevenham: If this isn't at the end then you have a race condition

---
In reply to: [107222599](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/370#discussion_r107222599) [](ancestors = 107222599)

PR #371, opened 2017-03-20
Remove unused code

lmolkova:
There were some helpers methods:
- TelemetryExtensionsForDependencyCollector
- WebRequestDependencyTrackingHelpers 

They have never been used by any production code. They seem to be an artifact of previous implementation.

They have been tested by corresponding tests.


SergeyKanzhelev:
Technically these were public functions. Can you please keep them with `[Obsolete("These functions will do nothing.")]`

Dmitry-Matveev:
Actually, I think I missed it an I agree with Sergey. @lmolkova, could you please return those, so we do not break public API and just have them as obsolete. We can remove them in the next major version then.

SergeyKanzhelev:
@Dmitry-Matveev we can remove the logic inside. As far as I know nobody used those ever

lmolkova:
I did not realize these were public.

I wonder if someone still uses them and do not configure profiler, there will be no correlation for them.
Let me just mark them as obsolete and let's keep the implementation for 2.4.0 version.

We probably will have a story to support correlation (with header injection) without profiler for both: WebRequest and HttpClient. I'll know more in a few days. If it will work, I'll clean them up.

lmolkova:
`WebRequestDependencyTrackingHelpers`:

```C#
                if (dependencyTelemetry.Context.User.Id != null)
                {
                    CreateAndAddCookie(webRequest, WebUserCookieName, dependencyTelemetry.Context.User.Id);
                }

                if (dependencyTelemetry.Context.Session.Id != null)
                {
                    CreateAndAddCookie(webRequest, WebSessionCookieName, dependencyTelemetry.Context.Session.Id);
                }
```
It seems ProfilerHttpProcessing never set cookies. Is it part of some supported scenario?

SergeyKanzhelev:
Closing for now. It was a long ago we wanted to improve correlation story, No need to keep the functionality, but keep the public methods to not break compilation

Issue #175, opened 2016-10-18
REST API app can have all requests filtered out


SergeyKanzhelev:
Reported by the customer:

> What I notice when running locally, the request fails it contains the response code but when its successful then it doesn’t I am not sure why?
> 
> This is RESTApi Project created using Visual Studio template for Azure Cloud Service -> WebRole -> Empty (Web API Check box) targeting .NET 4.5.2. Project just contain REST Endpoints there is no MVC used. Application Insights configured using Visual Studio (right click on the project and select add AI). 

Removing filters from the collection helped to resolve the issue:

``` xml
<Add Type="Microsoft.ApplicationInsights.Web.RequestTrackingTelemetryModule, Microsoft.AI.Web">
  <Handlers>
<!-- 
        Add entries here to filter out additional handlers: 

        NOTE: handler configuration will be lost upon NuGet upgrade.

     Remove handlers below:
-->
     <Add>System.Web.Handlers.TransferRequestHandler</Add>
     <Add>Microsoft.VisualStudio.Web.PageInspector.Runtime.Tracing.RequestDataHttpHandler</Add>
     <Add>System.Web.StaticFileHandler</Add>
     <Add>System.Web.Handlers.AssemblyResourceLoader</Add>
     <Add>System.Web.Optimization.BundleHandler</Add>
     <Add>System.Web.Script.Services.ScriptHandlerFactory</Add>
     <Add>System.Web.Handlers.TraceHandler</Add>
     <Add>System.Web.Services.Discovery.DiscoveryRequestHandler</Add>
     <Add>System.Web.HttpDebugHandler</Add>
    </Handlers>
</Add>
```

SergeyKanzhelev:
BTW, this issue may be a continuation of this one: https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/16

codecadwallader:
I was having the same issue, and also saw that commenting out the one line in the ApplicationInsights.config caused successful requests to start getting logged.  Web API 2 project.

```xml
        <!--<Add>System.Web.Handlers.TransferRequestHandler</Add>-->
```

SergeyKanzhelev:
@codecadwallader thanks for reporting!

cijothomas:
Not reproed. With the default AI Config I am getting all (sucess,failed) requests in WebApi project.

cijothomas:
[...]

PR #376, opened 2017-03-22
Tricks around System.Web.Handlers.TransferRequestHandler for WCF and WebServices

iusafaro:
REST API app can have all requests filtered out #175

### DIFF
--- a/Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
+++ b/Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
@@ -230,10 +230,21 @@
                 return false;
             }
 
-            if (httpContext.Response.StatusCode < 400)
+            if (httpContext.Response.StatusCode < 400 && httpContext.Handler != null)
             {
-                if (this.IsHandlerToFilter(httpContext.Handler))
+                string handlerName = httpContext.Handler.GetType().FullName;
+                if (this.IsHandlerToFilter(handlerName))
                 {
+                    return false;
+                }
+
+                string requestPath = httpContext.Request.UnvalidatedGetPath();
+                if (string.Equals(handlerName, RequestTrackingConstants.TransferHandlerType, StringComparison.Ordinal)
+                    && (requestPath.IndexOf(".svc", StringComparison.OrdinalIgnoreCase) != -1 || requestPath.IndexOf(".asmx", StringComparison.OrdinalIgnoreCase) != -1))
+                {
+                    // additionally filter out System.Web.Handlers.TransferRequestHandler for WCF applications 
+                    // and Web Services as it leads to duplicate requests
+                    WebEventSource.Log.WebRequestFilteredOutByRequestHandler();
                     return false;
                 }
             }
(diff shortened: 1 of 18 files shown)


Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -230,12 +230,23 @@ internal bool NeedProcessRequest(HttpContext httpContext)
...
             {
-                if (this.IsHandlerToFilter(httpContext.Handler))
+                string handlerName = httpContext.Handler.GetType().FullName;
+                if (this.IsHandlerToFilter(handlerName))
                 {
                     return false;
                 }
+
+                string requestPath = httpContext.Request.UnvalidatedGetPath();
+                if (string.Equals(handlerName, RequestTrackingConstants.TransferHandlerType, StringComparison.Ordinal)
```
Dmitry-Matveev: These comparisons seem to happen for every successful request; can we think of some tweaks for performance of those three comparisons?
- "System.Web.Handlers" seems to be a common part for many handlers (but at the same time protects us from disabling similarly named handler in the other namespace), will using Hash comparison be faster (maybe not...)?
- We seem to use full path of the request for indexOf(), is there any parsed request path object available where we can only use part of the path for indexOf()?

SergeyKanzhelev:
@tomasr do you have any ideas how to fix this better and why this may be the case? The problem is that `TransferHandler` sometimes will server as an "extra" request (it will have it's own httpmodule.End call) and in other cases - it will be set as a handler for the actual application response.

SergeyKanzhelev:
@drago-draganov can you help here? Or you know who we can ask about it?

drago-draganov:
TransferRequestHandler is an HttpHandler used to intercept extensionless URLs. Registered as ExtensionlessUrlHandler-Integrated-4.0 in web.config. It makes possible requests without file extension to enter the managed pipeline. Remember that without properly mapped managed handler, IIS will not invoke (by default) any managed module due to optimization.
TransferRequestHandler doesn't do much by itself. If the request doesn't change the handler (ex: MVC/Web API/Web Routing), it will come down to TransferRequestHandler. At that point it will schedule child request (internally) in such way that preserves the original request, but TransferRequestHandler is excluded of that child request mapping. This new child request starts the pipeline again and proceeds accordingly. The parent request moves to completed at this point. This is the reason you see "extra" request. 
Since TransferRequestHandler doesn't serve any response, the whole reason is to give a chance to other handlers (ex. StaticFileHandler) to take care for the actual response in case remapping didn't happen.

iusafaro:
@drago-draganov, thanks for the explanation! So, for WCF I see duplicate requests and that perfectly makes sense now, for MVC/Web API I don't see TransferRequestHandler being called, but in some applications I see only TransferRequestHandler called and no other handler being executed in the pipeline. Can OWIN make any difference here? - seems that last case reproduced only on OWIN applications. Can you give us any recommendations how to differentiate these?

https://github.com/Microsoft/ApplicationInsights-Home/files/841553/AINotWorking.zip

drago-draganov:
@iusafaro
MVC/Web API case:
When the initial extensionless URL request comes into the pipeline IIS sees TransferRequestHandler mapped ("*." path mapping). It tells IIS that the request is "managed", so the managed pipeline can interfere. Then MVC/Web API remaps (replaces) the http handler (at that moment is still TransferRequestHandler) during PostResolveRequestCache with their own.  So TransferRequestHandler goes away from the scene and the request is handled by MVC/Web API http handler.

Only TransferRequestHandler case:
The initiated child request either completes early or it's mapped and handled by native modules (ex: StaticFile, DefaultDocument, DirectoryListing), so managed code didn't see it. I would also check if ExtensionlessUrlHandler is configured to map to path="*", instead of path="*.". I have seen this with before with OData apps.

TransferRequestHandler acts as a stop plug for extensionless urls. It's executed only when no one set own http handler before MapRequestHandler.

Issue #294, opened 2017-01-16
Also target HttpContextBase in HttpContextExtension extension methods


snboisen:
A long time ago, there was only HttpContext, which was not mockable and so developers cried in anguish. Then came .NET 3.5 and with it HttpContextBase and HttpContextWrapper which opened the context up to mocking and there was much rejoicing.

Anyway, AFAIK using HttpContext.Current is not recommended under ASP.NET MVC and WebApi. If we absolutely need to access it, it should be done using `request.Properties["MS_HttpContext"]`. The issue here is that this place holds an instance of HttpContextWrapper and HttpContextWrapper does not allow access to the wrapped HttpContext. Therefore the provided extension method `GetRequestTelemetry` in [HttpContextExtension](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/Web/Web.Shared.Net/HttpContextExtension.cs) is useless when accessing the context in this way.

It would be nice to have the current and future methods here available for HttpContextBase.

dnduffy:
The extension method will take whatever HttpContext you pass it, where you get the HttpContext instance from is not relevant to the extension method.
As for getting it in MVC 5 and older, you can still get HttpContext.Current in a controller action but you need to be aware that if you don't qualify it with System.Web.HttpContext.Current it will instead resolve to a local property on the controller with the same name but of type HttpContextBase.
In MVC 6 I have read that Current was removed but I believe you can get it from HttpContextAccessor.HttpContext or with "ContextBase.Current as HttpContext".
Please feel free to reactive and provide more information if this doesn't work for you.

snboisen:
I am writing a pure ASP.NET WebAPI 2 webservice (no MVC). I don't seem to have access to `HttpContextAccessor` or `ContextBase`? `HttpContextBase` does not have a `Current` property.

So as I understand it, the only way I can to `HttpContext` in WebAPI is through `HttpContext.Current`?

SergeyKanzhelev:
[...]

PR #377, opened 2017-03-22
Add GetRequestTelemetry extension method to HttpContextBase

iusafaro:
Also target HttpContextBase in HttpContextExtension extension methods #294


Src/Web/Web.Shared.Net/HttpContextBaseExtension.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// <param name="context">HttpContextBase instance.</param>
+        /// <returns>Request telemetry instance or null.</returns>
+        public static RequestTelemetry GetRequestTelemetry(this HttpContextBase context)
+        {
+            if (context == null || context.ApplicationInstance == null || context.ApplicationInstance.Context == null)
+            {
+                return null;
+            }
+
+            return context.ApplicationInstance.Context.Items[RequestTrackingConstants.RequestTelemetryItemName] as RequestTelemetry;
```
Dmitry-Matveev: Is there a case when ApplicationInstance will throw NotImplmented if not overridden as per [MSDN](https://msdn.microsoft.com/en-us/library/system.web.httpcontextbase.applicationinstance(v=vs.110).aspx)? Or it's always overridden and implemented in the scenarios AI used in?

iusafaro: In regular Web API application it works as expected and doesn't throw NotImplemented exception and this is used only as extension called by customer - we're not using it internally anywhere, so we're safe here.

Issue #379, opened 2017-03-24
After upgrading to 2.3 Beta3 started seeing decorated dependencies QP endpoint


AlexBulankou:
Immediately after deploying the service (InsightsPortal) that was upgraded to 2.3 Beta3 I started seeing the following dependencies from the following QP endpoint appearing in our telemetry:

quickpulse-ppe.aisvc.visualstudio.com | I5UqryrMvK+2zqbkrtAaUyMtQB0wPvhKJjcXYr7b7ic=

![image](https://cloud.githubusercontent.com/assets/11968377/24275884/18a32fd6-0ff0-11e7-8e7a-725968f9ae85.png)

I think what it means is that it is unfiltered dependency from QP service that is onboarded to 2.3 Beta1/2 and the reason it is not filtered because it checks against "quickpulse-ppe.aisvc.visualstudio.com" but not against "quickpulse-ppe.aisvc.visualstudio.com | I5UqryrMvK+2zqbkrtAaUyMtQB0wPvhKJjcXYr7b7ic="
The fix would be to modify the check to consider that target will be modified with postfix.

@Dmitry-Matveev, @nizarq

iusafaro:
Fixed by using dependency target instead of name.

Issue #356, opened 2017-03-13
[Exceptions Statistics] Live Metrics threads should be marked as Internal operation


SergeyKanzhelev:
I see a few exceptions in exceptions statistics from the live stream feature. Please mark live stream threads as SDK Internal so they would be marked as "Internal" for statistics

https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/PerformanceCollector/Shared/PerformanceCollectorModule.cs#L235

SergeyKanzhelev:
Marking for 2.3 unless it is a big risk for the stable SDK

Dmitry-Matveev:
@tokaplan , were you planning to implement this for 2.3? What would be your risk assessment?
Note: There is a bigger discussion on enabling/disabling exception statistics for 2.3. If the answer is to disable it, this issue might be postponed as well into 2.4.

PR #380, opened 2017-03-24
Prevent collection of decorated dependencies from QP endpoint

iusafaro:
After upgrading to 2.3 Beta3 started seeing decorated dependencies QP endpoint #379


SergeyKanzhelev:
I'd suggest to add an additional safeguard - implement this https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/356 and in the filter check whether we are inside the internal operation and do not track dependency if we are

PR #385, opened 2017-03-29
Upgrade Core SDK to 2.4.0-beta1-build06239

lmolkova:


Src/DependencyCollector/Shared.Tests/Implementation/TelemetryExtensionsForDependencyCollectorTests.cs
```diff
@@ -34,15 +35,16 @@ public void TestInitialize()
             DependencyTrackingTelemetryModule module = new DependencyTrackingTelemetryModule();
             module.Initialize(configuration);
             this.telemetryClient = new TelemetryClient(configuration);
-            var operationHolder = this.telemetryClient.StartOperation<DependencyTelemetry>("operationName");
```
lmolkova: there is a bug in Activity: https://github.com/dotnet/corefx/pull/17676 (is it not Serializable) that causes nasty problems with MsTest when Activity was not cleaned properly (it tries to move activity.current across app-domains and fails).

Even though, bug will be fixed, it's always good to clean up after tests

PR #386, opened 2017-03-30
Introduces a check that stops retries if the error when fetching appI…

nizarq:
…D is 4XX as such an error is unlikely to resolve with a retry.

### DIFF
--- a/Src/Common/CorrelationIdLookupHelper.cs
+++ b/Src/Common/CorrelationIdLookupHelper.cs
@@ -30,14 +30,14 @@
 
         // We have arbitrarily chosen 5 second delay between trying to get app Id once we get a failure while trying to get it. 
         // This is to throttle tries between failures to safeguard against performance hits. The impact would be that telemetry generated during this interval would not have x-component correlation id.
-        private readonly TimeSpan intervalBetweenFailedRetries = TimeSpan.FromSeconds(5);
+        private readonly TimeSpan intervalBetweenFailedRetries = TimeSpan.FromSeconds(30);
 
         private Uri endpointAddress;
 
         private ConcurrentDictionary<string, string> knownCorrelationIds = new ConcurrentDictionary<string, string>();
 
         // Stores failed instrumentation keys along with the time we tried to retrieve them.
-        private ConcurrentDictionary<string, DateTime> failingInstrumenationKeys = new ConcurrentDictionary<string, DateTime>();
+        private ConcurrentDictionary<string, FailedResult> failingInstrumenationKeys = new ConcurrentDictionary<string, FailedResult>();
 
         private Func<string, Task<string>> provideAppId;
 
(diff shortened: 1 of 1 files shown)


Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -30,14 +30,14 @@ internal class CorrelationIdLookupHelper
 
         // We have arbitrarily chosen 5 second delay between trying to get app Id once we get a failure while trying to get it.
```
pharring: Comment is now incorrect. Change "5" to "30"

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -209,5 +229,56 @@ private string GetExceptionDetailString(Exception ex)
...
+                this.FailureTime = failureTime;
+
+                // Unknown failure code.
+                this.FailureCode = int.MinValue;
+            }
+
+            /// <summary>
+            /// Gets the time of failure.
+            /// </summary>
+            public DateTime FailureTime { get; private set; }
```
pharring: Make it a getter-only autoprop. i.e. remove the "private set;" since you only set it in the ctor.

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -209,5 +229,56 @@ private string GetExceptionDetailString(Exception ex)
...
+
+            /// <summary>
+            /// Gets the integer value for response code representing the type of failure.
+            /// </summary>
+            public int FailureCode { get; private set; }
```
pharring: Make it a getter-only autoprop. i.e. remove the "private set;" since you only set it in the ctor.

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -209,5 +229,56 @@ private string GetExceptionDetailString(Exception ex)
...
+
+            /// <summary>
+            /// Gets a value indicating whether the failure is likely to go away when a retry happens.
+            /// </summary>
+            public bool ShouldRetry
```
pharring: Nit: use expression-bodied property. i.e.
```C#
public bool ShouldRetry => !(this.FailureCode >= 400 && this.FailureCode < 500);

File: Src/Common/CorrelationIdLookupHelper.cs [hunk 5]
```diff
@@ -199,6 +197,28 @@ private Uri GetAppIdEndPointUri(string instrumentationKey)
...
+            var webException = ex as WebException;
+            if (webException != null)
+            {
+                this.failingInstrumenationKeys[instrumentationKey] = new FailedResult(
+                    DateTime.UtcNow,
+                    ((HttpWebResponse)webException.Response).StatusCode);
+            }
+            else
+            {
+                this.failingInstrumenationKeys[instrumentationKey] = new FailedResult(DateTime.UtcNow);
```
pharring: Suggestion:  Add a ```FailedResult``` constructor that takes an exception and move this logic there. This method then becomes a one-liner and you can remove the other unused constructors.
```C#
this.failingInstrumentationKeys[instrumentationKey] = new FailedResult(DateTime.UtcNow, ex);

pharring:
@nizarq please see my comments.

Issue #381, opened 2017-03-27
Initialize has not been called on this module yet


haldiggs:
recently we started seeing this exception being raised when an application is first being loaded after an appPool cycle. Has something changed with the telemetryClient init?

    public void OnError(HttpContext context)
    {
            if (this.telemetryClient == null)
            {
                throw new InvalidOperationException("Initialize has not been called on this module yet.");
            }
shouldn't this thing get initialized somewhere rather than just throwing an exception? It doesn't seem like I have anyway to deal with this except to tap the site a 2nd time and tell my users to deal with it.

SergeyKanzhelev:
My guess is that this may happen when the initialization of any other module has failed before the module initialization. Please check the telemetry - look for the traces with the "AI:" prefix indicating configuration error.

We should definitely fix the logic to not throw in this case. It is a clear violation of SDK error handling policy

lmolkova:
@SergeyKanzhelev it affects also feature I'm working on.
What is the most appropriate thing to do? Write error to event log or just silently swallow it?

SergeyKanzhelev:
@lmolkova are you talking about `PreRequestHandler` callback? I'd say to trace `Error` at first so it will be visible for customer by default and then `Verbose` without crashing the request.

haldiggs:
but the bigger issue is that this is throwing errors into the event log and there is nothing I can do to recover from it. Why the object has not been created? Its not as if there is anything I can change/call from within my application is there?

This is just odd because its being thrown extremely often and causing my IT department headaches. They see this as an IIS crash issue. Not really a good thing.

SergeyKanzhelev:
@haldiggs I understand that the issue is critical and we will try to at least remove throwing logic as soon as we can. 

As a workaround - have you ensured that the `ApplicationInsights.config` is correct and there are no errors reported by Application Insights SDK? You can also try to run this code manually in Global.asax:

``` csharp
foreach (var module in Microsoft.ApplicationInsights.Extensibility.Implementation.TelemetryModules.Instance.Modules)
{
  module.Initializer(TelemetryConfiguration.Active);
}
```

I'm sorry for the headaches this issue causing.

haldiggs:
thanks for the work around

SergeyKanzhelev:
@haldiggs, can you please confirm that the work around help. We will definitely fix it in upcoming versions

haldiggs:
[...]

PR #390, opened 2017-03-30
Do not throw from HttpModule

SergeyKanzhelev:
Quick fix for https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/381 for 2.3


kstevenham:
It looks like you would be sending an error message for every request? If that is what would happen, I don't think you'd want to do that as it would kill perf. Sending once should be sufficient.

SergeyKanzhelev:
@kstevenham we have a throttling limits. It will be no more than 3 errors in an hour or something like this. And it should never happen =).

I agree, let me fix it...

SergeyKanzhelev:
@kstevenham better now?

PR #396, opened 2017-03-31
Implement request-dependency correlation with Activity

lmolkova:
This change implements telemetry correlation
1. Supports [HTTP protocol](https://github.com/lmolkova/correlation/blob/master/http_protocol_proposal_v1.md)
2. Uses [Activity](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/ActivityUserGuide.md) to store and pass correlation Ids and context

--------
Depends on https://github.com/Microsoft/ApplicationInsights-dotnet/pull/505. Do not merge until base changes are approved

### DIFF
--- a/Src/Common/AppInsightsActivity.cs
+++ b/Src/Common/AppInsightsActivity.cs
@@ -1,0 +1,145 @@
+﻿namespace Microsoft.ApplicationInsights.Common
+{
+    using System;
+    using System.Diagnostics;
+    using System.Threading;
+
+    // this is a temporary solution that mimics System.Diagnostics.Activity and Correlation HTTP protocol:
+    // https://github.com/lmolkova/correlation/blob/master/http_protocol_proposal_v1.md
+    // it is copied from https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs
+    // and intended to be used on .NET 4.0
+    internal class AppInsightsActivity
+    {
+        private const int RequestIdMaxLength = 1024;
+
+        // Used to generate an ID:
+        // instance unique prefix
+        private static string uniqPrefix;
+
+        // A unique number inside the appdomain, randomized between appdomains. 
+        // Int gives enough randomization and keeps hex-encoded s_currentRootId 8 chars long for most applications
+        private static long currentRootId = (uint)GetRandomNumber();
+
+        public static string GenerateRequestId(string parentId = null)
+        {
+            string ret;
+            if (parentId != null && parentId.Length != 0)
+            {
+                // Start from outside the process (e.g. incoming HTTP)
+                // sanitize external RequestId as it may not be hierarchical. 
+                // we cannot update ParentId, we must let it be logged exactly as it was passed.
+                parentId = parentId[0] == '|' ? parentId : '|' + parentId;
+                if (parentId[parentId.Length - 1] != '.')
+                {
+                    parentId += '.';
+                }
+
+                ret = AppendSuffix(parentId, Interlocked.Increment(ref currentRootId).ToString("x"), '_');
+            }
+            else
+            {
+                // A Root Activity (no parent).  
+                ret = GenerateRootId();
+            }
+
+            // Useful place to place a conditional breakpoint.  
+            return ret;
+        }
+
+        public static string GenerateDependencyId(string parentId)
+        {
+            string ret;
+            if (parentId != null && parentId.Length != 0)
+            {
+                // Start from outside the process (e.g. incoming HTTP)
+                // sanitize external RequestId as it may not be hierarchical. 
+                // we cannot update ParentId, we must let it be logged exactly as it was passed.
+                parentId = parentId[0] == '|' ? parentId : '|' + parentId;
+                if (parentId[parentId.Length - 1] != '.')
+                {
+                    parentId += '.';
+                }
+
+                ret = AppendSuffix(parentId, Interlocked.Increment(ref currentRootId).ToString("x"), '_');
+            }
+            else
+            {
+                // A Root Activity (no parent).  
+                ret = GenerateRootId();
+            }
+
+            // Useful place to place a conditional breakpoint.  
+            return ret;
+        }
+
+        public static string GetRootId(string id)
+        {
+            // id MAY start with '|' and contain '.'. We return substring between them
+            // ParentId MAY NOT have hierarchical structure and we don't know if initially rootId was started with '|',
+            // so we must NOT include first '|' to allow mixed hierarchical and non-hierarchical request id scenarios
+            int rootEnd = id.IndexOf('.');
+            if (rootEnd < 0)
+            {
+                rootEnd = id.Length;
+            }
+
+            int rootStart = id[0] == '|' ? 1 : 0;
+            return id.Substring(rootStart, rootEnd - rootStart);
+        }
+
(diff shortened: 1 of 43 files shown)


Src/Common/WebHeaderCollectionExtensions.cs
```diff
@@ -82,23 +121,84 @@ public static void SetNameValueHeaderValue(this NameValueCollection headers, str
...
                     }
                 }
 
-                headers[headerName] = string.Join(", ", headerValues);
+                headers[headerName] = string.Join(",", headerValues);
 
                 if (!found)
                 {
-                    headers[headerName] += string.Format(CultureInfo.InvariantCulture, ", {0}={1}", keyName, value);
+                    headers[headerName] += string.Format(CultureInfo.InvariantCulture, ",{0}={1}", keyName, value);
```
pharring: Use ```FormatKeyValueHeader``` here too? i.e.
```C#
headers[headerName] = headers[headerName] + "," + FormatKeyValueHeader(keyName, value);

File: Src/Common/WebHeaderCollectionExtensions.cs [hunk 2]
```diff
@@ -82,23 +121,84 @@ public static void SetNameValueHeaderValue(this NameValueCollection headers, str
...
+                throw new ArgumentNullException(nameof(headerName));
+            }
+
+            var requiredHeader = headers[headerName];
+
+            // do not set header if it's present
+            if (string.IsNullOrEmpty(requiredHeader))
+            {
+                StringBuilder headerValue = new StringBuilder();
+                foreach (var pair in keyValuePairs)
```
pharring:
```C#
var headerValue = string.Join(',', keyValuePairs.Select(pair => FormatKeyValueHeader(pair.Key, pair.Value));

File: Src/Common/WebHeaderCollectionExtensions.cs [hunk 3]
```diff
@@ -82,23 +121,84 @@ public static void SetNameValueHeaderValue(this NameValueCollection headers, str
...
+                {
+                    headers[headerName] = headerValue.ToString();
+                }
+            }
+        }
+
+        private static bool TryParseKeyValueHeader(string pairString, out string key, out string value)
+        {
+            Debug.Assert(pairString != null, "pairString is null");
+            var keyValue = pairString.Split('=');
```
pharring: ```string.Split``` is expensive in memory allocations. For something as simple as this, you don't need it. The alternative doesn't look 'pretty', but I notice this is called in an inner loop, so it's probably worth optimizing. 
```C#
const char separator = '=';
int indexOfSeparator = pairString.IndexOf(separator, StringComparison.Ordinal);
// Must be a separator, and cannot be first or last
if (indexOfSeparator <= 0 || indexOfSeparator == pairString.Length -1) return false;
// Must be exactly one separator
if (pairString.IndexOf('=', indexOfSeparator + 1, StringComparison.Ordinal) >= 0) return false;
var key = pairString.Substring(0, indexOfSeparator).Trim();
var value = pairString.Substring(indexOfSeparator + 1).Trim();
if (key.Length > 0 && value.Length > 0)
{
    return true;
}

key = null;
value = null;
return false;

File: Src/Common/WebHeaderCollectionExtensions.cs [hunk 4]
```diff
@@ -82,23 +121,84 @@ public static void SetNameValueHeaderValue(this NameValueCollection headers, str
...
+            }
+
+            key = null;
+            value = null;
+            return false;
+        }
+
+        private static string FormatKeyValueHeader(string key, string value)
+        {
+            return string.Format(CultureInfo.InvariantCulture, "{0}={1}", key.Trim(), value.Trim());
```
pharring: Recommend using simple string concatenation here. ```string.Format``` is unnecessary.
```C#
return key.Trim() + "=" + value.Trim();

File: Src/Common/WebHeaderCollectionExtensions.cs (hunk 3 above)
Comment (lmolkova) [inline Src/Common/WebHeaderCollectionExtensions.cs]: Thanks!

File: Src/Common/AppInsightsActivity.cs [hunk 5]
```diff
@@ -0,0 +1,145 @@
+﻿namespace Microsoft.ApplicationInsights.Common
+{
+    using System;
+    using System.Diagnostics;
+    using System.Threading;
+
+    // this is a temporary solution that mimics System.Diagnostics.Activity and Correlation HTTP protocol:
```
dnduffy: nitpick: Thank-you for all the comments, I love comments! :)  Please convert them to XML doc comments.

Src/Common/WebHeaderCollectionExtensions.cs
```diff
@@ -19,10 +22,7 @@ internal static class WebHeaderCollectionExtensions
         /// <returns>Value against the given parameters.</returns>
         public static string GetNameValueHeaderValue(this NameValueCollection headers, string headerName, string keyName)
         {
-            if (string.IsNullOrEmpty(headerName))
-            {
-                throw new ArgumentNullException(nameof(headerName));
-            }
+            Debug.Assert(headerName != null, "headerName must not be null");
```
dnduffy: If a debug build assert is all that is left, what is the point of even passing headerName as a parameter at all?  It wasn't used in the original call either other than to throw if it was null or empty which seems pointless.

lmolkova: headerName is used later:
``
var requiredHeader = headers[headerName];
```

File: Src/Common/AppInsightsActivity.cs (hunk 5 above)
Comment (lmolkova) [inline Src/Common/AppInsightsActivity.cs]: ok :)

File: Src/Common/WebHeaderCollectionExtensions.cs (hunk 6 above)
Comment (dnduffy) [inline Src/Common/WebHeaderCollectionExtensions.cs]: It is used in different functions but not this one that I can see.

Comment (lmolkova) [inline Src/Common/WebHeaderCollectionExtensions.cs]:
```
       /// <summary>
        /// For the given header collection, for a given header of name-value type, find the value of a particular key.
        /// </summary>
        /// <param name="headers">Header collection.</param>
        /// <param name="headerName">Name of the header in the collection.</param>
        /// <param name="keyName">Desired key of the key-value list.</param>
        /// <returns>Value against the given parameters.</returns>
        public static string GetNameValueHeaderValue(this NameValueCollection headers, string headerName, string keyName)
        {
            Debug.Assert(headerName != null, "headerName must not be null");

            if (string.IsNullOrEmpty(keyName))
            {
                throw new ArgumentNullException(nameof(keyName));
            }

            var requiredHeader = headers[headerName];

```

There is one more throw there, I'll replace it with Debug.Assert as well. It could only happen if developer passed null by mistake

File: Src/Common/AppInsightsActivity.cs [hunk 7]
```diff
@@ -0,0 +1,164 @@
...
+
+    // See
+    // https://github.com/lmolkova/correlation/blob/master/http_protocol_proposal_v1.md
+    // https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs
+
+    /// <summary>
+    /// Mimics System.Diagnostics.Activity and Correlation HTTP protocol 
+    /// and intended to be used on .NET 4.0.
+    /// </summary>
+    internal class AppInsightsActivity
```
lmolkova: It's shared between Web.Net40 and DependencyCollector.Net40

PR #397, opened 2017-03-31
Refactored HTTP dependency parsing to use dedicated parsers

TomMilos:


Src/DependencyCollector/Shared/HttpParsers/AzureBlobHttpParser.cs
```diff
@@ -0,0 +1,74 @@
+﻿namespace Microsoft.ApplicationInsights.DependencyCollector.HttpParsers
```
SergeyKanzhelev: We typically place internal classes under `Implementation` folder

TomMilos: Done

SergeyKanzhelev:
@TomMilos you'd need to rebase locally so only your changes will be included

TomMilos:
Branch got corrupted after merges and rebasing

PR #401, opened 2017-04-03
QuickPulse filtering (work/filtering2)

tokaplan:
QuickPulse is now capable of sending metrics and full telemetry documents based on custom collection configuration received from QPS.
QPS will now respond with what we call "collection configuration" to SDK's ping's and post's in order to specify custom criteria for collecting performance counters, full telemetry documents, and calculated metrics based on telemetry items.

Principal new players:
- Everything in Filtering.Shared/Implementation/Service contract - DTO-style data contracts for transmitting collection configuration between SDK and QPS. No business logic.
- CollectionConfiguration. Represents the collection configuration - a customizable description of performance counters, metrics, and full telemetry documents to be collected by the SDK.
- OperationalizedMetric. Represents a single configured metric that needs to be calculated and reported on top of the telemetry items that pass through the pipeline. Includes a set of filters that define which telemetry items to consider, a projection which defines which field to use as a value, and an aggregation which dictates the algorithm of arriving at a single reportable value within a second.
- FilterConjunctionGroup and Filter. Represent an AND-OR structure of filters which define which telemetry items are considered when collecting metrics and sending out full telemetry documents.
- DocumentStream. Represents a concept of a uniquely identifiable set of full telemetry documents that are being reported by the SDK. The notion of a stream is needed since multiple UX sessions might be querying for full telemetry documents with different filtering criteria simultaneously. 
- CollectionConfigurationAccumulator, AccumulatedValue - thread-safe lock-free accumulators that are being updated as telemetry items are passing through the pipeline.
- QuickPulseTelemetryProcessor - an entry point for processing telemetry items.
- QuickPulseMetricProcessor - an entry point for processing aggregated metrics (new API).


Src/PerformanceCollector/Shared/Implementation/QuickPulse/Service contract/TelemetryDocument/EventTelemetryDocument.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Runtime.Serialization;
+
+    [DataContract]
+    internal struct EventTelemetryDocument : ITelemetryDocument
+    {
+        [DataMember(EmitDefaultValue = false)]
+        public Guid Id { get; set; }
```
SergeyKanzhelev: what does this GUID represents? Please do not diverge from schema too much. Remember, you promised to start using built-in types

Src/PerformanceCollector/Filtering/CollectionConfiguration.cs
```diff
@@ -0,0 +1,327 @@
...
+        private readonly List<OperationalizedMetric<ExceptionTelemetry>> exceptionTelemetryMetrics =
+            new List<OperationalizedMetric<ExceptionTelemetry>>();
+
+        private readonly List<OperationalizedMetric<EventTelemetry>> eventTelemetryMetrics = new List<OperationalizedMetric<EventTelemetry>>();
+
+        private readonly List<OperationalizedMetric<TraceTelemetry>> traceTelemetryMetrics = new List<OperationalizedMetric<TraceTelemetry>>();
+
+        private readonly List<OperationalizedMetric<MetricValue>> metricMetrics = new List<OperationalizedMetric<MetricValue>>();
+
+        private readonly List<Tuple<string, AggregationType>> telemetryMetadata = new List<Tuple<string, AggregationType>>();
```
SergeyKanzhelev: why do you use `OperationalizedMetric` above and `Tuple` instead of `AggregatedValue` here?

Src/PerformanceCollector/Filtering/CollectionConfiguration.cs
```diff
@@ -0,0 +1,327 @@
...
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.Implementation.QuickPulse.Helpers;
+
+    internal class CollectionConfiguration
+    {
+        private readonly CollectionConfigurationInfo info;
+
+        private readonly List<OperationalizedMetric<RequestTelemetry>> requestTelemetryMetrics = new List<OperationalizedMetric<RequestTelemetry>>();
+
+        private readonly List<OperationalizedMetric<DependencyTelemetry>> dependencyTelemetryMetrics =
```
SergeyKanzhelev: why do you keep a list per type? Would it be better to make configuration more generic?

Src/PerformanceCollector/Filtering/Implementation/Service contract/CollectionConfigurationInfo.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Filtering
+{
+    using System.Runtime.Serialization;
+
+    /// <summary>
+    /// Represents the entire collection configuration for this SDK.
```
SergeyKanzhelev: description is not descriptive. Can you please clarify what does "entire collection configuration" mean?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/IQuickPulseServiceClient.cs
```diff
@@ -2,7 +2,8 @@
 {
     using System;
     using System.Collections.Generic;
-    using System.Security.Cryptography.X509Certificates;
+
+    using Microsoft.ApplicationInsights.Extensibility.Filtering;
```
SergeyKanzhelev: why the namespace called "Filtering"?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/IQuickPulseServiceClient.cs
```diff
@@ -16,15 +17,33 @@ internal interface IQuickPulseServiceClient
         /// </summary>
         /// <param name="instrumentationKey">InstrumentationKey for which to submit data samples.</param>
         /// <param name="timestamp">Timestamp to pass to the server.</param>
+        /// <param name="configurationETag">Current configuration ETag that the client has.</param>
+        /// <param name="authApiKey">Authentication API key.</param>
+        /// <param name="configurationInfo">When available, the deserialized response data received from the server.</param>
         /// <returns><b>true</b> if data is expected, otherwise <b>false</b>.</returns>
-        bool? Ping(string instrumentationKey, DateTimeOffset timestamp);
+        bool? Ping(
```
SergeyKanzhelev: Looking at this interface the best name would be `TryGetConfiguration`. Since you are hiding an important detail that this method is actually used to transmit `roleInstance` name used for visualization - this renaming would be logical. Please re-think the interface.

Another idea - why wouldn't you initialize the new client whenever you want a new `instrumentationKey`, or `etag` or `authApiKey` to be transmitted?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/IQuickPulseServiceClient.cs
```diff
@@ -16,15 +17,33 @@ internal interface IQuickPulseServiceClient
...
+        /// <param name="collectionConfigurationErrors">Errors to be reported back to the server.</param>
         /// <returns><b>true</b> if the client is expected to keep sending data samples, <b>false</b> otherwise.</returns>
-        bool? SubmitSamples(IEnumerable<QuickPulseDataSample> samples, string instrumentationKey);
+        bool? SubmitSamples(
+            IEnumerable<QuickPulseDataSample> samples,
+            string instrumentationKey,
+            string configurationETag,
+            string authApiKey,
+            out CollectionConfigurationInfo configurationInfo,
+            CollectionConfigurationError[] collectionConfigurationErrors);
```
SergeyKanzhelev: why errors are not a part of `samples`? You already mixing up counters and events in a single collection. Why another abstraction here?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
```diff
@@ -180,6 +233,37 @@ public TimeSpan UpdateState(string instrumentationKey)
...
+                            Tuple.Create("ETag", configurationInfo.ETag)));
+                }
+
+                if (errors != null)
+                {
+                    this.collectionConfigurationErrors.AddRange(errors);
+                }
+
+                this.currentConfigurationETag = configurationInfo.ETag;
+            }
```
SergeyKanzhelev: what if the new `configurationIfnfo` is `null`? Should the old configuration be cleared?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
```diff
@@ -180,6 +233,37 @@ public TimeSpan UpdateState(string instrumentationKey)
             return this.DetermineBackOffs();
         }
 
+        private void UpdateConfiguration(CollectionConfigurationInfo configurationInfo)
+        {
+            // we only get here if Etag in the header is different from the current one, but we still want to check if Etag in the body is also different
+            if (configurationInfo != null && !string.Equals(configurationInfo.ETag, this.currentConfigurationETag, StringComparison.Ordinal))
+            {
+                this.collectionConfigurationErrors.Clear();
```
SergeyKanzhelev: This is not the best pattern for applying the configuration. Is there any reason to not use "snapsots" for configuration rather than messing up with instantiated config and track state changes?

Src/PerformanceCollector/Shared/Implementation/IPerformanceCollector.cs
```diff
@@ -31,5 +30,12 @@ internal interface IPerformanceCollector
         /// <param name="error">Captures the error logged.</param>
         /// <param name="blockCounterWithInstancePlaceHolder">Boolean that controls the registry of the counter based on the availability of instance place holder.</param>
         void RegisterCounter(string perfCounter, string reportAs, bool isCustomCounter, out string error, bool blockCounterWithInstancePlaceHolder);
+
+        /// <summary>
+        /// Removes a counter.
+        /// </summary>
+        /// <param name="perfCounter">Name of the performance counter to remove.</param>
+        /// <param name="reportAs">ReportAs value of the counter to remove.</param>
+        void RemoveCounter(string perfCounter, string reportAs);
```
SergeyKanzhelev: Consider using snapshot approach for configuring. Instead of adding and removing counters - just drop the old collector and instantiate the new one. This way you'd ensure better consistency between what customer wanted to configure and what is actually configured

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
```diff
@@ -125,8 +146,33 @@ public TimeSpan UpdateState(string instrumentationKey)
...
+                    {
+                        ////!!! better solution?
+                        // some samples are still being processed, wait a little to give them a chance to finish
+                        Thread.Sleep(this.coolDownTimeout);
+
+                        bool allCooledDown =
+                            dataSamplesToSubmit.All(sample => Interlocked.Read(ref sample.CollectionConfigurationAccumulator.ReferenceCount) == 0);
+
+                        QuickPulseEventSource.Log.TroubleshootingMessageEvent(
+                            string.Format(
```
SergeyKanzhelev: do not use `string.Format` in event source messages. Event source already allows to have parametrized messages

Src/PerformanceCollector/Filtering/CollectionConfigurationAccumulator.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// so the consumer should keep the reference to it post-swap and make the best effort not to send
+    /// prematurely. <see cref="ReferenceCount"/> indicates that the accumulator is still being processed
+    /// when non-zero.
+    /// </summary>
+    internal class CollectionConfigurationAccumulator
+    {
+        /// <summary>
+        /// Used by writers to indicate that a processing operation is still in progress.
+        /// </summary>
+        public long ReferenceCount = 0;
```
SergeyKanzhelev: Consider protecting the ReferenceCount by exposing interlocked methods, not relying on callers to do it.

Src/PerformanceCollector/Filtering/CollectionConfigurationAccumulator.cs
```diff
@@ -0,0 +1,43 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Filtering
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    /// <summary>
+    /// Represents the operationalized metric-related component of the QuickPulse accumulator.
```
SergeyKanzhelev: what is "operationalized metric-related component"?

Src/PerformanceCollector/Filtering/DocumentStream.cs
```diff
@@ -0,0 +1,237 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Globalization;
+
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.Implementation.QuickPulse.Helpers;
+
+    /// <summary>
+    /// Document stream defines a stream of full telemetry documents that need to be collected and reported.
```
SergeyKanzhelev: it is not a stream as it's not inherited from `Stream`. Can you please clarify the comment defining what does stream means

Src/PerformanceCollector/Filtering/AccumulatedValue.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Filtering
+{
+    using System.Collections.Concurrent;
+
+    /// <summary>
+    /// Accumulator for operationalized metrics.
```
tokaplan: What would be your preferred name as of right now?

Src/PerformanceCollector/Filtering/AccumulatedValue.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// </summary>
+    internal class AccumulatedValue
+    {
+        public readonly ConcurrentStack<double> Value = new ConcurrentStack<double>();
```
tokaplan: This particular accumulator works by accumulating values as they pass through the pipeline in a list. The aggregation happens once a second elsewhere, this is just a DTO. The word "to accumulate" is used in its original meaning, which is "to gather together".

Src/PerformanceCollector/Filtering/AccumulatedValue.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        public AccumulatedValue(string metricId, AggregationType aggregationType)
+        {
+            this.MetricId = metricId;
+            this.AggregationType = aggregationType;
+        }
+
+        public string MetricId { get; }
```
tokaplan: We could, but then it would need to be a field. I'd rather have a get-only property than a readonly public field, that feels weird :)

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Service contract/TelemetryDocument/EventTelemetryDocument.cs
tokaplan: This has been there from the beginning of full telemetry documents in QuckPulse - I planned to use it for internal QuickPulse-deduping originally. We ended up not using it, I'll drop it in the next stable version if it's still useless.

Src/PerformanceCollector/Filtering/CollectionConfiguration.cs
tokaplan: These are used on the perf-cricital path inside QuickPulseTelemetryProcessor. I'm fighting for speed there, so I only determine the actual telemetry item type once and then use if's and generic methods to process. We're not likely to have many more telemetry types, so I think it's worth the effort.

Src/PerformanceCollector/Filtering/CollectionConfiguration.cs
tokaplan: telemetryMetadata here is different from everything that is above in that it's not actually used to filter and calculate things in runtime (as telemetry items are coming in), but rather it's used to create the accumulator to which data is stored at runtime. Hence it's called "metadata". It serves as a single collection which allows the accumulator to generate an internal dictionary which includes all necessary sub-accumulators to hold all metrics that we're about to collect.
Contrary to that, AggregatedValue is a runtime class which is used to actually store data.

Src/PerformanceCollector/Filtering/Implementation/Service contract/CollectionConfigurationInfo.cs
tokaplan: Done.

DTO that represents the collection configuration - a customizable description of performance counters, metrics, and full telemetry documents to be collected by the SDK. Processed and encapsulated by <see cref="CollectionConfiguration"/> at the time of actual collection.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/IQuickPulseServiceClient.cs
tokaplan: Same reason as the shared project is called "filtering" - the original tag by which this functionality was introduced was "filtering telemetry to calculate metrics". Can you think of a better name? This one does seem a bit outdated at this point.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/IQuickPulseServiceClient.cs
tokaplan: Well.
The reason why QuickPulseServiceClient exists is to communicate with QPS. QPS supports two major operations - PING and POST. For POST, I call it SubmitSamples in the interface (since POST is confusing on many levels :) ). For PING, I just use Ping, because that's essentially what it does - informs the service that we're alive.
Of course, it also does many other things, like passing local timestamp and identifying information, etc. It now also supplies configuration etag and reads an updated configuration back if present in the response. Still, it's the QPS's PING operation, and I think it makes sense to call it that.
It never made any particular sense to me to create a new instance of this class more than once. In fact, a lot of the stuff that it uses never or almost never changes, so I should probably make it part of its internal state. I think I'll do it in the next iteration since parameter lists do become a bit long.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/IQuickPulseServiceClient.cs
tokaplan: These are different. 
Each sample represents a second-long data aggregate. The reason why there are more than one is to account for the fact that we might have failed to send a sample from the second before - and we want to backfill. In most cases, there will only be 1 sample sent out each second.
collectionConfigurationErrors represent all problems that were encountered by the SDK while parsing the current configuration. If, for instance, one of the metrics had a faulty filter that couldn't compile, we still want to collect the metric in accordance with other filters, and we still want to collect all other metrics, performance counters, and full telemetry documents, but we need to let the UX know that one of the filters failed. So each time we might be reporting any number of errors related to the currently active configuration.
Which is why there's no correlation between data and errors here.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
tokaplan: Damn I keep doing it!
Done.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
tokaplan: This does do a snapshot. What's happening here is simply a creation of a new instance and storing it for later. The new configuration is not used until the next once-a-second snapshot arrives and Interlock-swaps it in place of the old configuration.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
tokaplan: If it's null, the old configuration must remain. Null in this context means we failed to get a configuration for some reason, even though there was an indication that it changed. Such situation would indicate a weird networking issue when headers have made it through, but the body of the response could not have been read. We will continue supplying the old configuration's etag to the service in that case, and eventually will update the configuration.

Src/PerformanceCollector/Shared/Implementation/IPerformanceCollector.cs
tokaplan: I wish I could. The problem here is that the counters are updated by multiple UX sessions, and we must ensure that other UX sessions (which are most likely viewed by other people) don't experience any sort of interruption when one UX session changes configuration. That's why we can't just recreate everything from scratch here, and must maintain state.
We do do snapshots for the metric configuration since there we can recreate the accumulator exactly the same as before - and it doesn't cost anything. Here, counter creation is very expensive, plus the first read upon creation returns 0.

Src/PerformanceCollector/Filtering/CollectionConfigurationAccumulator.cs
tokaplan: Changed to
"Represents the part of the QuickPulse accumulator which holds operationalized metric data."

Src/PerformanceCollector/Filtering/CollectionConfigurationAccumulator.cs
tokaplan: Great catch, done.

Src/PerformanceCollector/Filtering/DocumentStream.cs
tokaplan: Clarified the concept.

Represents a concept of a uniquely identifiable set of full telemetry documents that are being reported by the SDK. The notion of a stream is needed since multiple UX sessions might be querying for full telemetry documents with different filtering criteria simultaneously.

Src/PerformanceCollector/Filtering/CollectionConfiguration.cs
```diff
@@ -0,0 +1,331 @@
...
+            out CollectionConfigurationError[] errors,
+            Clock timeProvider,
+            IEnumerable<DocumentStream> previousDocumentStreams)
+        {
+            var errorList = new List<CollectionConfigurationError>();
+            var documentStreamIds = new HashSet<string>();
+
+            // quota might be changing concurrently on the collection thread, but we don't need the exact value at any given time
+            // we will try to carry over the last known values to this new configuration
+            Dictionary<string, Tuple<float, float, float, float, float>> previousQuotasByStreamId =
```
SergeyKanzhelev: why quotas are part of configuration?

Src/PerformanceCollector/Filtering/CollectionConfiguration.cs
```diff
@@ -0,0 +1,331 @@
...
+                        CollectionConfigurationErrorType.PerformanceCounterDuplicateIds,
+                        string.Format(CultureInfo.InvariantCulture, "Duplicate performance counter id '{0}'", duplicateMetricId),
+                        null,
+                        Tuple.Create("MetricId", duplicateMetricId)));
+            }
+
+            errors = errorList.ToArray();
+        }
+
+        private void CreateDocumentStreams(
```
SergeyKanzhelev: Method creates what? It returns void and errors

SergeyKanzhelev: What is stream?

Src/PerformanceCollector/Filtering/CollectionConfiguration.cs
```diff
@@ -0,0 +1,331 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Globalization;
+    using System.Linq;
+
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.Implementation.QuickPulse.Helpers;
+
+    /// <summary>
+    /// Represents the collection configuration - a customizable description of performance counters, metrics, and full telemetry documents
```
SergeyKanzhelev: it also have streams and capable of creating streams. I'm confused

Src/PerformanceCollector/Filtering/CollectionConfiguration.cs
```diff
@@ -0,0 +1,331 @@
...
+        /// </summary>
+        public IEnumerable<Tuple<string, AggregationType>> TelemetryMetadata => this.telemetryMetadata;
+
+        /// <summary>
+        /// Metric type only (handled by QuickPulseMetricProcessor).
+        /// </summary>
+        public IEnumerable<Tuple<string, AggregationType>> MetricMetadata => this.metricMetadata;
+
+        /// <summary>
+        /// Document streams (handled by QuickPulseTelemetryProcessor).
```
SergeyKanzhelev: what does "handled by QuickPulseTelemetryProcessor" means?

Src/PerformanceCollector/Filtering/CollectionConfiguration.cs
```diff
@@ -0,0 +1,331 @@
...
+
+            try
+            {
+                metrics.Add(new OperationalizedMetric<TTelemetry>(metricInfo, out errors));
+            }
+            catch (Exception e)
+            {
+                // error creating the metric
+                errors =
+                    errors.Concat(
```
SergeyKanzhelev: Can errors be an `ArrayList` instead? All this complication for nothing

Src/PerformanceCollector/Filtering/CollectionConfiguration.cs
```diff
@@ -0,0 +1,331 @@
...
+                try
+                {
+                    Tuple<float, float, float, float, float> initialQuotas;
+                    previousQuotasByStreamId.TryGetValue(documentStreamInfo.Id, out initialQuotas);
+
+                    var documentStream = new DocumentStream(
+                        documentStreamInfo,
+                        out localErrors,
+                        timeProvider,
+                        initialQuotas?.Item1,
```
SergeyKanzhelev: use named parameters so it's clear what every parameter represents

Src/PerformanceCollector/Filtering/CollectionConfigurationAccumulator.cs
```diff
@@ -0,0 +1,59 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Filtering
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Threading;
+
+    /// <summary>
+    /// Represents the part of the QuickPulse accumulator which holds operationalized metric data.
```
SergeyKanzhelev: please make it more readable. Does this class accumulate configurations?

Src/PerformanceCollector/Filtering/DocumentStream.cs
```diff
@@ -0,0 +1,239 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Globalization;
+
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.Implementation.QuickPulse.Helpers;
+
+    /// <summary>
+    /// Represents a concept of a uniquely identifiable set of full telemetry documents that are being reported by the SDK.
```
SergeyKanzhelev: There should be configuration per stream? Why Configuration has list of streams and not the other way around? Why this class also has lists per type. Are those lists configurations or actual values

Src/PerformanceCollector/Filtering/DocumentStream.cs
```diff
@@ -0,0 +1,239 @@
...
+            var errorList = new List<CollectionConfigurationError>();
+            bool leastOneConjunctionGroupPassed = false;
+
+            if (filterGroups.Count == 0)
+            {
+                // no filters for the telemetry type - filter out, we're not interested
+                return false;
+            }
+
+            // iterate over AND-connected filter groups (groups are connected via OR)
```
SergeyKanzhelev: Can you please translate this comment. Are those `AND` or `OR` connected?

Src/PerformanceCollector/Filtering/DocumentStream.cs
```diff
@@ -0,0 +1,239 @@
...
+                        case TelemetryType.Trace:
+                            this.traceFilterGroups.Add(new FilterConjunctionGroup<TraceTelemetry>(documentFilterConjunctionGroupInfo.Filters, out groupErrors));
+                            break;
+                        default:
+                            throw new ArgumentOutOfRangeException(string.Format(CultureInfo.InvariantCulture, "Unsupported TelemetryType: '{0}'", documentFilterConjunctionGroupInfo.TelemetryType));
+                    }
+
+                    errorList.AddRange(groupErrors);
+                }
+                catch (Exception e)
```
SergeyKanzhelev: what exception do you expect here?

Src/PerformanceCollector/Filtering/DocumentStream.cs
```diff
@@ -0,0 +1,239 @@
...
+                            this.exceptionFilterGroups.Add(new FilterConjunctionGroup<ExceptionTelemetry>(documentFilterConjunctionGroupInfo.Filters, out groupErrors));
+                            break;
+                        case TelemetryType.Event:
+                            this.eventFilterGroups.Add(new FilterConjunctionGroup<EventTelemetry>(documentFilterConjunctionGroupInfo.Filters, out groupErrors));
+                            break;
+                        case TelemetryType.Trace:
+                            this.traceFilterGroups.Add(new FilterConjunctionGroup<TraceTelemetry>(documentFilterConjunctionGroupInfo.Filters, out groupErrors));
+                            break;
+                        default:
+                            throw new ArgumentOutOfRangeException(string.Format(CultureInfo.InvariantCulture, "Unsupported TelemetryType: '{0}'", documentFilterConjunctionGroupInfo.TelemetryType));
```
SergeyKanzhelev: can you please avoid throwing an exception that will be caught immediately. Just add the message to the list of `groupErrors`

Src/PerformanceCollector/Filtering/Filter.cs
```diff
@@ -0,0 +1,596 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Globalization;
+    using System.Linq;
+    using System.Linq.Expressions;
+    using System.Reflection;
+
+    /// <summary>
+    /// Filter determines whether a telemetry document matches the criterion.
```
SergeyKanzhelev: Can you please include the expected configuration language and filtering login into the summary tag.

Src/PerformanceCollector/Filtering/Filter.cs
```diff
@@ -0,0 +1,596 @@
...
+
+        private const string FieldNameCustomMetricsPrefix = "CustomMetrics.";
+
+        private const string FieldNameAsterisk = "*";
+
+        private const string CustomMetricsPropertyName = "Metrics";
+
+        private const string CustomDimensionsPropertyName = "Properties";
+
+        private static readonly MethodInfo DoubleToStringMethodInfo = GetMethodInfo<double, string>(x => x.ToString(CultureInfo.InvariantCulture));
```
SergeyKanzhelev: why do you need reflection? Will use of Func or delegate be sufficient?

Src/PerformanceCollector/Filtering/Filter.cs
```diff
@@ -0,0 +1,596 @@
...
+        private static readonly MethodInfo UriToStringMethodInfo = GetMethodInfo<Uri, string>(x => x.ToString());
+
+        private static readonly MethodInfo StringIndexOfMethodInfo =
+            GetMethodInfo<string, string, int>((x, y) => x.IndexOf(y, StringComparison.OrdinalIgnoreCase));
+
+        private static readonly MethodInfo StringEqualsMethodInfo =
+            GetMethodInfo<string, string, bool>((x, y) => x.Equals(y, StringComparison.OrdinalIgnoreCase));
+
+        private static readonly MethodInfo DoubleParseMethodInfo = typeof(double).GetMethod(
+            "Parse",
```
SergeyKanzhelev: Use `TryParse` to minimize the number of First Chance Exceptions

Src/PerformanceCollector/Shared/Implementation/IPerformanceCollector.cs
SergeyKanzhelev: Can every UX has it's own configuration then? You can create a new configuration and switch to it only after it was fully enabled. Instead of merging all streams and managing all different stream IDs you can just make each stream be responsible for collecting data needed for it.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/IQuickPulseServiceClient.cs
SergeyKanzhelev: my concern is that parameter list is long, but not complete. As you said you also pass role instance for instance. 

I'd say you need multiple instances of this one to support scenario when one process runs multiple "roles". It will be a service fabric case.

Src/PerformanceCollector/Filtering/AccumulatedValue.cs
SergeyKanzhelev: Class is calles `AccumulatedValue`. Rename to `AccumulatedValues` at least. Will aggregation differ depending on the order? Why `Stack`?

Src/PerformanceCollector/Filtering/AccumulatedValue.cs
SergeyKanzhelev: aggregated metrics? Preaggregated? Calculated? Segmented? Filtered? Again, operationalized for me is when you use it for alerting or dashboarding. Make it useful for "operations".

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Service contract/TelemetryDocument/EventTelemetryDocument.cs
SergeyKanzhelev: Please do. Keep types consistent with built-on types.

Src/PerformanceCollector/Filtering/CollectionConfiguration.cs
SergeyKanzhelev: I clearly need more information in summary tag that explain what this class do

Src/PerformanceCollector/Shared/Implementation/QuickPulse/IQuickPulseServiceClient.cs
SergeyKanzhelev: You need to explain to me what your feature do so I can recommend something

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
SergeyKanzhelev: `Clear` implies working with the same object. Just dispose the old object.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
SergeyKanzhelev: How service will reset the configuration and keep default collection only? Will it send "empty" configuration? 

What if service crashed and forgot all etags? 

I think the default behavior should be to stop collecting any custom configuration. Otherwise it SDK may stuck with the bad config.

Src/PerformanceCollector/Filtering/AccumulatedValue.cs
tokaplan: Done. Renamed to Calculated rather than operationalized.

Src/PerformanceCollector/Filtering/AccumulatedValue.cs
tokaplan: Done. Renamed AccumulatedValue to AccumulatedValues.
ConcurrentStack is a highly performant spin-lock collection. The ordering of elements doesn't matter, so we don't actually need the stacking logic. There are a bunch of such collections, each varying in internal implementation quite significantly. This particular implementation is the one I've researched at one point, and it has proven itself well in production code, so I chose it here.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Service contract/TelemetryDocument/EventTelemetryDocument.cs
tokaplan: Will do.

Src/PerformanceCollector/Filtering/CollectionConfiguration.cs
tokaplan: Done.
I've added a remarks section at the top, and split member fields into two regions based on their usage.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/IQuickPulseServiceClient.cs
tokaplan: The feature is to calculate configurable metrics for people to view in QuickPulse :)
Originally this was Filtering because we did (and still do?) plan to reuse the same codebase in "mainland" SDK to be able to collect configurable (calculated) metrics in the regular pipeline, not just QuickPulse. We though of the feature as of "filtering" telemetry items to leave only the ones we're interested in. Then as the idea evolved, we have introduced projections, aggregation types, etc., so right now filtering is only the first stage of "calculating" a metric. At the same time, for full documents (that you see on the QuickPulse blade towards the right side), it's still just filtering and nothing else. So I'm personally fine with leaving it in as "filtering", but you might have other suggestions.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/IQuickPulseServiceClient.cs
tokaplan: What do you mean one process runs multiple "roles"? Are you talking about multiple instances of one given SDK module running side-by-side within a single process? Not sure I'm familiar with the scenario you're mentioning.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
tokaplan: Oh, I think I misunderstood. Are you talking about this.collectionConfigurationErrors? It's just a list of errors that we need to update, it's not even the configuration. I have no reason to create a new instance of a List<> when I can just Clear it.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
tokaplan: - the service will send a default configuration if it wants to "reset" to default.
- if something horrible happens and the UX -> QPS pipeline somehow crashes/resets/reboots - the service will send the default configuration and SDK will get to the default state.
- the service doesn't need to remember any etags, it will send new configuration whenever its etag is different from what SDK provides, so any misfortune with service's internal state will lead to SDK getting a default configuration.
We can't stop collection if the service is unreachable since that might be transient. If the service is unreachable for a while - we stop collecting anything anyway, and the configuration is not important.

Src/PerformanceCollector/Shared/Implementation/IPerformanceCollector.cs
tokaplan: Especially since we have favorites, chances are that if you have 10 UXs open - most of the metrics and document streams they query for will actually be exactly the same. The reason SDK needs to merge configuration from all clients is because we can't afford to apply the same filter 10 times instead of 1, send the same document 10 times instead of 1, and (most importantly for this particular case) collect the same counter 10 times instead of 1.
So we merge :)

Src/PerformanceCollector/Filtering/CollectionConfiguration.cs
tokaplan: "Also"? What else does that? I've revisited the comment.

Src/PerformanceCollector/Filtering/CollectionConfiguration.cs
tokaplan: It means that this collection is used from QuickPulseTelemetryProcessor at collection-time. I've revised the comment.

Src/PerformanceCollector/Filtering/CollectionConfiguration.cs
tokaplan: Yeah, it is ugly... could be List<>... This is not a performance-critical path, so probably no big deal,  but let me look at it. Thanks for pointing it out, this is a left-over from the early days when slopes were still snowy...

Src/PerformanceCollector/Filtering/CollectionConfiguration.cs
tokaplan: It's Create*, not Get* :) It changes the internal state, doesn't necessary have to return :) I could have returned errors, but I don't really like it, a painful reminder of the HRESULT days :)
Would it make you feel better if I changed it to Initialize*?

tokaplan: See description for the DocumentStream class. I think whoever is looking at this private method implementation will get to that point, no need to copy that comment here.

Src/PerformanceCollector/Filtering/CollectionConfigurationAccumulator.cs
tokaplan: Revised the comment making it clear that it is the actual collection-time data that is being accumulated.

Src/PerformanceCollector/Filtering/DocumentStream.cs
tokaplan: - imagine there are 10 UXs out there. Each UX can configure a document stream, which is essentially a set of criteria to which  telemetry items must comply in order to be included into the stream. Each UX can configure more than one document stream if it wishes to. QPS will merge those streams into a single configuration and pass it on to the SDK. SDK doesn't have to know anything about individual UXs, it only has to know about individual streams and their criteria. Think of a document stream as of a set of filters, and it will become clear why there are many of them for a single configuration (of which there's only one on the SDK side).
- the reason why there are per-type lists is because of performance concerns (same reason why there are per-type lists of CalculatedMetrics in CollectionConfiguration). We don't want to cast types, we only do that once and then do if's and compile-time generic methods to execute code for particular ITelemetry type. Since we don't expect many more ITelemetry implementations but do expect huge numbers of items every second, this appears to make sense.
- those are neither configurations nor values. Those are objects created based on configurations that are capable of determining if an ITelemetry passes the test specified by this DocumentStream's criteria. The actual values are only stored in accumulators. Same as for CalculatedMetric instances in CollectionConfiguration.

Src/PerformanceCollector/Filtering/DocumentStream.cs
tokaplan: :)
Not sure if I made it clearer, but I tried :)

Src/PerformanceCollector/Filtering/DocumentStream.cs
tokaplan: The only way this exception actually happens is if we ship something completely broken. I like the well-established pattern of throwing in default clause, and if this actually fires we have much bigger problems than an extra exception throw :) I don't think it makes sense to change.

Src/PerformanceCollector/Shared/QuickPulseMetricProcessor.cs
```diff
@@ -0,0 +1,88 @@
...
+
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Filtering;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+    using Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.Implementation.QuickPulse;
+
+    /// <summary>
+    /// Metric processor for collecting QuickPulse data.
+    /// </summary>
+    internal sealed class QuickPulseMetricProcessor : IMetricProcessor
```
SergeyKanzhelev: >QuickPulseMetricProcessor [](start = 26, length = 25)

internal class should go into `Implementation` folder

Src/PerformanceCollector/Filtering/DocumentStream.cs
tokaplan: Either something broke completely and we get an exception from the switch's default clause, or there's something wrong with configuration and we get an exception from FilterConjunctionGroup's constructor which parses user-supplied filter configuration. I don't care which exception it is, my reaction is to report it up the call chain, hence catch(Exception).

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
```diff
@@ -120,6 +124,12 @@ public QuickPulseTelemetryModule()
         public bool DisableTopCpuProcesses { get; set; }
 
         /// <summary>
+        /// Gets the auth API key.
```
SergeyKanzhelev: >Gets the auth API key. [](start = 12, length = 22)

Can you please elaborate on how it is used and where to obtain it as this is a public contract for our users

Src/PerformanceCollector/Filtering/Filter.cs
tokaplan: There is no language per se, the configuration is specified by FilterInfo DTO. I've updated the comment to reflect that.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Service contract/TelemetryDocument/EventTelemetryDocument.cs
```diff
@@ -0,0 +1,44 @@
...
+                return TelemetryDocumentType.Event.ToString();
+            }
+
+            private set
+            {
+            }
+        }
+
+        [DataMember(EmitDefaultValue = false)]
+        public string[] DocumentStreamIds { get; set; }
```
SergeyKanzhelev: >DocumentStreamIds [](start = 24, length = 17)

The same comment as for `Id`. This should be sent separately

SergeyKanzhelev:
OK. This PR is huge and I recommend a lot of refactoring and architecture improvements to be made. In order to lock on it tomorrow I assume that the SDK to Server API is not public and not intended to become public. I'd like you to address the following to put it inside the SDK:

1. The only new public setting is `AuthApiKey`. Please rename to `AuthenticationApiKey` to follow C# guidelines. Doc should be updated to be user friendly
2. No new threads started (I think I validated it)
3. "Average" configuration is added to the performance tests (not only enable = true, but also couple of filters)
4. No exceptions on success path and all threads are marked Internal for exceptions statistics feature https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/356
5. You confirm that you tested internal classes on edge scenarios and onboarded quick pulse for extra validation

I think those 5 will be enough for me for signoff, All other comments are mostly recommendations (again, assuming that this is not a public interface and will never be public in the current shape without review)

SergeyKanzhelev:
I'll try to look at your replies later

SergeyKanzhelev:
And @Dmitry-Matveev said he is looking at it as well =)

Src/PerformanceCollector/Filtering/Filter.cs
tokaplan: Great catch, done. Thanks!

Src/PerformanceCollector/Filtering/CollectionConfiguration.cs
tokaplan: There's a global quota (which is not part of the configuration), and also individual per-telemetry type quotas for each document stream. They have to be independent for each stream to ensure that a chatty stream doesn't use up the entire quota and other streams (and users) get nothing.

Src/PerformanceCollector/Filtering/CollectionConfiguration.cs
tokaplan: Done.

Src/PerformanceCollector/Filtering/Filter.cs
tokaplan: This is not reflection. This MethodInfo is used to construct a MethodCallExpression which is then compiled for performance and executed at collection-time for each ITelemetry.

Src/PerformanceCollector/Shared/QuickPulseMetricProcessor.cs
tokaplan: Done.

Src/PerformanceCollector/Shared/QuickPulseTelemetryModule.cs
tokaplan: I don't actually know how to obtain it exactly just yet :) I think we'll post a blog post once everything is finalized. For now, I've added a comment stating that the key is created in the portal and serves the purpose of secure configuration distribution.

Src/PerformanceCollector/Shared/Implementation/QuickPulse/Service contract/TelemetryDocument/EventTelemetryDocument.cs
tokaplan: This is different. _DocumentStreamId_ doesn't have anything to do with the telemetry item in broader SDK sense, but rather with the QuickPulse's concept of the document stream itself. This value is used to identify which document streams this telemetry item belongs to. We can't afford to send out the entire document 10 times for 10 different UXs, so we must maintain a list of 10 stream ids here in that case.

Src/PerformanceCollector/Filtering/Filter.cs
```diff
@@ -0,0 +1,617 @@
...
+                    {
+                        // in order for the expression to compile, we must cast to double unless it's already double
+                        // we're using double as the common lowest denominator for all numerical types
+                        Expression fieldConvertedExpression = fieldTypeCode == TypeCode.Double ? fieldExpression : Expression.ConvertChecked(fieldExpression, typeof(double));
+
+                        switch (this.predicate)
+                        {
+                            case Predicate.Equal:
+                                this.ThrowOnInvalidFilter(fieldType, !this.comparandDouble.HasValue);
+                                return Expression.Equal(fieldConvertedExpression, Expression.Constant(this.comparandDouble.Value));
```
Dmitry-Matveev: Equality in double is somewhat a gamble especially if one cannot guarantee that exactly the same steps were applied to operands on both sides of comparison while they were cast/converted into double. Can it happen here with filters that were set in UI?

Src/PerformanceCollector/Shared/Implementation/QuickPulse/QuickPulseCollectionStateManager.cs
```diff
@@ -125,8 +146,29 @@ public TimeSpan UpdateState(string instrumentationKey)
...
                     return this.DetermineBackOffs();
                 }
+                else
+                {
+                    // we have samples
+                    if (dataSamplesToSubmit.Any(sample => sample.CollectionConfigurationAccumulator.GetRef() != 0))
+                    {
+                        ////!!! better solution?
+                        // some samples are still being processed, wait a little to give them a chance to finish
+                        Thread.Sleep(this.coolDownTimeout);
```
Dmitry-Matveev: Should we sleep and block the thread or have a Task.Delay or some other waiting solution instead?

SergeyKanzhelev:
Please squash commits when merging: 

image

PR #403, opened 2017-04-04
set Success for web requests

SergeyKanzhelev:


Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -107,10 +107,21 @@ public void OnEndRequest(HttpContext context)
...
 
-            // Success will be set in Sanitize on the base of ResponseCode 
+            var success = true;
             if (string.IsNullOrEmpty(requestTelemetry.ResponseCode))
             {
-                requestTelemetry.ResponseCode = context.Response.StatusCode.ToString(CultureInfo.InvariantCulture);
+                var statusCode = context.Response.StatusCode;
+                requestTelemetry.ResponseCode = statusCode.ToString(CultureInfo.InvariantCulture);
+
+                if (statusCode >= 400 && statusCode != 401)
```
Dmitry-Matveev: Do we need the similar change in ASP.NET core SDK now if we are about to remove it from the common place in Sanitize()?

PR #409, opened 2017-04-05
Keithh 2.4.0 beta1

kstevenham:
I updated the nuget package to Microsoft.ApplicationInsights.2.4.0-beta1-build07987 which touched LOTS of files. Otherwise, this check-in enables exception statistics and emits 2 example exceptions / minute for each exception for a given ProblemId.

### DIFF
--- a/Src/WindowsServer/WindowsServer.Shared.Tests/FirstChanceExceptionStatisticsTelemetryModuleTest.cs
+++ b/Src/WindowsServer/WindowsServer.Shared.Tests/FirstChanceExceptionStatisticsTelemetryModuleTest.cs
@@ -456,9 +456,14 @@
             Assert.Equal(1, metrics[0].Value, 15);
             Assert.Equal(0, metrics[1].Value, 15);
 
-            Assert.Equal(1, this.items.Count);
-            Assert.Equal(2, ((MetricTelemetry)this.items[0]).Count);
-            Assert.Equal(1, ((MetricTelemetry)this.items[0]).Sum);
+            Assert.Equal(2, this.items.Count);
+
+            Assert.Equal(1, ((MetricTelemetry)this.items[0]).Count);
+            Assert.Equal(1, ((MetricTelemetry)this.items[1]).Count);
+
+            // One of them should be 0 as re-thorwn, another - one
+            Assert.Equal(0, Math.Min(((MetricTelemetry)this.items[0]).Sum, ((MetricTelemetry)this.items[1]).Sum), 15);
+            Assert.Equal(1, Math.Max(((MetricTelemetry)this.items[0]).Sum, ((MetricTelemetry)this.items[1]).Sum), 15);
         }
 
         [TestMethod]
--- a/Src/WindowsServer/WindowsServer.Shared.Tests/FirstChanceExceptionStatisticsTelemetryModuleTest.cs
+++ b/Src/WindowsServer/WindowsServer.Shared.Tests/FirstChanceExceptionStatisticsTelemetryModuleTest.cs
@@ -480,23 +480,24 @@
             }
         }
 
-        [TestMethod]
-        public void FirstChanceExceptionStatisticsTelemetryModuleWasTrackedReturnsTrueForInnerException()
-        {
-            using (var module = new FirstChanceExceptionStatisticsTelemetryModule())
-            {
-                module.Initialize(this.configuration);
+        //// This test is temporarily removed until WasExceptionTracked method is enhanced.
+        ////[TestMethod]
+        ////public void FirstChanceExceptionStatisticsTelemetryModuleWasTrackedReturnsTrueForInnerException()
+        ////{
+        ////    using (var module = new FirstChanceExceptionStatisticsTelemetryModule())
+        ////    {
+        ////        module.Initialize(this.configuration);
 
-                var exception = new Exception();
+        ////        var exception = new Exception();
 
-                Assert.False(module.WasExceptionTracked(exception));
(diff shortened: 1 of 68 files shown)


Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -221,48 +291,210 @@ private void CalculateStatistics(object sender, FirstChanceExceptionEventArgs fi
...
+                this.exceptionKeyValues.RwLock.EnterWriteLock();
+
+                if (this.exceptionKeyValues.ValueCache.Contains(exceptionKey) == false)
+                {
+                    this.exceptionKeyValues.ValueCache.Add(exceptionKey);
+
+                    if (exceptionTelemetry == null)
+                    {
+                        exceptionTelemetry = new ExceptionTelemetry(exception);
+                        this.telemetryClient.Initialize(exceptionTelemetry);
```
SergeyKanzhelev: You can do all this logic outside of WriteLock.

Src/WindowsServer/WindowsServer.Shared.Tests/FirstChanceExceptionStatisticsTelemetryModuleTest.cs
```diff
@@ -434,6 +547,37 @@ public void FirstChanceExceptionStatisticsTelemetryModuleWasTrackedReturnsFalseF
...
+                    {
+                        throw new Exception("Second exception");
+                    }
+                    catch
+                    {
+                    }
+                }
+
+                // There should have been two unique TrackExceptions
+                Assert.Equal(2, this.items.Count);
```
kstevenham: I don't see any Metric items, should I?

Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -104,14 +127,15 @@ internal bool WasExceptionTracked(Exception exception)
...
+            ////if (!wasTracked)
+            ////{
+            ////    var innerException = exception.InnerException;
+            ////    if (innerException != null)
+            ////    {
+            ////        wasTracked = IsTracked(innerException);
+            ////    }
+            ////}
 
             if (!wasTracked && exception is AggregateException)
```
SergeyKanzhelev: comment this one as well then

Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -221,49 +299,210 @@ private void CalculateStatistics(object sender, FirstChanceExceptionEventArgs fi
...
+            if (this.exceptionKeyValues.ValueCache.Contains(exceptionKey) == false)
+            {
+                this.exceptionKeyValues.ValueCache.Add(exceptionKey);
+
+                this.exceptionKeyValues.RwLock.ExitWriteLock();
+
+                if (exceptionTelemetry == null)
+                {
+                    exceptionTelemetry = new ExceptionTelemetry(exception);
+                    this.telemetryClient.Initialize(exceptionTelemetry);
```
SergeyKanzhelev: you decided to keep it inside the write lock?

kstevenham: It's not

Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -168,40 +199,79 @@ private void CalculateStatistics(object sender, FirstChanceExceptionEventArgs fi
...
                 if (exception == null)
                 {
                     WindowsServerEventSource.Log.FirstChanceExceptionCallbackExeptionIsNull();
                     return;
                 }
 
-                WindowsServerEventSource.Log.FirstChanceExceptionCallbackCalled();
+                if (this.WasExceptionTracked(exception) == true)
+                {
+                    return;
```
kstevenham: This should now be resolved

Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -221,49 +299,210 @@ private void CalculateStatistics(object sender, FirstChanceExceptionEventArgs fi
...
+
+                this.exceptionKeyValues.RwLock.ExitWriteLock();
+
+                if (exceptionTelemetry == null)
+                {
+                    exceptionTelemetry = new ExceptionTelemetry(exception);
+                    this.telemetryClient.Initialize(exceptionTelemetry);
+                }
+
+                StackTrace st = new StackTrace(3, true);
```
Dmitry-Matveev: Is there a way to ensure that no in-lining or other optimization happens and skipping 3 frames always leads to the consistent result? The previous methods do not seem to satisfy any of the in-lining/optimization requirements, so this should work as is unless those requirements change in CLR.

lmolkova:
It seems that after this change TestAsmx_CheckExceptionAndRequestCollectedForResourceNotFound Functional test does not work, because it expect just one exception to be tracked and now there are multiple.
@kstevenham Could you please check it?

PR #413, opened 2017-04-06
Use Microsoft.AspNet.Diagnostics.DiagnosticsHttpModule on NET4.5 to handle incoming requests

lmolkova:
Microsoft.AspNet.Diagnostics.DiagnosticsHttpModule tracks incoming requests, creates Activities and notifies AppInsights with DiagnosticsSource
Module was implemented by ASP.NET team: [code](https://github.com/aspnet/AspNetCorrelationIdTracker)


NuGet.config
```diff
@@ -19,5 +19,6 @@
     <add key="nuget.org" value="https://www.nuget.org/api/v2/" />
     <add key="applicationinsights" value="https://www.myget.org/F/applicationinsights/api/v3/index.json" />
     <add key="dotnet" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" protocolVersion="3" />
+    <add key="aspnetmodule" value="\\ddfiles\Team\Public\Correlation\AI"/>
```
SergeyKanzhelev: this may not work on build servers. Can you publish to MyGet?

Test/Web/FunctionalTests/TestApps/WebAppFW45/Web.config
```diff
@@ -32,6 +32,8 @@
       <add name="ExtensionlessUrlHandler-Integrated-4.0" path="*." verb="*" type="System.Web.Handlers.TransferRequestHandler" preCondition="integratedMode,runtimeVersionv4.0" />
     </handlers>
     <modules runAllManagedModulesForAllRequests="true">
+      <add name="DiagnosticsHttpModule" type="Microsoft.AspNet.Diagnostics.DiagnosticsHttpModule, Microsoft.AspNet.Diagnostics, Version=1.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" preCondition="integratedMode,managedHandler" />
+      <add name="DiagnosticsHttpModule2" type="Microsoft.AspNet.Diagnostics.DiagnosticsHttpModule, Microsoft.AspNet.Diagnostics, Version=1.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" preCondition="integratedMode,managedHandler" />
```
SergeyKanzhelev: >DiagnosticsHttpModule2 [](start = 17, length = 22)

why do we need a second one? Can you please add a comment if it's intentional

Test/Web/FunctionalTests/TestApps/WebAppFW45/packages.config
```diff
@@ -1,11 +1,13 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="Microsoft.ApplicationInsights" version="2.4.0-beta1-build07987" targetFramework="net451" />
   <package id="Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel" version="2.4.0-beta1-build07987" targetFramework="net451" />
+  <package id="Microsoft.AspNet.Diagnostics" version="1.0.0-beta-10405-01" targetFramework="net451" />
```
SergeyKanzhelev: >"1.0.0-beta-10405-01 [](start = 53, length = 20)

why the version of this module is not aligned with the rest of ASP.NET? Should it be 4.7 something?

Test/Web/FunctionalTests/TestApps/WebAppFW45/ApplicationInsights.config
```diff
@@ -43,6 +43,7 @@
       </Handlers>
     </Add>
     <Add Type="Microsoft.ApplicationInsights.Web.ExceptionTrackingTelemetryModule, Microsoft.AI.Web"/>
+    <Add Type="Microsoft.ApplicationInsights.Web.AspNetDiagnosticTelemetryModule, Microsoft.AI.Web"/>
```
SergeyKanzhelev: >AspNetDiagnosticTelemetryModule [](start = 49, length = 31)

so we need both - `RequestTrackingTelemetryModule` and `AspNetDiagnosticTelemetryModule`?

Test/Web/FunctionalTests/FunctionalTests/Helpers/EtwEventSession.cs
```diff
@@ -94,7 +95,11 @@ private bool IsInterestingTrace(TraceEvent data)
 
             bool result = true;
 
-            if (data.PayloadNames.Length > 0)
+            if (data.ProviderName == "Microsoft-AspNet-Diagnostics")
+            {
+                result = true;
```
SergeyKanzhelev: >result = true; [](start = 16, length = 14)

we check later on that the trace is coming from the process we monitor. Simple `return true` can mess up traces from the parrallel tests executions.

Test/Web/FunctionalTests/FunctionalTests/Helpers/EtwEventSession.cs
```diff
@@ -14,6 +14,7 @@ public class EtwEventSession : IDisposable
             "Microsoft-ApplicationInsights-Core",
             "Microsoft-ApplicationInsights-WindowsServer-TelemetryChannel",
             "Microsoft-ApplicationInsights-Extensibility-WindowsServer",
+            "Microsoft-AspNet-Diagnostics",
```
SergeyKanzhelev: >"Microsoft-AspNet-Diagnostics", [](start = 12, length = 31)

what output this event source will generate? Will it report errors?

Src/Web/Web.Shared.Net.Tests/Helpers/HttpModuleHelper.cs
```diff
@@ -67,7 +67,9 @@ public static HttpContext GetFakeHttpContext(IDictionary<string, string> headers
 
             var workerRequest = new SimpleWorkerRequestWithHeaders(UrlPath, UrlQueryString, new StringWriter(CultureInfo.InvariantCulture), headers);
             
-            return new HttpContext(workerRequest);
+            var context = new HttpContext(workerRequest);
+            HttpContext.Current = context;
```
SergeyKanzhelev: >HttpContext.Current = context; [](start = 12, length = 30)

wouldn't this mess up with all tests using the Fake Http Context?

SergeyKanzhelev: Where is this cleaned up?

---
In reply to: [110214089](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/413#discussion_r110214089) [](ancestors = 110214089)

Src/Web/Web.Shared.Net/Implementation/ActivityHelpers.cs
```diff
@@ -1,114 +1,24 @@
...
 #if NET40
+    using Microsoft.ApplicationInsights.DataContracts;
     using Microsoft.ApplicationInsights.Net40;
 #endif
     using Microsoft.ApplicationInsights.Web.Implementation;
 
     internal class ActivityHelpers
     {
-        internal const string RequestActivityItemName = "Microsoft.AppInsights.Web.Request";
         internal const string CorrelationContextItemName = "Microsoft.AppInsights.Web.CorrelationContext";
```
SergeyKanzhelev: >AppInsights [](start = 70, length = 11)

please do not use `AppInsights`. Use `ApplicationInsights`

Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
```diff
@@ -153,6 +156,7 @@ private void AddCorreleationHeaderOnSendRequestHeaders(HttpApplication httpAppli
             }
         }
 
+#if NET40
```
SergeyKanzhelev: why wouldn't we do the same for NET45? We can restore the context just in case, can we? Or ASP.NET module will do it guaranteed?

SergeyKanzhelev:
>                 // TODO: write event log, when https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/390/commits/c38d6d48e666114fedba90b8df6a4f36571cbb66 is merged to 2.4.0

this is merged

---
Refers to: Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs:102 in 180e036. [](commit_id = 180e03689bf31538ed1bb28c5afe6fabb8c4ec55, deletion_comment = False)

SergeyKanzhelev:
This PR needs CHANGELOG.md update. Please include the link to github for the ASP.NET diagnostics module

Test/Web/FunctionalTests/TestApps/WebAppFW45/Web.config
lmolkova: This is a very good questions.
The next two lines in this app are
```
       <add name="ApplicationInsightsHttpModule" type="Microsoft.ApplicationInsights.Web.ApplicationInsightsHttpModule, Microsoft.AI.Web" preCondition="managedHandler" />
        <add name="ApplicationInsightsHttpModule2" type="Microsoft.ApplicationInsights.Web.ApplicationInsightsHttpModule, Microsoft.AI.Web" preCondition="managedHandler" />		        <add name="ApplicationInsightsHttpModule2" type="Microsoft.ApplicationInsights.Web.ApplicationInsightsHttpModule, Microsoft.AI.Web" preCondition="managedHandler" />
```

So AI module is added two times. I assume this test checks besides other things, that even if there are 2 modules, 1 request telemetry will be reported anyway. It's questionable whether AI needs to handle this properly, but anyway. 

But I'm fine with leaving just one DiagnosticHttpModule (BTW it will be renamed to TelemetryCorrelationModule)

Test/Web/FunctionalTests/TestApps/WebAppFW45/packages.config
lmolkova: Why do you think it should be aligned? It's an HTTP module, it could work with any ASP.NET version

Test/Web/FunctionalTests/TestApps/WebAppFW45/ApplicationInsights.config
lmolkova: Yes.

Test/Web/FunctionalTests/FunctionalTests/Helpers/EtwEventSession.cs
lmolkova: I added it for the debugging, will remove

Test/Web/FunctionalTests/FunctionalTests/Helpers/EtwEventSession.cs
lmolkova: it could log some issues for invalid headers

Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
lmolkova: ASp.NET HttpModule restores context in PreHandlerRequestExecute if it was lost

Src/Web/Web.Shared.Net.Tests/Helpers/HttpModuleHelper.cs
lmolkova: Every MSTest test is executed in it's own AppDomain, it does not affect other tests

lmolkova:
@SergeyKanzhelev 

> // TODO: write event log, when https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/390/commits/c38d6d48e666114fedba90b8df6a4f36571cbb66 is merged to 2.4.0

merged to develop, [not 2.4.0-beta1](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/2.4.0-beta1/Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs#L73)

NuGet.config
lmolkova: I'm waiting for the package from the AspNet team. Can we temporary put it to AI myget once i have it? It depends on DiagSource too, so they can't publish it to nuget until DiagSource is published (i'm working on it)

Test/Web/FunctionalTests/FunctionalTests/Helpers/EtwEventSession.cs
lmolkova: I've added it for the debuggin, will remove

lmolkova:
@SergeyKanzhelev 
ready for review

Src/Web/Web.Net45/AspNetDiagnosticTelemetryModule.cs
```diff
@@ -0,0 +1,211 @@
...
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+    using Microsoft.ApplicationInsights.Web.Implementation;
+
+    /// <summary>
+    /// Listens to ASP.NET DiagnosticSource and enables instrumentation with Activity: let ASP.NET create root Activity for the request.
+    /// </summary>
+    public class AspNetDiagnosticTelemetryModule : IObserver<DiagnosticListener>, IDisposable, ITelemetryModule
+    {
+        private const string AspNetListenerName = "Microsoft.AspNet.TelemetryCorrelation";
```
SergeyKanzhelev: >Microsoft.AspNet.TelemetryCorrelation [](start = 51, length = 37)

will it report begin and end callback in future as well? Maybe RequestsTracking? RequestsTelemetryTracking? (this name is better than Diagnostics though)

Src/Web/Web.Nuget/Package.nuspec
```diff
@@ -24,14 +24,15 @@
       <group targetFramework="net45">
         <dependency id="Microsoft.ApplicationInsights" version="[$coresdkversion$]" />
         <dependency id="Microsoft.ApplicationInsights.WindowsServer" version="$version$" />
+        <dependency id="Microsoft.AspNet.TelemetryCorrelation" version="1.0.0-preview1-10407-02" />
```
SergeyKanzhelev: Do we need a Diagnostics Source here as well? Since we are using it explicitly - it will be nice to have it as a dependency to point to the specific version of it and allow nuget to do necessary versions resolutions and bindingredirects

Src/Web/Web.Net45/AspNetDiagnosticTelemetryModule.cs
```diff
@@ -0,0 +1,211 @@
...
+            if (dispose)
+            {
+                this.aspNetSubscription?.Dispose();
+                this.allListenerSubscription?.Dispose();
+            }
+        }
+
+        private class AspNetEventObserver : IObserver<KeyValuePair<string, object>>
+        {
+            private const string FirstRequestFlag = "Microsoft.ApplicationInsights.FirstRequestFlag";
```
SergeyKanzhelev: >private const string FirstRequestFlag = "Microsoft.ApplicationInsights.FirstRequestFlag" [](start = 12, length = 88)

can/should we use RequestTelemetry object instead? Also why would AspNetTelemetry module call this twice for the same context? Should this check go there?

Src/Web/Web.Net45/AspNetDiagnosticTelemetryModule.cs
lmolkova: Yes, the only point of the module to parse headers and report request start/stop.
we had a long discussion with Asp.Net team on naming and RequestTracking was one of the suggestions. There is already something in ASP.NET that called request tracing or tracking and this name must not be used.
TelemetryCorrelation was chosen because it explains high-level goal of the whole feature it serves for.

SergeyKanzhelev:
@lmolkova please address the review comments in a separate PR. I'm merging to try it out myself

Src/Web/Web.Nuget/Package.nuspec
lmolkova: agree, I'll add diag source explicitely

Src/Web/Web.Net45/AspNetDiagnosticTelemetryModule.cs
lmolkova: 1. RequestTelemetry is set in the BeginRequest and always in the context at the time of EndRequest. So it cannot be used as an indicator of the first request.
2. Why it could happen: 
AppInsights.Web protects itself from the case when module is added twice into the web.config.
It is questionable whether there should be protection from it, but this is already there and there are even functional tests for it. I don't want to break it.
So the AspNetEventObserver just repease the logic of ApplicationInsightsHttpModule.

TelemetryCorrelationModule, being generic module anyone could consume, does not try to protect listener from it.

Issue #400, opened 2017-04-03
Unresolved assembly reference when onboarding a net 4.0 application onto 2.3.0-beta3 SDK


tokaplan:
I’ve tried to onboard a net 4.0 application onto a previously released Microsoft.ApplicationInsights.PerfCounterCollector 2.3.0-beta3 nuget from nuget.org, and I get these errors that prevent me from running the app.

![image](https://cloud.githubusercontent.com/assets/9950081/24634550/4cc33004-1883-11e7-853f-fcd032702e34.png)

  Installing BCL nugets onto the application resolves this. Is this expected? Will customers know to do that? What’s the reason behind not introducing an actual dependency on BCL nuget if it needs to be installed?

  The issue also repros on 2.4.

tokaplan:
https://ibb.co/eTEwDv

Dmitry-Matveev:
Let's investigate in 2.4-Beta1 milestone to fix in 2.4. If ApplicationInsights.Web nuget package is installed in 4.0 application - it brings both: Bcl and PercCounterCollector dependencies. If PerfCounterCollector is installed separately on 4.0 application, then it only brings EventSource and ApplicationInsights dependencies but no Bcl.

To investigate:
- Why does PerfCounterCollector need Bcl on 4.0;
- Why Bcl is not in the list of PerfCounterCollector nuget dependencies for 4.0? (Could've been intentional if someone thought PerfCounterCollector always goes in bundle with AI Web nuget);

PR #415, opened 2017-04-06
remove bcl references from projects

SergeyKanzhelev:
#400


SergeyKanzhelev:
I'll send another PR with the merge from develop that will include this fix

PR #416, opened 2017-04-06
Move Sergey's fix to remove BCL to develop

iusafaro:


Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -186,7 +188,42 @@ private void GenerateCorrelationIdAndAddToDictionary(string ikey, string appId)
...
+
+                    using (HttpWebResponse response = (HttpWebResponse)request.GetResponse())
+                    {
+                        using (StreamReader reader = new StreamReader(response.GetResponseStream()))
+                        {
+                            return reader.ReadToEnd();
+                        }
+                    }
+                }
+                finally
```
SergeyKanzhelev: I just realized - without the ContinueWith we need to catch exception here

PR #420, opened 2017-04-07
Merge latest develop changes into 2.4.0-beta1

SergeyKanzhelev:


Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -119,18 +133,45 @@ public bool TryGetXComponentCorrelationId(string instrumentationKey, out string
...
+                                }
+                                else
+                                {
+                                    this.RegisterFailure(instrumentationKey, ex);
+                                }
+                            }
                             catch (Exception ex)
                             {
-                                CrossComponentCorrelationEventSource.Log.FetchAppIdFailed(this.GetExceptionDetailString(ex));
+                                this.RegisterFailure(instrumentationKey, ex);
```
iusafaro: double checking - is it intentional change and the latest/greatest is in 2.3.0?

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -183,6 +269,28 @@ private Uri GetAppIdEndPointUri(string instrumentationKey)
             return new Uri(this.endpointAddress, string.Format(CultureInfo.InvariantCulture, AppIdQueryApiRelativeUriFormat, instrumentationKey));
         }
 
+        /// <summary>
+        /// Registers failure for further action in future.
+        /// </summary>
+        /// <param name="instrumentationKey">Instrumentation key for which the failure occurred.</param>
+        /// <param name="ex">Exception indicating failure.</param>
+        private void RegisterFailure(string instrumentationKey, Exception ex)
```
iusafaro: we can move logic for AggregateException to this method and it'll make code simpler

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -172,7 +218,47 @@ private void GenerateCorrelationIdAndAddToDictionary(string ikey, string appId)
...
+                    {
+                        using (StreamReader reader = new StreamReader(response.GetResponseStream()))
+                        {
+                            return reader.ReadToEnd();
+                        }
+                    }
+                }
+                catch (Exception ex)
+                {
+                    CrossComponentCorrelationEventSource.Log.FetchAppIdFailed(this.GetExceptionDetailString(ex));
```
iusafaro: should be this.RegisterFailure(instrumentationKey, ex);

Src/Common/CorrelationIdLookupHelper.cs
SergeyKanzhelev: @nizarq can you comment? I think it is https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/Common/CorrelationIdLookupHelper.cs#L174

Src/Common/CorrelationIdLookupHelper.cs
SergeyKanzhelev: @nizarq?

Src/Common/CorrelationIdLookupHelper.cs
nizarq: This looks like faulty merge. Note that I changed appId.Result on line 120/134 to appId.GetAwaiter().GetResult(). That change makes the AggregateException block obsolete. I would suggest take what's in develop.

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -119,18 +133,45 @@ public bool TryGetXComponentCorrelationId(string instrumentationKey, out string
...
                             {
-                                CrossComponentCorrelationEventSource.Log.FetchAppIdFailed(this.GetExceptionDetailString(ex));
+                                this.RegisterFailure(instrumentationKey, ex);
                             }
                         });
 
                         return false;
                     }
                 }
+                catch (AggregateException ex)
```
nizarq: This catch block shouldn't be there - take what's in develop.

Src/Common/CorrelationIdLookupHelper.cs
nizarq: It seems like Commit d03ab986 made by @iusafaro  with the following comments have corrupted this in develop. The version of the file before that change in develop is the correct version. Please address that merge.

```
Move Sergey's fix to remove BCL to develop (#416)

* remove bcl references from projects

* Switch versions back to 2.3.0

* Switch version back to 2.3.0

* Add catch exception

* Catch AggregateException

```

SergeyKanzhelev: @nizarq @iusafaro please discuss the best solution. We removed Bcl reference so there are no `GetAwaiter` to call

nizarq: In that case just move the logic to unwrap aggregate exception in RegisterFailure call. Almost everything stays the same as it was before develop except for the following: 
1. GetAwaiter is gone - use .Result directly. 
2. Register Failure can check if the exception is Aggreagate and unwrap it before logging it.

SergeyKanzhelev: would you pls comment here so your comment will not be lost: https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/421 thanks!

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -172,7 +218,47 @@ private void GenerateCorrelationIdAndAddToDictionary(string ikey, string appId)
...
         }
+#else
+        /// <summary>
+        /// Retrieves the app id given the instrumentation key.
+        /// </summary>
+        /// <param name="instrumentationKey">Instrumentation key for which app id is to be retrieved.</param>
+        /// <returns>App id.</returns>
+        private Task<string> FetchAppIdFromService(string instrumentationKey)
+        {
+            var task = new Task<string>(() =>
```
nizarq: You never start this task. This will hang forever. Please fix this on develop. User TaskFactory.StartNew to create your task.

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -119,18 +133,45 @@ public bool TryGetXComponentCorrelationId(string instrumentationKey, out string
                             {
                                 this.GenerateCorrelationIdAndAddToDictionary(instrumentationKey, appId.Result);
                             }
+                            catch (AggregateException ex)
+                            {
+                                ex.Flatten();
```
nizarq: ex.Flatten() wouldn't flatten the exception in place. It will return a new flattened result.

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -119,18 +133,45 @@ public bool TryGetXComponentCorrelationId(string instrumentationKey, out string
...
+                                if (ex.InnerExceptions != null && ex.InnerExceptions.Count > 0 && ex.InnerExceptions[0] != null)
+                                {
```
nizarq: Can just be ex.InnerException != null which returns the first exception or null

Src/Common/CorrelationIdLookupHelper.cs
pharring: This should have a unit test. It probably doesn't do what the author intended.

PR #422, opened 2017-04-07
Azure Storage HTTP parsing improvements

TomMilos:
* support national clouds
* simplify/reuse parsing code
* improve Blob parsing
  * support container hierarchies
  * put container and blob names as properties

### DIFF
--- a/Src/DependencyCollector/Shared/Implementation/HttpParsers/AzureQueueHttpParser.cs
+++ b/Src/DependencyCollector/Shared/Implementation/HttpParsers/AzureQueueHttpParser.cs
@@ -1,6 +1,7 @@
 ﻿namespace Microsoft.ApplicationInsights.DependencyCollector.Implementation.HttpParsers
 {
-    using System;
+    using System.Collections.Generic;
+
     using DataContracts;
     using Implementation;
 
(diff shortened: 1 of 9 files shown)


Src/DependencyCollector/Shared/Implementation/HttpParsers/AzureQueueHttpParser.cs
```diff
@@ -27,37 +36,30 @@ internal static bool TryParse(ref DependencyTelemetry httpDependency)
...
-            var idx = nameWithoutVerb.IndexOf('/', isFirstSlash); // typically first symbol of the path is '/'
-            string queueName = idx != -1 ? nameWithoutVerb.Substring(isFirstSlash, idx - isFirstSlash) : nameWithoutVerb.Substring(isFirstSlash);
+            List<string> pathTokens = HttpParsingHelper.TokenizeRequestPath(nameWithoutVerb);
+            string queueName = pathTokens.Count > 0 ? pathTokens[0] : string.Empty;
 
             httpDependency.Type = RemoteDependencyConstants.AzureQueue;
-            httpDependency.Name = verb + account + '/' + queueName;
+            httpDependency.Name = string.IsNullOrEmpty(verb)
+                                      ? account + '/' + queueName
+                                      : verb + " " + account + '/' + queueName;
```
Dmitry-Matveev: Minor: Does "+" gives the best performance here with least memory footprint? I do not remember the most performant solution off top of my head but I remember that there were some "gotchas" around string immutability in .NET and Concat() vs. "+" vs. Builder for repetitive operations with similar strings. In particular cases one of them was way faster than others and used less memory.

PR #428, opened 2017-04-09
Role correlation

nizarq:
This change extends the correlation design for app map to be able to show relationships between services writing to the same AI resource.
Also includes some fixes to issues introduced changes made when Bcl async nuget package was removed.
I've also made changes to resolve my changes against Dan's dependency collector implemenatation for .net core.

### DIFF
--- a/Src/Common/CorrelationIdLookupHelper.cs
+++ b/Src/Common/CorrelationIdLookupHelper.cs
@@ -209,7 +209,7 @@
         /// <returns>App id.</returns>
         private Task<string> FetchAppIdFromService(string instrumentationKey)
         {
-            var task = new Task<string>(() =>
+            return Task.Factory.StartNew(() =>
             {
                 try
                 {
@@ -238,8 +238,6 @@
                     SdkInternalOperationsMonitor.Exit();
                 }
             });
-
-            return task;
         }
 #endif
         /// <summary>
--- a/Src/Common/CorrelationIdLookupHelper.cs
+++ b/Src/Common/CorrelationIdLookupHelper.cs
@@ -136,18 +136,6 @@
                             {
                                 this.GenerateCorrelationIdAndAddToDictionary(instrumentationKey, appId.Result);
                             }
-                            catch (AggregateException ex)
-                            {
-                                ex.Flatten();
-                                if (ex.InnerExceptions != null && ex.InnerExceptions.Count > 0 && ex.InnerExceptions[0] != null)
-                                {
-                                    this.RegisterFailure(instrumentationKey, ex.InnerExceptions[0]);
-                                }
-                                else
-                                {
-                                    this.RegisterFailure(instrumentationKey, ex);
-                                }
-                            }
                             catch (Exception ex)
                             {
                                 this.RegisterFailure(instrumentationKey, ex);
@@ -156,21 +144,6 @@
 
                         return false;
                     }
-                }
-                catch (AggregateException ex)
-                {
-                    ex.Flatten();
-                    if (ex.InnerExceptions != null && ex.InnerExceptions.Count > 0 && ex.InnerExceptions[0] != null)
-                    {
-                        this.RegisterFailure(instrumentationKey, ex.InnerExceptions[0]);
-                    }
-                    else
-                    {
-                        this.RegisterFailure(instrumentationKey, ex);
-                    }
-
-                    correlationId = string.Empty;
-                    return false;
                 }
                 catch (Exception ex)
                 {
@@ -287,6 +260,18 @@
         private void RegisterFailure(string instrumentationKey, Exception ex)
         {
 #if !NETCORE
+            var ae = ex as AggregateException;
+
+            if (ae != null)
+            {
+                ae = ae.Flatten();
+                if (ae.InnerException != null)
+                {
+                    RegisterFailure(instrumentationKey, ae.InnerException);
+                    return;
+                }
+            }
+
             var webException = ex as WebException;
             if (webException != null)
             {
--- a/Src/Common/CorrelationIdLookupHelper.cs
+++ b/Src/Common/CorrelationIdLookupHelper.cs
@@ -230,7 +230,7 @@
                 }
                 catch (Exception ex)
                 {
-                    CrossComponentCorrelationEventSource.Log.FetchAppIdFailed(this.GetExceptionDetailString(ex));
+                    AppMapCorrelationEventSource.Log.FetchAppIdFailed(this.GetExceptionDetailString(ex));
                     return null;
                 }
                 finally
@@ -267,7 +267,7 @@
                 ae = ae.Flatten();
                 if (ae.InnerException != null)
                 {
-                    RegisterFailure(instrumentationKey, ae.InnerException);
+                    this.RegisterFailure(instrumentationKey, ae.InnerException);
                     return;
                 }
             }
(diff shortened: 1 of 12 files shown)


Src/Common/CrossComponentCorrelationEventSource.cs
```diff
@@ -15,12 +15,12 @@
     /// <summary>
     /// ETW EventSource tracing class.
     /// </summary>
-    [EventSource(Name = "Microsoft-ApplicationInsights-Extensibility-CrossComponentCorrelation")]
-    internal sealed class CrossComponentCorrelationEventSource : EventSource
+    [EventSource(Name = "Microsoft-ApplicationInsights-Extensibility-AppMapCorrelation")]
+    internal sealed class AppMapCorrelationEventSource : EventSource
```
SergeyKanzhelev: Why rename? Can you please add it to the list of event sources in functional tests

nizarq: Because it's not just cross component correlation related anymore. It also deals with retrieving headers for correlating services within the same AI resource (so I just chose a generic name app map correlation).

I'm not sure which list in functional tests you are talking about. Can you please point me to it.

SergeyKanzhelev: Two of them:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/75373e57dcf8d646c54ee188461c373f2cc98939/Test/PerformanceCollector/FunctionalTests/Helpers/EtwEventSession.cs#L11

https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/de8b7c3a4595ae4de6625814a54462b16663d004/Test/Web/FunctionalTests/FunctionalTests/Helpers/EtwEventSession.cs#L11

nizarq: Thank you! Done!

SergeyKanzhelev:
Update `CHANGELOG.md` with the feature description please. I'm still reviewing and tryin to understand what this PR does

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -136,18 +136,6 @@ public bool TryGetXComponentCorrelationId(string instrumentationKey, out string
                             {
                                 this.GenerateCorrelationIdAndAddToDictionary(instrumentationKey, appId.Result);
                             }
-                            catch (AggregateException ex)
```
SergeyKanzhelev: so this change is irrelevant for the overall PR, correct?

nizarq: yes - that's just refactoring / cleanup. That's why that's in a separate commit.

Src/Common/RequestResponseHeaders.cs
```diff
@@ -13,12 +13,22 @@ internal static class RequestResponseHeaders
         /// <summary>
         /// Source key in the request context header that is added by an application while making http requests and retrieved by the other application when processing incoming requests.
```
SergeyKanzhelev: Was the `Request-Context` introduced in separate PR? I do not remember it. Can you please send me a link

Src/Common/RequestResponseHeaders.cs
```diff
@@ -13,12 +13,22 @@ internal static class RequestResponseHeaders
...
         /// <summary>
         /// Target key in the request context header that is added to the response and retrieved by the calling application when processing incoming responses.
         /// </summary>
-        public const string RequestContextTargetKey = "appId"; // Although the name of Source and Target key is the same - appId. Conceptually they are different and hence, we intentionally have two constants here. Makes for better reading of the code.
+        public const string RequestContextCorrleationTargetKey = "appId"; // Although the name of Source and Target key is the same - appId. Conceptually they are different and hence, we intentionally have two constants here. Makes for better reading of the code.
+
+        /// <summary>
+        /// Source-RoleName key in the request context header that is added by an application while making http requests and retrieved by the other application when processing incoming requests.
+        /// </summary>
+        public const string RequestContextSourceRoleNameKey = "sourceRoleName";
```
SergeyKanzhelev: why do you need `source` and `target` prefixes? Let's just do `roleName`

SergeyKanzhelev: Oh, just read your comment. 

> Although the name of Source and Target key is the same - appId. Conceptually they are different and hence, we intentionally have two constants here. Makes for better reading of the code.

What are you optimizing for? backend? SDK code? I think the only thing to optimize for is a customer code and what one see in fiddler. Hence just `roleName` will be better from my standpoint

Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
```diff
@@ -472,6 +472,22 @@ private void OnEnd(object exception, object thisObj, object returnValue)
...
+                                }
+                                catch (Exception ex)
+                                {
+                                    AppMapCorrelationEventSource.Log.GetComponentRoleNameHeaderFailed(ex.ToInvariantString());
+                                }
+
+                                if (!string.IsNullOrEmpty(targetRoleName))
+                                {
+                                    telemetry.Type = RemoteDependencyConstants.AI;
+                                    telemetry.Target += " | roleName:" + targetRoleName;
```
SergeyKanzhelev: do we need a space after `:`? you have spaces before and after `|` so why not here?

Src/Common/RequestResponseHeaders.cs
```diff
@@ -13,12 +13,22 @@ internal static class RequestResponseHeaders
...
         /// Source key in the request context header that is added by an application while making http requests and retrieved by the other application when processing incoming requests.
         /// </summary>
-        public const string RequestContextSourceKey = "appId";
+        public const string RequestContextCorrelationSourceKey = "appId";
 
         /// <summary>
         /// Target key in the request context header that is added to the response and retrieved by the calling application when processing incoming responses.
         /// </summary>
-        public const string RequestContextTargetKey = "appId"; // Although the name of Source and Target key is the same - appId. Conceptually they are different and hence, we intentionally have two constants here. Makes for better reading of the code.
+        public const string RequestContextCorrleationTargetKey = "appId"; // Although the name of Source and Target key is the same - appId. Conceptually they are different and hence, we intentionally have two constants here. Makes for better reading of the code.
```
Dmitry-Matveev: Minor: Typo in Correlation.

Src/Common/RequestResponseHeaders.cs
nizarq: I don't have a strong preference either ways. I can update per your suggestion but @karolz-ms had a strong inclination towards having source and target keywords in the headers. Karol can you please comment

Src/DependencyCollector/Shared/Implementation/ProfilerHttpProcessing.cs
nizarq: By design - We do have spaces around `|` but we don't have space around `:` for appId either. This is the format `cid-v1:{0}`. I'll keep things consistent by not adding spaces

Src/Common/RequestResponseHeaders.cs
karolz-ms: @SergeyKanzhelev @nizarq IMO "roleName" is ambiguous enough to warrant clarification--is it the name of the role that makes the request or the name of the role that handles the request? For the purpose of debugging I think it is better to have a prefix that differentiates between the two, i.e. "source-rolename" or "client-rolename" vs. "target-rolename" or "server-rolename".

PR #429, opened 2017-04-09
Adding support to inject headers and intercept Http outgoing requests

brahmnes:
using DiagnosticSource. This frees the user from having to use the
profiler.  Also introduced HttpProcessing as the base class for
FrameworkHttpProcessing and ProfilerHttpProcessing to unify the logic
from both paths in handling correlation headers and dependency telemetry
handling.

A temporary HttpHandlerDiagnosticListener implementation is added to
make all this work without a new version of DiagnosticSource nuget. This
class is currently under reviewed in the corefx repo and will be part of
DiagnosticSource. Once that happens, we will remove this temporary
implementation of HttpHandlerDiagnosticListener.


Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -0,0 +1,361 @@
...
+
+            servicePointTableField.SetValue(null, newTable);
+        }
+
+        /// <summary>
+        /// Initializes all the reflection objects it will ever need. Reflection is costly, but it's better to take
+        /// this one time performance hit than to get it multiple times later, or do it lazily and have to worry about
+        /// threading issues. If Initialize has been called before, it will not doing anything.
+        /// </summary>
+        private void Initialize()
```
pharring: I think you should use a static constructor here and avoid the need for initializationLock.

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -0,0 +1,361 @@
...
+    /// <summary>
+    /// A HttpHandlerDiagnosticListener is a DiagnosticListener for .NET 4.5/4.6 where HttpClient
+    /// doesn't have a DiagnosticListener built-in. This class is not used for .NET Core because
+    /// HttpClient in .NET Core already emits DiagnosticSource events. This class compensates for
+    /// that in .NET 4.5/4.6. HttpHandlerDiagnosticListener has no public constructor, but it has a
+    /// singleton style Initialize method to control its creation.
+    /// To use this, the application just needs to call HttpHandlerDiagnosticListener.Initialize(), and this
+    /// will register itself with the DiagnosticListener's all listeners collection.
+    /// </summary>
+    internal class HttpHandlerDiagnosticListener : DiagnosticListener
```
pharring: sealed

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -0,0 +1,361 @@
...
+    {
+        internal static HttpHandlerDiagnosticListener SingletonInstance = new HttpHandlerDiagnosticListener();
```
pharring: readonly

pharring: You could call this just ```Instance```. The "Singleton" prefix doesn't really add anything that ```static``` doesn't already tell you.

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -0,0 +1,361 @@
...
+    using System.Reflection;
+    using System.Reflection.Emit;
+
+    /// <summary>
+    /// A HttpHandlerDiagnosticListener is a DiagnosticListener for .NET 4.5/4.6 where HttpClient
+    /// doesn't have a DiagnosticListener built-in. This class is not used for .NET Core because
+    /// HttpClient in .NET Core already emits DiagnosticSource events. This class compensates for
+    /// that in .NET 4.5/4.6. HttpHandlerDiagnosticListener has no public constructor, but it has a
+    /// singleton style Initialize method to control its creation.
+    /// To use this, the application just needs to call HttpHandlerDiagnosticListener.Initialize(), and this
```
pharring: Is this right? Initialize is private. How does the application call it?

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -0,0 +1,361 @@
...
+        internal static HttpHandlerDiagnosticListener SingletonInstance = new HttpHandlerDiagnosticListener();
+
+#region private fields
+        private const string DiagnosticListenerName = "System.Net.Http.Desktop.V4";
+        private const string RequestWriteName = "System.Net.Http.Request";
+        private const string ResponseWriteName = "System.Net.Http.Response";
+        private const string ExceptionEventName = "System.Net.Http.Exception";
+
+        // Fields for reflection
+        private static FieldInfo connectionGroupListField;
```
pharring: If you use a static constructor instead of Initialize, then all these static fields should become ```readonly```

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -0,0 +1,361 @@
...
+        private static Func<HttpWebRequest, HttpWebResponse> httpResponseAccessor;
+
+        // Fields for controlling initialization of the HttpHandlerDiagnosticListener singleton
+        private object initializationLock = new object();
+        private bool initialized = false;
+#endregion
+
+        /// <summary>
+        /// Prevents a default instance of the <see cref="HttpHandlerDiagnosticListener" /> class from being created.
+        /// Make sure every caller has to create this object via the Initialize method.
```
pharring: Comment out of date? Seems that (internal) callers just access the singleton, right?

Src/DependencyCollector/Shared/Implementation/HttpDiagnosticSourceSubscriber.cs
```diff
@@ -0,0 +1,109 @@
...
+    internal class HttpDiagnosticSourceSubscriber : IObserver<DiagnosticListener>, IDisposable
+    {
+        private HttpDiagnosticSourceListener parent;
+        private IDisposable allListenersSubscription;
+        private IDisposable sourceSubscription;
+        private bool disposed = false;
+
+        internal HttpDiagnosticSourceSubscriber(HttpDiagnosticSourceListener parent)
+        {
+            GC.KeepAlive(HttpHandlerDiagnosticListener.SingletonInstance);
```
pharring: GC,KeepAlive is a bit odd (I don't think the GC will ever collect static fields), but perhaps it captures the intent: This is the place to create the singleton and to hook up its events. Right?

pharring: Also, is this bit of code going to migrate into Corefx too? If not, then the SingletonInstance presumably needs to be made public.

SergeyKanzhelev:
>Once that happens, we will remove this temporary
implementation of HttpHandlerDiagnosticListener.

@brahmnes are we talking `2.4.0-beta1` or `2.4..0` timeframe?

SergeyKanzhelev:
Please update `CHANGELOG.md`

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -0,0 +1,361 @@
...
+        /// </summary>
+        public override IDisposable Subscribe(IObserver<KeyValuePair<string, object>> observer, Func<string, object, object, bool> isEnabled)
+        {
+            this.Initialize();
+            return base.Subscribe(observer, isEnabled);
+        }
+
+#region private methods
+
+        private static void PrepareReflectionObjects()
```
SergeyKanzhelev: We need somebody from System.Net to sign off on these changes.

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -0,0 +1,361 @@
...
+                    this.initialized = true;
+
+                    try
+                    {
+                        PrepareReflectionObjects();
+                        PerformInjection();
+                    }
+                    catch (Exception)
+                    {
+                        // If anything went wrong, just no-op
```
SergeyKanzhelev: >// If anything went wrong, just no-o [](start = 24, length = 36)

Please add tracing to troubleshoot why exactly nothing works

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -0,0 +1,361 @@
...
+        {
+            this.Initialize();
+            return base.Subscribe(observer, isEnabled);
+        }
+
+#region private methods
+
+        private static void PrepareReflectionObjects()
+        {
+            // At any point, if the operation failed, it should just throw, including NullReferenceException.
```
SergeyKanzhelev: >// At any point, if the operation failed, it should just throw, including NullReferenceException. [](start = 12, length = 97)

please minimize the number of thrown exceptions. Ideally make it `TryPrepareReflectionObjects` instead and return the inner class containing all the needed reflection types/delegates

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -0,0 +1,361 @@
...
+
+            // Double checking to make sure we have all the pieces initialized
+            if (connectionGroupListField == null ||
+                connectionGroupType == null ||
+                connectionListField == null ||
+                connectionType == null ||
+                writeListField == null ||
+                httpResponseAccessor == null)
+            {
+                throw new InvalidOperationException("Unable to initialize all required reflection objects");
```
SergeyKanzhelev: >"Unable to initialize all required reflection objects" [](start = 52, length = 54)

if this class would become the part of Diagnostics Source - I'd suggest to minimize the number of places your use English strings as ultimately all of them will need to be translated. Application Insights at the moment has more relaxed requirements on localization. So maybe just return false.

Src/DependencyCollector/Net45/DependencyCollector.Net45.csproj
```diff
@@ -45,6 +45,7 @@
     <Reference Include="System.Windows.Forms" />
     <Reference Include="System.Xml" />
     <Reference Include="System.Xml.Linq" />
+    <Reference Include="Microsoft.CSharp" />
```
SergeyKanzhelev: >"Microsoft.CSharp" /> [](start = 23, length = 21)

is it needed for `Reflection.Emit`?

Src/DependencyCollector/Shared/Implementation/FrameworkHttpProcessing.cs
```diff
@@ -1,40 +1,42 @@
...
+    /// received from framework events for HTTP. This class receives and uses two types of events, which
+    /// are EventSource from FrameworkHttpEventListener, and DiagnosticSource from
+    /// HttpDiagnosticSourceListener. The challenge is, the diagnostic source events have the WebRequest
+    /// object which we need for header injection, but the events can be fired multiple times. So you won't
+    /// know if it's the first request, or if it's the last response. The event source events fire at the right
+    /// locations and just once, but they don't have the rich information. This class coordinates both events,
+    /// store information in DependencyTelemetry properly, and fire the telemetry only on the EventSource
+    /// response receive to guarantee it's done just once and at the right time.
+    /// </summary>
+    internal sealed class FrameworkHttpProcessing : HttpProcessing
```
SergeyKanzhelev: >Framework [](start = 26, length = 9)

`Framework` prefix meant "collected via built-in framewrok instrumentation". As opposite to `Profiler` prefix. I'd suggest to differentiate RDD collected via EventSource and Diagnostics Source and has their prefixes separately.

Src/DependencyCollector/Shared/Implementation/FrameworkHttpProcessing.cs
```diff
@@ -1,40 +1,42 @@
...
     {
         internal CacheBasedOperationHolder TelemetryTable;
         private readonly ApplicationInsightsUrlFilter applicationInsightsUrlFilter;
-        private readonly TelemetryClient telemetryClient;
 
-        internal FrameworkHttpProcessing(TelemetryConfiguration configuration, CacheBasedOperationHolder telemetryTupleHolder)
+        internal FrameworkHttpProcessing(TelemetryConfiguration configuration, CacheBasedOperationHolder telemetryTupleHolder, bool setCorrelationHeaders, ICollection<string> correlationDomainExclusionList, string appIdEndpoint)
+            : base(configuration, SdkVersionUtils.GetSdkVersion("rdd" + RddSource.Framework + ":"), null, setCorrelationHeaders, correlationDomainExclusionList, appIdEndpoint)
```
SergeyKanzhelev: >"rdd" + RddSource.Framework [](start = 64, length = 27)

need new sdk version prefix for diagnostics source

Src/DependencyCollector/Shared/Implementation/HttpProcessing.cs
```diff
@@ -0,0 +1,404 @@
...
+#if NET40
+    using Microsoft.ApplicationInsights.Net40;
+#endif
+    using Microsoft.ApplicationInsights.Web.Implementation;
+
+    /// <summary>
+    /// Concrete class with all processing logic to generate RDD data from the callbacks
+    /// received from Profiler instrumentation for HTTP or HTTP EventSource/DiagnosticSource events.   
+    /// </summary>
+    internal abstract class HttpProcessing
```
SergeyKanzhelev: >abstract [](start = 13, length = 8)

we found this generalization a bad idea. Abstract class starts knowing all sorts of internal details of children classed. In this code review - Profiler and Diagnostics Source may share headers reading and injection logic as well as appending VERB to the name. So perhaps should share the base class as well. Also adding new collectors becoming a very hard task as you may need to refactor how the base class works. Consider adding helper classes instead of the base class.

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
```diff
@@ -177,9 +183,12 @@ protected virtual void Dispose(bool disposing)
         private void InitializeForFrameworkEventSource()
```
SergeyKanzhelev: >InitializeForFrameworkEventSource [](start = 21, length = 33)

Now you have two implementations  - one that requires reflection and may fail to initialize and another that requires less. I do not see the fall back logic implemented. Am I missing something?

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
brahmnes: This file is temporary. The real implementation will sit in System.Diagnostics.DiagnosticSource, see this review at:
https://github.com/dotnet/corefx/pull/18101

Once it's in DiagnosticSource dll, this file will be removed. It also makes more sense if you look at how we trigger the initialization from DiagnosticListener, it becomes clear that you don't want to use a static constructor because it should get initialized only if someone is subscribing for listening.

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
brahmnes: Sure

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
brahmnes: I will leave it as is since it passes Stylecop now, and this file will be removed in a week or two (hopefully). In DiagnosticSource, this is named as s_instance and it's consistent with fields in DiagnosticListener

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
brahmnes: The GC.KeepAlive on the instance field initializes it and calls the constructor, which then inserts itself into DiagnosticListener.AllListeners. Then the caller gets a callback in OnNext for subscribing, which then triggers the initialization. The flow is a little complicating, but that's basically how DiagnosticListener subscription works today.

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
brahmnes: Again, it's a initialized only on subscribe, so can't use static constructor here.

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
brahmnes: Yes, will update the comment.

Src/DependencyCollector/Shared/Implementation/HttpDiagnosticSourceSubscriber.cs
brahmnes: Yes, this code will go into corefx.
You are correct that GC.KeepAlive is probably not necessary, your assumption is correct.

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
brahmnes: There is a parallel PR in corefx:
https://github.com/dotnet/corefx/pull/18101

brahmnes: To your earlier question, this file will be removed hopefully within in a few weeks. So yes, it's in 2.4.0 timeframe.

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
brahmnes: I probably won't change this here. I will look at corefx to see what tracing they use there

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
brahmnes: The FrameworkHttpEventListener path is unchanged. FrameworkHttpProcessing.OnBeginHttpCallback and OnEndHttpCallback are still there. The worst case is, you just get what you get today.

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
brahmnes: Will take your suggestion about changing these functions to return bool, so these strings should go away.

Src/DependencyCollector/Net45/DependencyCollector.Net45.csproj
brahmnes: It's needed for the "dynamic" keyword

Src/DependencyCollector/Shared/Implementation/HttpDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,105 @@
...
+                        this.httpProcessingFramework.OnRequestSend((HttpWebRequest)propertyBag.Request);
+                        break;
+                    case "System.Net.Http.Response":
+                        this.httpProcessingFramework.OnResponseReceive((HttpWebRequest)propertyBag.Request, (HttpWebResponse)propertyBag.Response);
+                        break;
+                }
+            }
+            catch (Exception)
+            {
+                // This shouldn't happen, but just in case it does, we shouldn't break the application.
```
lmolkova: as far as i can tell, all exceptions are handled in OnBegin/OnEnd.
However if there is a reason to catch them, consider writing error to the event log

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
SergeyKanzhelev: We will need tracing in Application Insights as well. We get support questions on why dependencies weren't collected way too often

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
brahmnes: Addressing it in the corefx repo

Src/DependencyCollector/Shared/Implementation/FrameworkHttpProcessing.cs
brahmnes: It's the same processing class, you cannot split this between Diagnostic and Framework.

Src/DependencyCollector/Shared/Implementation/FrameworkHttpProcessing.cs
brahmnes: I don't understand that that means. Can you tell me specifically what I need to change?

Src/DependencyCollector/Shared/Implementation/HttpProcessing.cs
brahmnes: This is done to minimize changes to the ProfilerHttpProcessing class as much as possible. I believe given the current circumstances, this is probably the best option with lowest chance of regressing tests and functionality and be able to ship on time. If I start refactoring this into a helper class, a whole bunch of references and unit tests would break because the things they reference into ProfilerHttpProcessing have all of a sudden gone missing. To really address it the way you want, much more time is needed with consideration of what our test strategy should be.

Src/DependencyCollector/Shared/Implementation/HttpDiagnosticSourceListener.cs
brahmnes: Adding a call to DependencyCollectorEventSource

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
brahmnes: Actually it looks like this is not an issue in the corefx repo, will keep the exceptions and strings. The exception will be thrown at most once during initialization, so it's not high traffic.

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
brahmnes: Will write an event.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - Report status code for the dependencies failed with non-protocol issue like DNS resolution or SSL shakeup problems.
 - Implemented automatic telemetry correlation: all telemetry reported within the scope of the request is correlated to RequestTelemetry reported for the request.
 - Implemented [Correlation HTTP protocol](https://github.com/lmolkova/correlation/blob/master/http_protocol_proposal_v1.md): default headers to pass Operation Root Id and Parent Id were changed. This version is backward compatible with previously supported headers. 
+- Implemented injection into the HTTP stack to leverage DiagnosticSource to gain access to the WebRequest and WebResponse objects for header injections, without the need of using the profiler.
```
cijothomas: does this work in all .net versions? if not, can you mention it in here?

Src/DependencyCollector/Net45.Tests/DependencyTrackingTelemetryModuleTest.cs
```diff
@@ -64,7 +79,20 @@ public void TestHttpPostRequestsAreCollected()
         public void TestHttpRequestsWithQueryStringAreCollected()
         {
             ITelemetry sentTelemetry = null;
-            var channel = new StubTelemetryChannel { OnSend = telemetry => sentTelemetry = telemetry };
+
+            var channel = new StubTelemetryChannel
```
cijothomas: this stubchannel code can be reused without copy pasting for every method.

Src/DependencyCollector/Shared.Tests/Implementation/FrameworkHttpProcessingTest.cs
```diff
@@ -21,17 +25,18 @@ namespace Microsoft.ApplicationInsights.DependencyCollector.Implementation
     [TestClass]
     public sealed class FrameworkHttpProcessingTest : IDisposable
     {
-        #region Fields
-        private const string TestUrl = "http://www.microsoft.com/";
-        private const string TestUrlNonStandardPort = "http://www.microsoft.com:911/";
+#region Fields
+        private const string RandomAppIdEndpoint = "http://app.id.endpoint"; // appIdEndpoint - this really won't be used for tests because of the app id provider override.
```
cijothomas: good comments :)

CHANGELOG.md
brahmnes: It's only for .NET46 and above, and not .NET core.
Sure, I can clarify

Src/DependencyCollector/Net45.Tests/DependencyTrackingTelemetryModuleTest.cs
```diff
@@ -64,7 +79,20 @@ public void TestHttpPostRequestsAreCollected()
...
+            {
+                OnSend = telemetry =>
+                {
+                    // The correlation id lookup service also makes http call, just make sure we skip that
+                    DependencyTelemetry depTelemetry = telemetry as DependencyTelemetry;
+                    if (depTelemetry != null && !depTelemetry.Data.StartsWith(FakeProfileApiEndpoint, StringComparison.OrdinalIgnoreCase))
```
Dmitry-Matveev: Minor: It looks like Data can be null if Dependency telemetry was initialized with default constructor. This particular test should not be affected.

Src/DependencyCollector/Shared.Tests/Implementation/FrameworkHttpProcessingTest.cs
```diff
@@ -211,61 +229,275 @@ public void OnEndHttpCallbackSetsSuccessToFalseForMoreThan400()
...
-        /// Validates HttpProcessingFramework calculates startTime from the start of very first BeginGetRequestStream if any
+        /// Validates that even if multiple events have fired, as long as there is only
+        /// one HttpWebRequest, only one event should be written, during the first call
+        /// to OnEndHttpCallback.
+        /// </summary>
+        [TestMethod]
+        public void RddTestHttpProcessingFrameworkNoDuplication()
+        {
+            Stopwatch stopwatch = Stopwatch.StartNew();
+            HttpWebRequest request = (HttpWebRequest)WebRequest.Create(this.testUrl);
```
Dmitry-Matveev: Is it possible that customers reuse the same request on purpose, e.g. for some "ping" test or it will be returned from cache and they would need to create a new one each time to go via network? I'm asking to make sure we do not have an edge case here where the behavior should be the opposite (collect all, not the first one).

Src/DependencyCollector/Shared/Implementation/FrameworkHttpProcessing.cs
cijothomas: currently rddf vs rddp identifies whether the dependency was collected with help of framework event source or via profiler. Since diagnostics source is a third mechanism (its different from framework eventsource) we need a way to know if the dependency was collected with it.

Dmitry-Matveev: I think they mean to add new value to RddSource enumeration like "ds" (after diagnostic source) and use it here instead of the RddSource.Framework.

Src/DependencyCollector/Net45.Tests/DependencyTrackingTelemetryModuleTest.cs
brahmnes: I tried to factor this out, but the lambda capture of sentTelemetry makes it hard to put it in a separate method.
Non-ideal, but leaving it this way may be the easiest.

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
Dmitry-Matveev: If it's not removed from 2.4 by stable milestone, I would recommend to implement `TryPrepare..()` in 2.4 context rather than wait for CoreFX, we would like to minimize # of the exceptions that are just thrown/swallowed especially on the "happy path".

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -0,0 +1,365 @@
...
+            // At any point, if the operation failed, it should just throw, including NullReferenceException.
+            // The caller should catch all exceptions and swallow.
+
+            // First step: Get all the reflection objects we will ever need.
+            Assembly systemNetHttpAssembly = typeof(ServicePoint).Assembly;
+            connectionGroupListField = typeof(ServicePoint).GetField("m_ConnectionGroupList", BindingFlags.Instance | BindingFlags.NonPublic);
+            connectionGroupType = systemNetHttpAssembly?.GetType("System.Net.ConnectionGroup");
+            connectionListField = connectionGroupType?.GetField("m_ConnectionList", BindingFlags.Instance | BindingFlags.NonPublic);
+            connectionType = systemNetHttpAssembly?.GetType("System.Net.Connection");
+            writeListField = connectionType?.GetField("m_WriteList", BindingFlags.Instance | BindingFlags.NonPublic);
```
Dmitry-Matveev: Do we need to assert for `ReflectionPermission` here before accessing `Nonpublic` fields?

Src/DependencyCollector/Net45.Tests/DependencyTrackingTelemetryModuleTest.cs
cijothomas: ok

Src/DependencyCollector/Shared/Implementation/FrameworkHttpProcessing.cs
brahmnes: Got it, fixed.

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
brahmnes: This file will certainly be removed from AI SDK.

Src/DependencyCollector/Net45.Tests/DependencyTrackingTelemetryModuleTest.cs
brahmnes: Well, this is test code if the assumption wrong, it will fail and we will know. ;)

Src/DependencyCollector/Shared.Tests/Implementation/FrameworkHttpProcessingTest.cs
brahmnes: .NET framework recommends that HttpWebRequest is not reused for multiple requests. In fact, this assumption was already in place in ProfilerHttpProcessing before this change, it used to look up the web request object from the cache and if it exists, it skips processing. This change doesn't change semantics.

Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs
brahmnes: It's better to just attempt the operation rather than checking it. If you have no access, it will just fail, and execution goes on. It falls back to just framework events without diagnostic source and no harm is done.

SergeyKanzhelev:
@brahmnes please resolve the conflicts

cijothomas:
Dont merge this until Functional tests are also updated.

brahmnes:
This PR is replaced with https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/449

PR #430, opened 2017-04-10
Normalized Process CPU Performance counter

iusafaro:


Src/PerformanceCollector/Shared/Implementation/StandardPerformanceCollector/NormalizedProcessCPUPerformanceCounter.cs
```diff
@@ -0,0 +1,70 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.Implementation.StandardPerformanceCollector
+{
+    using System;
+    using System.Diagnostics;
+    using System.Globalization;
+
+    /// <summary>
+    /// Interface represents normalized value of CPU Utilization by Process counter value.
```
cijothomas: not an interface.

Src/PerformanceCollector/Shared/Implementation/StandardPerformanceCollector/NormalizedProcessCPUPerformanceCounter.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        /// <summary>
+        /// Returns the current value of the counter as a <c ref="MetricTelemetry"/>.
+        /// </summary>
+        /// <returns>Value of the counter.</returns>
+        public double Collect()
+        {
+            try
+            {
+                return this.performanceCounter.NextValue() / Environment.ProcessorCount;
```
cijothomas: whats the behavior when no permission to read Environment variable? We keep throwing exception every time counter is refreshed?
Can we check if permission to ENV exists first, and if not, dont every bother with this counter.?

Src/PerformanceCollector/Shared/Implementation/PerformanceCounterData.cs
```diff
@@ -17,11 +17,27 @@ internal class PerformanceCounterData
             this.UsesInstanceNamePlaceholder = usesInstanceNamePlaceholder;
             this.IsCustomCounter = isCustomCounter;
             this.IsInBadState = isInBadState;
```
dnduffy: What is "IsInBadState" used for?  And how do you know if that is true or false when you create an instance of this class?

Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/CPUPercenageGauge.cs
```diff
@@ -37,6 +37,15 @@ public CPUPercenageGauge(string name, ICounterValue value)
...
         public double GetValueAndReset()
         {
+            return this.Collect();
+        }
+
+        /// <summary>
+        /// Returns the percentage of the CPU process utilization time with respect to the total duration.
+        /// </summary>
+        /// <returns>The value of the target metric.</returns>
+        protected virtual double Collect()
```
cijothomas: There is a typo in this file/class name. CPUPercenageGauge to CPUPercentageGauge - can you fix that as well if time permits.

Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/NormalizedCPUPercenageGauge.cs
```diff
@@ -0,0 +1,78 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.Implementation.WebAppPerformanceCollector
+{
+    using System;
+    using System.Globalization;
+
+    /// <summary>
+    /// Gauge that computes normalized CPU percentage utilized by a process by utilizing the last computed time.
```
cijothomas: Mention in comment that Normalized means taking into account the # of CPU Cores for clarity.

Src/PerformanceCollector/Unit.Tests/WebAppPerformanceCollector/NormalizedCPUPercenageGaugeTests.cs
```diff
@@ -0,0 +1,44 @@
...
+    [TestClass]
+    public class NormalizedCPUPercenageGaugeTests
+    {
+        [TestMethod]
+        public void NormalizedCPUPercenageGaugeBasicValidation()
+        {
+            int initialProcessorsCount = int.Parse(Environment.GetEnvironmentVariable("NUMBER_OF_PROCESSORS"));
+            NormalizedCPUPercenageGauge normalizedGauge = new NormalizedCPUPercenageGauge(
+                "CPU",
+                new RawCounterGauge(@"\Process(??APP_WIN32_PROC??)\Private Bytes * 2", "userTime", AzureWebApEnvironmentVariables.App, new CacheHelperTests()));
```
cijothomas: Private bytes?

Src/PerformanceCollector/Unit.Tests/WebAppPerformanceCollector/NormalizedCPUPercenageGaugeTests.cs
```diff
@@ -0,0 +1,44 @@
...
+
+            CPUPercenageGauge gauge = new CPUPercenageGauge(
+                "CPU",
+                new RawCounterGauge(@"\Process(??APP_WIN32_PROC??)\Private Bytes * 2", "userTime", AzureWebApEnvironmentVariables.App, new CacheHelperTests()));
```
cijothomas: private bytes?

cijothomas:
@tokaplan Can you also review?

Src/PerformanceCollector/Shared/Implementation/StandardPerformanceCollector/CounterFactory.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        /// <summary>
+        /// Gets a counter.
+        /// </summary>
+        /// <param name="originalString">Original string definition of the counter.</param>
+        /// <param name="categoryName">Category name.</param>
+        /// <param name="counterName">Counter name.</param>
+        /// <param name="instanceName">Instance name.</param>
+        /// <returns>The counter identified by counter name.</returns>
+        internal ICounterValue GetCounter(string originalString, string categoryName, string counterName, string instanceName)
```
SergeyKanzhelev: Can it be aligned with the factory for azure web site? Move parsing of `originalString` here

Src/PerformanceCollector/Shared/Implementation/StandardPerformanceCollector/NormalizedProcessCPUPerformanceCounter.cs
SergeyKanzhelev: Or just cache the value in local variable

Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/NormalizedCPUPercenageGauge.cs
```diff
@@ -0,0 +1,78 @@
...
+    /// </summary>
+    internal class NormalizedCPUPercenageGauge : CPUPercenageGauge
+    {
+        /// <summary>Specific environment variable for Azure App Services.</summary>
+        private const string ProcessorsCounterEnvironmentVariable = "NUMBER_OF_PROCESSORS";
```
SergeyKanzhelev: Ideally, just have a gauge that devides value from `ICounterProvider` to whatever you pass in constructor. And pass processor count in constructor with all the checks on being non-negative and not too huge

Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/CPUPercenageGauge.cs
iusafaro: done

Src/PerformanceCollector/Shared/Implementation/PerformanceCounterData.cs
iusafaro: It's defined based on the first collection attempt - if first read failed then counter is in the bad state and nobody will try to read it again.

Src/PerformanceCollector/Shared/Implementation/StandardPerformanceCollector/NormalizedProcessCPUPerformanceCounter.cs
iusafaro: In this case first read will also fail -> counter marked as in the bad state -> nobody will try to read it again

Src/PerformanceCollector/Shared/Implementation/PerformanceCounterStructure.cs
```diff
@@ -0,0 +1,38 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.Implementation
+{
+    /// <summary>Represents basic performance counter structure.</summary>
```
iusafaro: parsing of the full counter name into three pieces was made before based on PerformanceCounter class, now it should be more abstract from PerfCounter realization as we can have logical ones that are not mapped to the physical. also those objects were actually not used, but PerformanceCounter is disposable and nobody dealt with it correctly.

SergeyKanzhelev: If you need one - you probably doing something wrong. What we need is `reportAs` name and a class the capable extracting counters. There should be no reason to keep around the original category, counter name and instance names.

Src/PerformanceCollector/Shared/Implementation/StandardPerformanceCollector/CounterFactory.cs
iusafaro: More complicated as azure web apps don't worry about instance name as they receive snapshot for that particular process, here we need to move all logic for instance definition in CounterFactory and it requires much more refactoring without any real benefits

Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/NormalizedCPUPercentageGauge.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// <returns>The value of the target metric.</returns>
+        protected override double Collect()
+        {
+            if (!this.isInitialized)
+            {
+                lock (this.lockObject)
+                {
+                    if (!this.isInitialized)
+                    {
+                        this.processorsCount = this.GetProcessorsCount();
```
Dmitry-Matveev: Does the same logic apply here as with Non-WebApp counters? E.g. if it fails to get processors count once - it will stop trying? Based on this code it looks like it will keep trying every time because `isInitialized` with stay false. In this case we can think of moving it out of `Collect` to only call it once.

Src/PerformanceCollector/Shared/Implementation/StandardPerformanceCollector/CounterFactory.cs
SergeyKanzhelev: It is important to have error processing consistent. Easiest way to ensure it is to have code consistent. It may be ok to keep it as is for now.

Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/NormalizedCPUPercentageGauge.cs
iusafaro: added throwing an exception and now it's fully consistent with standard counter

Src/PerformanceCollector/Shared/Implementation/PerformanceCounterStructure.cs
iusafaro: maybe. but it's not my invention - i just moved these fields a little bit - and it requires a lot of refactoring to get rid of it.

Src/PerformanceCollector/Shared/Implementation/WebAppPerformanceCollector/NormalizedCPUPercentageGauge.cs
SergeyKanzhelev: Could you please not throw - change standard counters instead of aligning with them. Exceptions from performance counters collection are dominant in the exceptions statistics feature.

PR #433, opened 2017-04-11
Add dependency to DiagnosticSource in nuspecs

lmolkova:
This change consolidates different versions of DiagnosticSource referenced and adds dependency to DiagnosticSource in Web and DepenendecyCollector nuspec files

### DIFF
--- a/Src/DependencyCollector/NuGet/Package.nuspec
+++ b/Src/DependencyCollector/NuGet/Package.nuspec
@@ -23,12 +23,13 @@
       <group targetFramework="net45">
         <dependency id="Microsoft.ApplicationInsights" version="[$coresdkversion$]" />
         <dependency id="Microsoft.ApplicationInsights.Agent.Intercept" version="2.0.7" />
+        <dependency id="System.Diagnostics.DiagnosticSource" version="4.4.0-preview1-25210-01" />
       </group>
       <group targetFramework="netcoreapp1.0">
         <dependency id="Microsoft.ApplicationInsights" version="[$coresdkversion$]" />
         <dependency id="Microsoft.Extensions.DiagnosticAdapter" version="1.0.0" />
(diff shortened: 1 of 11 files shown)


Src/DependencyCollector/NuGet/Package.nuspec
```diff
@@ -23,12 +23,13 @@
       <group targetFramework="net45">
         <dependency id="Microsoft.ApplicationInsights" version="[$coresdkversion$]" />
         <dependency id="Microsoft.ApplicationInsights.Agent.Intercept" version="2.0.7" />
+        <dependency id="System.Diagnostics.DiagnosticSource" version="4.4.0-preview1-25210-01" />
```
Dmitry-Matveev: We are about to start releasing 2.4 Beta 1 in couple days - is this package present on the nuget servers reachable by customers? I looked at nuget.org and the last version is 4.3.0, however, this one might not just be visible.

lmolkova:
`<dependency id="System.Diagnostics.DiagnosticSource" version="4.4.0-preview1-25210-01" />`
this is a [minimum](https://docs.microsoft.com/en-us/nuget/create-packages/dependency-versions) version.
And this dependency needs to be fixed once stable DiagnosticSource is released.
@Dmitry-Matveev back to the question on availability.
I'm working on publishing the package, ETA 1-2 days.

@brahmnes is also doing changes (https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/429) to inject request headers and parse response headers without profiler.
Part of his change belong to DiagnositcSource, but not merged yet. It is **duplicated** now in AppInsights.
To prevent double-instrumentation we would have to 
1. specify **maximum** version of DiagSource (without Brahmnes changes) in the beta1. 
2. Release another version of DiagSource with the changes 
3. In the next AI beta remove duplicated code and specify a **minimum** DiagSource version, that has the instrumentation

PR #434, opened 2017-04-11
Add excluded domains list to .NET Core DependencyCollector

daschult:


Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
```diff
@@ -111,7 +110,9 @@ public void OnRequest(HttpRequestMessage request, Guid loggingRequestId)
         {
             try
             {
-                if (request != null && request.RequestUri != null && !this.applicationInsightsUrlFilter.IsApplicationInsightsUrl(request.RequestUri.ToString()))
+                if (request != null && request.RequestUri != null &&
+                    !this.applicationInsightsUrlFilter.IsApplicationInsightsUrl(request.RequestUri.ToString()) &&
+                    !this.correlationDomainExclusionList.Contains(request.RequestUri.Host))
```
SergeyKanzhelev: why are you mixing those two? `correlationDomainExclusionList` means "do not correlate", not "do not collect"

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
```diff
@@ -27,38 +27,37 @@ public class DependencyCollectorDiagnosticListener : IObserver<DiagnosticListene
         /// <returns>
         /// An IDisposable that can be disposed to disable the DependencyCollectorDiagnosticListener.
         /// </returns>
-        public static IDisposable Enable(TelemetryConfiguration configuration = null)
+        public static IDisposable Enable(TelemetryConfiguration configuration = null, IEnumerable<string> excludedDomains = null)
```
SergeyKanzhelev: I cannot find code that calls this `Enable` and passes the list of excluded domains

daschult: This is only exercised in the unit tests because Enable() is a function that will be called by users.

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
daschult: Fixed.

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
dnduffy: @SergeyKanzhelev We're going to use this when we call it from the .NET Core SDK.

dnduffy: @daschult Who's job is it to dispose of the IDisposable that Enable returns?  Would the .NET Core SDK need to keep a reference around until the application is torn down and remember to dispose of it?

SergeyKanzhelev: Wait, this should be called from `Initialize` of the DependencyTelemetryModule. So enabling of that telemetry module will ***automatically*** enable dependencies tracking.

daschult: No, you're referring to the DependencyTrackingTelemetryModule class, and that is only used in the .NET Framework scenario. This is specifically for the .NET Core scenario. In this case, a user has to explicitly call DependencyCollectorDiagnosticListener.Enable() to enable the dependency collector in their .NET Core application.

SergeyKanzhelev: Nope, `DependencyTrackingTelemetryModule` should be for all scenarios. Customer's code should not differ in .NET Core and Full framework cases. Just do `(new DependencyTrackingTelemetryModule()).Initializer(TelemetryConfiguraiton.Active);` and you have your dependencies collected.

Please change

daschult: Sure. I can do that. Just to let you know, that class is basically going to be hidden behind a #if !NETCORE ... #endif because almost all of it isn't compatible with .NET Core. Is that what you want?

dnduffy: @SergeyKanzhelev Aren't we going to have to call this differently from the .NET Core SDK to provide the list of domains to exclude because we don't have the config file that the web SDK module has access to?

SergeyKanzhelev: @daschult - the scenario it should support is:

1. Install Dependency Telemetry Module.
2. Call `(new DependencyTrackingTelemetryModule()).Initializer(TelemetryConfiguraiton.Active); ` implicitly or explicitly.
3. See dependencies in the portal.

@dnduffy - for now - browse thru the collection of telemetry modules in services. Find `DependencyTrackingTelemetryModule`. Set the additional configuration settings like the list of excluded modules.

lmolkova:
There is also `HttpDependenciesParsingTelemetryInitializer` that properly fills dependency type. It is configured with xml appinsights config that is not applicable to AspNetCore apps.
So it is never be invoked with DepedendeycCollector on .NET Core.

I believe it is related to this PR.

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
dnduffy: @SergeyKanzhelev We need a much better story than that, we already have a problem with really obtuse configuration and this example will become even worse.

dnduffy: After discussing offline we'll create the instance of the DependencyTrackingTelemetryModule which, when compiled for .NET Core, will create the DependencyCollectorDiagnosticListener instance.  Then we will set the excluded domains collection, and then when we call Initialize on the module it can call Initialize (Enable) on the listener and pass the excluded domains list on to it.
I'll expose a config option in the .NET Core SDK for customers to provide this list and they can always lookup the listener in the services list later and add to that list.

daschult: I just pushed out a new commit that removes the Enable() function and uses DependencyTrackingTelemetryModule, just as Sergey requested.

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
```diff
@@ -21,47 +21,33 @@ namespace Microsoft.ApplicationInsights.DependencyCollector
...
-                configuration = TelemetryConfiguration.Active;
-            }
-
-            return DiagnosticListener.AllListeners.Subscribe(new DependencyCollectorDiagnosticListener(configuration));
-        }
-
         private readonly ApplicationInsightsUrlFilter applicationInsightsUrlFilter;
         private readonly TelemetryClient client;
+        private readonly bool setComponentCorrelationHttpHeaders;
+        private readonly IEnumerable<string> correlationDomainExclusionList;
```
dnduffy: There needs to be a public property where people can access this list to add to it if they wish.

daschult: Based on Sergey's feedback, developers won't use this class anymore. They will interact directly with DependencyTrackingTelemetryModule instead, which can already do what you've requested.

daschult: This class is setup up in the background by the DependencyTrackingTelemetryModule class when the user calls Initialize().

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
```diff
@@ -21,47 +21,33 @@ namespace Microsoft.ApplicationInsights.DependencyCollector
     /// </summary>
     public class DependencyCollectorDiagnosticListener : IObserver<DiagnosticListener>
```
SergeyKanzhelev: Can this be marked internal now? There is no `Enable` on this class any longer

SergeyKanzhelev:
@daschult now the branch has conflicts. Please resolve

daschult:
@SergeyKanzhelev  I just resolved the merge conflicts.

PR #437, opened 2017-04-11
differenciate exceptions examples

SergeyKanzhelev:


Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -389,6 +389,9 @@ private void SendException(string operationName, string problemId, ExceptionTele
                     exceptionTelemetry.ProblemId = problemId;
                 }
 
+                // this property allows to differenciate examples from regular exceptions tracked using TrackException
```
pharring: typo: "differentiate"

Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -389,6 +389,9 @@ private void SendException(string operationName, string problemId, ExceptionTele
...
+                exceptionTelemetry.Properties.Add("_MS.ProcessedByMetricExtractors", "(Name: Exceptions, Ver: 1.0)");
```
Dmitry-Matveev: What's the reason behind this format, does this format simplify parsing on UI?
Is "_MS.ProcessedByMetricExtractors" used somewhere else as well, so it's a standard name? It adds ~20 bytes to each metrics due to "ByMetricExtractor", we can shorten it if not standard.

SergeyKanzhelev: It is the same as used by metrics aggregator from Greg. It is design to:
1. Be human readable
2. have a prefix indicating Microsoft ownership
3. tell the name and version of the extractor

The plan is to include it into the context as a well-known field eventually.

Do you have a better name in mind?

PR #438, opened 2017-04-11
Correlation implementation in DependencyCollector for netcoreapp 2.0

lmolkova:
Implements correlation for the Http dependencies and consumes new Activity events on netcoreapp2.0


Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
```diff
@@ -26,11 +28,19 @@ internal class DependencyCollectorDiagnosticListener : IObserver<DiagnosticListe
...
+        private readonly ConcurrentDictionary<string, IOperationHolder<DependencyTelemetry>> pendingTelemetry = new ConcurrentDictionary<string, IOperationHolder<DependencyTelemetry>>();
 
         internal DependencyCollectorDiagnosticListener(TelemetryConfiguration configuration, bool setComponentCorrelationHttpHeaders = true, IEnumerable<string> correlationDomainExclusionList = null, ICorrelationIdLookupHelper correlationIdLookupHelper = null)
         {
+            configuration.TelemetryInitializers.Add(new OperationCorrelationTelemetryInitializer());
+
             this.client = new TelemetryClient(configuration);
+            if (string.IsNullOrEmpty(this.client.InstrumentationKey))
+            {
+                this.client.InstrumentationKey = configuration.InstrumentationKey;
```
SergeyKanzhelev: no, please do not do it. Customer may override the `configuration.InstrumentationKey` for the `TelemetryConfiguration.Active` later than this module will be initialized

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
```diff
@@ -82,72 +100,186 @@ public void OnError(Exception error)
...
+        /// Handler for Exception event, it is sent when request processing cause an exception (e.g. because of DNS or network issues)
+        /// There will be anoth Stop event sent with null response.
         /// </summary>
-        [DiagnosticName("System.Net.Http.Request")]
-        public void OnRequest(HttpRequestMessage request, Guid loggingRequestId)
+        [DiagnosticName("System.Net.Http.Exception")]
+        public void OnException(Exception exception, HttpRequestMessage request)
         {
-            try
+            this.client.TrackException(exception);
```
SergeyKanzhelev: please do not track the exception from dependency collector. We will be recording them separately

SergeyKanzhelev: You probably need to stop operation on exception or add some properties like `WebException.Status` to the `resultCode` of the dependency telemetry object

lmolkova: stop event will be fired anyway.

Would you be able to automagically track all **handled** AND unhandled exceptions in AspNetCore?

SergeyKanzhelev: eventually. For now - it will be great if you'd extract some infroamtion from the exception like whether it was a DNS resolution error and associate this info with the dependency telemetry object

lmolkova: by associate you mean that dependency.ResultCode would have information about the exception?

The problem here is that HttpClient on .NET Desktop used to throw WebException that has a [WebExceptionStatus](https://msdn.microsoft.com/en-us/library/system.net.webexceptionstatus(v=vs.110).aspx)

On Core, there is no such thing. Instead there is WinHttpException (on windows) that does not have any status. There is only  a message like "The server name or address could not be resolved".
Do you think it's a good idea to have it in the resultCode and potentially group by it?

BTW, there is an association by root id (with correlation).

SergeyKanzhelev: Oh, you saying there is no "status"? Is there at least limited number of strings? May they be localized?

I'd say - add this string as a custom property for the dependency for now please.

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
```diff
@@ -126,7 +124,7 @@ public void OnActivityStart(HttpRequestMessage request)
             if (request != null &&
                 !this.applicationInsightsUrlFilter.IsApplicationInsightsUrl(request.RequestUri.ToString()))
             {
-                this.InjectRequestHeaders(request, this.client.Context.InstrumentationKey);
+                this.InjectRequestHeaders(request, this.configuration.InstrumentationKey);
```
SergeyKanzhelev: so this will not support the `roleName` integration that @nizarq was working on? I'd suggest to use the same approach he used - take instrumentation key from the `DependencyTelemetry` object. It is already initializaed so it will have it

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
lmolkova: Right, there is no status. For now I'll add custom property with the error message as you suggested.

I assume this is a [list](https://github.com/dotnet/corefx/blob/bffef76f6af208e2042a2f27bc081ee908bb390b/src/Common/src/Interop/Windows/wininet/Interop.wininet_errors.cs) of possible errors on, but it's internal and linux has different list of errors and string messages so we cannot use it.

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
```diff
@@ -82,129 +100,247 @@ public void OnError(Exception error)
...
-            try
+            Activity currentActivity = Activity.Current;
+            if (currentActivity == null)
+            {
+                DependencyCollectorEventSource.Log.CurrentActivityIsNull();
+                return;
+            }
+
+            this.pendingExceptions.TryAdd(currentActivity.Id, exception);
+            this.client.TrackException(exception);
```
lmolkova: Since automatic exception tracking is not yet implemented in AspNetCore SDK, I still think it makes sense to track exception here for now

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
lmolkova: the idea is to avoid storing DependencyTelemetry in dictionary between start and stop: all necessary information is available in Stop event. DependencyTelemetry is created there.

The only thing that is needed is instrumentationKey and I assume it is static (common for all requests). Role information is also static.

If you prefer to create telemetry in start and initialize it (and use more memory and CPU) just to fill in static properties that never change, I can do it. But I don;t think it's a great idea.

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
```diff
@@ -82,129 +100,247 @@ public void OnError(Exception error)
...
+            }
+        }
 
-                        telemetry.Stop();
-                        this.client.Track(telemetry);
+        private void ParseResponse(HttpResponseMessage response, DependencyTelemetry telemetry)
+        {
+            try
+            {
+                string targetApplicationId = HttpHeadersUtilities.GetRequestContextKeyValue(response.Headers, RequestResponseHeaders.RequestContextCorrleationTargetKey);
```
Dmitry-Matveev: Minor: correlation typo "Corrleation"

Src/DependencyCollector/Shared/Implementation/ClientServerDependencyTracker.cs
```diff
@@ -44,8 +44,6 @@ internal static DependencyTelemetry BeginTracking(TelemetryClient telemetryClien
 
             activity.Stop();
 #else
-            telemetryClient.Initialize(telemetry);
-
             // telemetry is initialized by Base SDK OperationCorrealtionTelemetryInitializer
```
Dmitry-Matveev: Would you need to remove/edit this comment in addition to removing Initialization above?

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
lmolkova: could TrackException raise an exception?

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
```diff
@@ -82,129 +100,247 @@ public void OnError(Exception error)
...
         {
-            try
+            Activity currentActivity = Activity.Current;
+            if (currentActivity == null)
+            {
+                DependencyCollectorEventSource.Log.CurrentActivityIsNull();
+                return;
+            }
+
+            this.pendingExceptions.TryAdd(currentActivity.Id, exception);
```
lmolkova: it's an interesting question.
In common case exceptions are rare. However if something is completely wrong (e.g. application configuration always calls wrong url that does not exists or times out), there may be a lot of exceptions. 

overflow will happen if there is MaxInt elements in the dict, that means there are max int parallel outgoing http request. I don't believe it is possible.

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
Dmitry-Matveev: I would expect so: it runs all the Telemetry Initializers synchronously. Some exceptions may happen with Track but will be caught, some exceptions may happen in Initializers (most will be caught, I think, except for the customer-implemented Telemetry Initializers).

Src/DependencyCollector/NetCore.Tests/DependencyCollectorDiagnosticListenerTests.Netstandard16.cs
```diff
@@ -17,7 +18,7 @@ namespace Microsoft.ApplicationInsights.DependencyCollector
     /// Unit tests for DependencyCollectorDiagnosticListener.
     /// </summary>
     [TestClass]
-    public class DependencyCollectorDiagnosticListenerTests
+    public partial class DependencyCollectorDiagnosticListenerTests
```
dnduffy: Why partial?

Src/DependencyCollector/NetCore.Tests/DependencyCollectorDiagnosticListenerTests.Netstandard16.cs
```diff
@@ -57,17 +58,32 @@ public void Initialize()
...
                 TelemetryChannel = telemetryChannel,
                 InstrumentationKey = instrumentationKey,
-            },
+            };
+
+            configuration.TelemetryInitializers.Add(new OperationCorrelationTelemetryInitializer());
+            listener = new DependencyCollectorDiagnosticListener(configuration,
             setComponentCorrelationHttpHeaders: true,
             correlationDomainExclusionList: new string[] { "excluded.host.com" },
             correlationIdLookupHelper: mockCorrelationIdLookupHelper);
```
dnduffy: nitpick: indent formatting for constructor function call.

Src/DependencyCollector/NetCore.Tests/DependencyCollectorDiagnosticListenerTests.Netstandard16.cs
lmolkova: there are netcoreapp2.0 specific tests in another file, these tests are still applicable

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
lmolkova: Do you mean I should handle them here?  I think exceptions in initialize is a bug and it's better not to handle them. 
I have never seen code in AI codebase that handles exceptions from Initialize or Track.So why this line of code is so special?

Src/DependencyCollector/Shared/Implementation/ClientServerDependencyTracker.cs
lmolkova: this comment  is for next 2 lines of code
```
            // telemetry is initialized by Base SDK OperationCorrealtionTelemetryInitializer
            // however it does not know about Activity on .NET40 and does not know how to properly generate Ids
            // let's fix it
            telemetry.Id = ApplicationInsightsActivity.GenerateDependencyId(telemetry.Context.Operation.ParentId);
            telemetry.Context.Operation.Id = ApplicationInsightsActivity.GetRootId(telemetry.Id);
```

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
lmolkova: I also discussed it with @nizarq and he thinks that for role correlation, it would be better to use some other source for the role information instead of telemetry properties

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
Dmitry-Matveev: Addressed in the offline email. I meant to d-check for possible stack overflows here (implied from the method name).

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
Dmitry-Matveev: Ok, so let's skip overflow discussion.
On the topic of wrong requests, it may be simple SQL DB outage or permissions issue that will result in huge amounts of failed calls especially in EntityFW. Let's discuss and address as a bug if required.

Src/DependencyCollector/Shared/Implementation/ClientServerDependencyTracker.cs
Dmitry-Matveev: Yes, it's below,  but it looked like the comment implies that the Initialization you removed is still executed.

lmolkova: it is still executed few lines of code above, I just removed second call to Initialize

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
lmolkova: this code is specific to for HTTP dependencies.

Src/DependencyCollector/NetCore/DependencyCollectorDiagnosticListener.cs
lmolkova: the Exception event is only fired for the exceptions happened during Http request processing. So if the exception is fired in TrackException, it will not cause another Exception event from Http DiganoticSource

Src/DependencyCollector/NetCore.Tests/DependencyCollectorDiagnosticListenerTests.Netstandard16.cs
```diff
@@ -317,5 +340,41 @@ public void OnResponseWithFailedResponseEventWithMatchingRequestAndDifferentTarg
             Assert.AreEqual(notFoundResultCode, telemetry.ResultCode);
             Assert.AreEqual(false, telemetry.Success);
         }
+
+        /// <summary>
+        /// Tests that ougoing request has proper context when done in scope of incoming request (incoming request activity).
```
pharring: Typo: "outgoing"

PR #442, opened 2017-04-12
Expose utilities that could be shared

xiaomi7732:
This is to expose the header utilities that these utility functions then can be shared by its dependencies.


SergeyKanzhelev:
Why was it made public? Who is the consumer of this utilities class? There is no need to extend the public surface without serious reason.

PR #453, opened 2017-04-13
Update BaseSdk version to 2.4.0-beta1

cijothomas:


cijothomas:
I'll follow this naming convention (2.4.0-beta1rc1) next release onwards. Base is already built without rc1 name.

PR #459, opened 2017-04-13
DependencyCollector for .net core should support netstandard1.3

lmolkova:
DependencyCollector for .NET Core will be used by AI AspNetCore SDK.
Core SDK will target netstandard1.3 (since AspNetCore itself targets this).

In order to collect dependencies from Http client on netstandard 1.3, DependencyCollector target should be aligned with Core SDK

### DIFF
--- a/Src/Common/CrossComponentCorrelationEventSource.cs
+++ b/Src/Common/CrossComponentCorrelationEventSource.cs
@@ -1,10 +1,10 @@
 ﻿namespace Microsoft.ApplicationInsights.Common
 {
     using System;
-#if NETCORE || NET45
+#if NETSTANDARD || NET45
     using System.Diagnostics.Tracing;
 #endif
-#if NETCORE
+#if NETSTANDARD
     using System.Reflection;
 #endif
     using Extensibility.Implementation.Tracing;
@@ -83,13 +83,11 @@
             string name;
             try
             {
-#if NETCORE
 #if NETSTANDARD1_3
                 name = "Undefined"; // netstandard1.3 does not have API to retrieve Assembly/AppDomain name
-#else // not NETSTANDARD1_3
+#elif NETSTANDARD1_5
                 name = Assembly.GetEntryAssembly().FullName;
-#endif 
-#else // not NETCORE
+#else // not NETSTANDARD
                 name = AppDomain.CurrentDomain.FriendlyName;
 #endif
             }
--- a/Src/Common/CrossComponentCorrelationEventSource.cs
+++ b/Src/Common/CrossComponentCorrelationEventSource.cs
@@ -84,8 +84,12 @@
             try
             {
 #if NETCORE
+#if NETSTANDARD1_3
+                name = "Undefined"; // netstandard1.3 does not have API to retrieve Assembly/AppDomain name
+#else // not NETSTANDARD1_3
                 name = Assembly.GetEntryAssembly().FullName;
-#else
+#endif 
+#else // not NETCORE
                 name = AppDomain.CurrentDomain.FriendlyName;
 #endif
             }
--- a/Src/Common/CrossComponentCorrelationEventSource.cs
+++ b/Src/Common/CrossComponentCorrelationEventSource.cs
@@ -84,7 +84,13 @@
             try
             {
 #if NETSTANDARD1_3
-                name = "Undefined"; // netstandard1.3 does not have API to retrieve Assembly/AppDomain name
+                // Assembly.GetEntryAssembly() is not available on netstandard1.3, but is available in runtime
+                var getEntryAssemblyMethod =
+                    typeof(Assembly).GetMethod("GetEntryAssembly", BindingFlags.Static | BindingFlags.NonPublic) ??
+                    typeof(Assembly).GetMethod("GetEntryAssembly", BindingFlags.Static | BindingFlags.Public);
+
+                var assembly = getEntryAssemblyMethod.Invoke(obj: null, parameters: Array.Empty<object>()) as Assembly;
+                name = assembly.FullName;
 #elif NETSTANDARD1_5
                 name = Assembly.GetEntryAssembly().FullName;
 #else // not NETSTANDARD
--- a/Src/Common/CrossComponentCorrelationEventSource.cs
+++ b/Src/Common/CrossComponentCorrelationEventSource.cs
@@ -86,8 +86,7 @@
 #if NETSTANDARD1_3
                 // Assembly.GetEntryAssembly() is not available on netstandard1.3, but is available in runtime
                 var getEntryAssemblyMethod =
-                    typeof(Assembly).GetMethod("GetEntryAssembly", BindingFlags.Static | BindingFlags.NonPublic) ??
-                    typeof(Assembly).GetMethod("GetEntryAssembly", BindingFlags.Static | BindingFlags.Public);
+                    typeof(Assembly).GetMethod("GetEntryAssembly", BindingFlags.Static | BindingFlags.NonPublic | BindingFlags.Public);
 
                 var assembly = getEntryAssemblyMethod.Invoke(obj: null, parameters: Array.Empty<object>()) as Assembly;
                 name = assembly.FullName;
--- a/Src/DependencyCollector/NetCore/DependencyCollector.NetCore.csproj
+++ b/Src/DependencyCollector/NetCore/DependencyCollector.NetCore.csproj
@@ -5,7 +5,7 @@
     <VersionPrefix>2.4.0-beta1</VersionPrefix>
     <Authors>Microsoft</Authors>
     <TargetFrameworks>netstandard1.3;netstandard1.5</TargetFrameworks>
-    <DefineConstants>$(DefineConstants);NETCORE</DefineConstants>
+    <DefineConstants>$(DefineConstants);NETSTANDARD</DefineConstants>
     <DelaySign>true</DelaySign>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <AssemblyName>Microsoft.AI.DependencyCollector</AssemblyName>
@@ -17,9 +17,6 @@
     <PackageTargetFallback>$(PackageTargetFallback);dnxcore50</PackageTargetFallback>
     <GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>
     <GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' ">
-    <DefineConstants>$(DefineConstants);NETSTANDARD1_3</DefineConstants>    
   </PropertyGroup>
 
   <ItemGroup>
--- a/Src/DependencyCollector/NetCore/DependencyCollector.NetCore.csproj
+++ b/Src/DependencyCollector/NetCore/DependencyCollector.NetCore.csproj
@@ -1,10 +1,11 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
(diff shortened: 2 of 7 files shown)


Src/Common/CrossComponentCorrelationEventSource.cs
```diff
@@ -84,8 +84,12 @@ private string GetApplicationName()
             try
             {
 #if NETCORE
+#if NETSTANDARD1_3
+                name = "Undefined"; // netstandard1.3 does not have API to retrieve Assembly/AppDomain name
```
SergeyKanzhelev: make sure functional tests wouldn't fail with this

Src/DependencyCollector/NetCore/DependencyCollector.NetCore.csproj
```diff
@@ -1,10 +1,11 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <Copyright>Copyright � Microsoft. All Rights Reserved.</Copyright>
+    <Copyright>Copyright © Microsoft. All Rights Reserved.</Copyright>
     <VersionPrefix>2.4.0-beta1</VersionPrefix>
     <Authors>Microsoft</Authors>
-    <TargetFramework>netstandard1.6</TargetFramework>
+    <TargetFrameworks>netstandard1.3;netstandard1.5</TargetFrameworks>
```
dnduffy: Are we going to get two packages as a result?
Why do we need to target 1.3 and 1.5 vs. just 1.6 (or even just 1.5)?

Src/DependencyCollector/Shared/Implementation/DependencyCollectorEventSource.cs
```diff
@@ -267,8 +267,12 @@ private string GetApplicationName()
             try
             {
 #if NETCORE
+#if NETSTANDARD1_3
+                name = "Undefined"; // netstandard1.3 does not have API to retrieve Assembly/AppDomain name
```
dnduffy: "Undefined" does not seem like an acceptable name.  Even in 1.3 there must be something we can call.

SergeyKanzhelev: Maybe just a process name?

lmolkova: there is nothing in netstandard 1.3 that could be used. Not even a process.

You can check it youself: [netstandard1.3 API list](https://github.com/dotnet/standard/blob/master/docs/versions/netstandard1.3_ref.md)

Src/DependencyCollector/NetCore/DependencyCollector.NetCore.csproj
lmolkova: the package will have following targets: 
* net40 (from the DependencyCollector.Net40 project)
* net45 (from the DependencyCollector.Net45 project)
* netstandard1.3 (from this project)
* netstandard1.5 (from this project)

Since AspNetCore runs on netstandard1.3, then AI SDK for Core must also allow to be installed on 1.3. And we need a way to collect dependencies for this combination.

The reason for netstandard1.5 target is only to show proper application name in EventSource.

Src/Common/CrossComponentCorrelationEventSource.cs
lmolkova: func tests run on netcoreapp1.0 that is netstandard1.5.

The only way to hit this path is to have .net46 framework app with aspnetcore and AI for aspnetcore.
I can create a new app for functional tests and test it, but I'm not sure this weird combination worth effort. Also, it seems nothing bad could happen in production because it's undefined, correct?

Src/DependencyCollector/NetCore/DependencyCollector.NetCore.csproj
```diff
@@ -17,6 +18,9 @@
     <GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>
     <GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
   </PropertyGroup>
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' ">
+    <DefineConstants>$(DefineConstants);NETSTANDARD1_3</DefineConstants>
```
pharring: I don't believe this is necessary because you get it 'for free' from the default targets.

Src/DependencyCollector/NetCore/DependencyCollector.NetCore.csproj
```diff
@@ -1,10 +1,11 @@
...
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <Copyright>Copyright � Microsoft. All Rights Reserved.</Copyright>
+    <Copyright>Copyright © Microsoft. All Rights Reserved.</Copyright>
     <VersionPrefix>2.4.0-beta1</VersionPrefix>
     <Authors>Microsoft</Authors>
-    <TargetFramework>netstandard1.6</TargetFramework>
+    <TargetFrameworks>netstandard1.3;netstandard1.5</TargetFrameworks>
+    <DefineConstants>$(DefineConstants);NETCORE</DefineConstants>
```
pharring: This seems like the wrong value to use. This _isn't_ a .Net Core library (is there such a thing?) This is a Netstandard class library. It's a subtle difference, I know. I would stick with the built-in NETSTANDARD1_3 and NETSTANDARD1_5 defines that you get from the SDK targets and use ```#if NETSTANDARD1_3 || NETSTANDARD1_5``` in your code.
If that's too long, then I agree having a single ```NETSTANDARD``` define might be acceptable.

Src/DependencyCollector/NuGet/Package.nuspec
```diff
@@ -49,9 +49,9 @@
...
     <file src="$configuration$\Src\DependencyCollector\Net45\Microsoft.AI.DependencyCollector.xml" target="lib\net45" />
 
-    <!-- For .NET Standard 1.6 projects -->
-    <file src="$configuration$\Src\DependencyCollector\NetCore\netstandard1.6\Microsoft.AI.DependencyCollector.dll" target="lib\netstandard1.6" />
-    <file src="$configuration$\Src\DependencyCollector\NetCore\netstandard1.6\Microsoft.AI.DependencyCollector.pdb" target="lib\netstandard1.6" />
-    <file src="$configuration$\Src\DependencyCollector\NetCore\netstandard1.6\Microsoft.AI.DependencyCollector.xml" target="lib\netstandard1.6" />
+    <!-- For .NET Standard 1.3 projects -->
+    <file src="$configuration$\Src\DependencyCollector\NetCore\netstandard1.3\Microsoft.AI.DependencyCollector.dll" target="lib\netstandard1.3" />
+    <file src="$configuration$\Src\DependencyCollector\NetCore\netstandard1.3\Microsoft.AI.DependencyCollector.pdb" target="lib\netstandard1.3" />
+    <file src="$configuration$\Src\DependencyCollector\NetCore\netstandard1.3\Microsoft.AI.DependencyCollector.xml" target="lib\netstandard1.3" />
```
pharring: So, we're building netstandard 1.5, but not actually including it in the NuGet? Why not?

Src/DependencyCollector/Shared/Implementation/DependencyCollectorEventSource.cs
```diff
@@ -267,8 +267,12 @@ private string GetApplicationName()
             try
             {
 #if NETCORE
```
pharring: Can I suggest:
```C#
#if NETSTANDARD1_3
    name = "Undefined";
#elif NETSTANDARD1_5
    name = Assembly.GetEntryAssembly().FullName;
#else
    name = AppDomain.CurrentDomain.FriendlyName;
#endif
```

Src/DependencyCollector/NuGet/Package.nuspec
lmolkova: oh, that's a mistake, thanks for noticing it!

Src/DependencyCollector/Shared/Implementation/DependencyCollectorEventSource.cs
lmolkova: Argee

Src/DependencyCollector/Shared/Implementation/DependencyCollectorEventSource.cs
lmolkova: Getting assembly name with reflection now.
Even though API is not available on netstandard1.3, the method will be there anyway in runtime with .NET 4.0+ or netcoreapp1.0+

Src/DependencyCollector/NetCore/DependencyCollector.NetCore.csproj
dnduffy: The AI SDK for .NET Core has been targeting 1.6 for the last five months or so.  Either we need to change it to "downgrade" to 1.3 or this attempt to support 1.3 is not relevant.

Now that you've got reflection code to get the assembly name from the runtime is there really still a need for two netstandard packages?

Src/DependencyCollector/Shared/Implementation/DependencyCollectorEventSource.cs
```diff
@@ -266,9 +266,17 @@ private string GetApplicationName()
             string name;
             try
             {
-#if NETCORE
+#if NETSTANDARD1_3
+                // Assembly.GetEntryAssembly() is not available on netstandard1.3, but is available in runtime
+                var getEntryAssemblyMethod =
+                    typeof(Assembly).GetMethod("GetEntryAssembly", BindingFlags.Static | BindingFlags.NonPublic) ??
+                    typeof(Assembly).GetMethod("GetEntryAssembly", BindingFlags.Static | BindingFlags.Public);
```
dnduffy: Why not combine the BindingFlags for Public and NonPublic into a single call instead of doing the ?? check?

Src/DependencyCollector/Shared/Implementation/DependencyCollectorEventSource.cs
```diff
@@ -266,9 +266,17 @@ private string GetApplicationName()
             string name;
             try
```
dnduffy: We should probably cache the value of name and not make the reflection calls each time.

dnduffy:
Thank-you for getting a value for 1.3.
Since the AI SDK for .NET Core targets Netstandard1.6 anyway (and has for months) are we going to need to "downgrade" it or is the effort to target 1.3 here not really necessary?

Src/DependencyCollector/Shared/Implementation/DependencyCollectorEventSource.cs
lmolkova: this method is called once in ctor, so the value is cached

Src/DependencyCollector/NetCore/DependencyCollector.NetCore.csproj
lmolkova: We need to downgrade it and support netstandard1.3 and 1.5 in the same way AspNetCore does.

We need 2 packages to have netstandard1.3 target and make it possible to use AI AspNetCore SDK on NET4.6

Src/DependencyCollector/Shared/Implementation/DependencyCollectorEventSource.cs
lmolkova: fixed

lmolkova:
The purpose of this PR was to enable DependencyCollector for AI AspNetCore SDK 3.0 that would depend on AspNEtCore 2.0 and target netstandard1.3 and 1.5.

However today it was decided to keep dependency on AspNetCore 1.1 and keep existing targets (net451 and netstandard 1.6).

This PR is no longer needed, I will submit another PR that will include the DiagnosticSource listener for DependencyCollector NET45: it would allow to have all possible instrumentations in DependencyCollector NET45.

PR #461, opened 2017-04-14
Update DiagnosticSource package to version available on nuget.org

lmolkova:


Src/DependencyCollector/Net45.Tests/DependencyTrackingTelemetryModuleTest.cs
```diff
@@ -63,7 +63,7 @@ public void TestHttpPostRequestsAreCollected()
                 var item = (DependencyTelemetry)sentTelemetry;
                 Assert.AreEqual(url, item.Data);
                 Assert.AreEqual(url.Host, item.Target);
-                Assert.AreEqual("POST " + url.AbsolutePath, item.Name);
```
lmolkova: rolling back change done by @brahmnes in https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/449.

This in NET45 tests and NET45 remains instrumented with EventSource or Profiler only

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
```diff
@@ -0,0 +1,136 @@
+﻿namespace Microsoft.ApplicationInsights.DependencyCollector
```
lmolkova: specific tests for Http Reflection hook implemented in DiagnosticSource and used in DependencyCollector

Src/DependencyCollector/Shared/Implementation/HttpDiagnosticSourceListener.cs
```diff
@@ -100,6 +105,81 @@ protected virtual void Dispose(bool disposing)
                 this.disposed = true;
             }
         }
+
```
lmolkova: Http reflection hook is now implemented in DiagnosticSource package, so the Src/DependencyCollector/Shared/Implementation/HttpHandlerDiagnosticListener.cs is removed.

This class implements efficient property fetching with reflection

Src/DependencyCollector/Shared/Implementation/HttpProcessing.cs
```diff
@@ -217,9 +217,10 @@ internal object OnBegin(object thisObj, bool skipIfNotNew)
                         if (webRequest.Headers[RequestResponseHeaders.CorrelationContextHeader] == null)
                         {
 #if NET45
-                            if (Activity.Current.Baggage.Any())
+                            var currentActivity = Activity.Current;
```
lmolkova: Actual fix is here, but it's important to update DiagnosticSource to the one published to nuget to make sure everything works well together

Src/DependencyCollector/Shared/Implementation/HttpDiagnosticSourceListener.cs
```diff
@@ -47,15 +45,22 @@ public void OnNext(KeyValuePair<string, object> value)
         {
             try
             {
-                dynamic propertyBag = value.Value;
```
brahmnes: Since you are removing use of dynamic, you can try removing the Microsoft.CSharp reference in the csproj. I added that to use dynamic

lmolkova:
@cijothomas this PR now addresses DiagnosticSource update only

Src/DependencyCollector/Shared/Implementation/HttpDiagnosticSourceListener.cs
lmolkova: done

Src/DependencyCollector/Shared/Implementation/HttpDiagnosticSourceListener.cs
```diff
@@ -2,31 +2,29 @@
...
+        private readonly PropertyFetcher requestFetcherResponseEvent;
+        private readonly PropertyFetcher responseFetcher;
         private bool disposed = false;
 
         internal HttpDiagnosticSourceListener(FrameworkHttpProcessing httpProcessing)
         {
             this.httpProcessingFramework = httpProcessing;
             this.subscribeHelper = new HttpDiagnosticSourceSubscriber(this);
+            this.requestFetcherRequestEvent = new PropertyFetcher("Request");
+            this.requestFetcherResponseEvent = new PropertyFetcher("Request");
```
dnduffy: Why is "Response" looking for "Request"?

Src/DependencyCollector/Shared/Implementation/HttpDiagnosticSourceListener.cs
dnduffy: Reflection is slow, why not reference the properties directly through their getters?

Src/DependencyCollector/Shared/Implementation/HttpDiagnosticSourceListener.cs
lmolkova: There are 2 events:
1. For Start request, there is only Request in the payload
2. For Get Response: there are Request and Response in the payload.

In order to make reflection fast, we use the fancy efficient fetcher that only slow for the first time property is fetched from the anonymous object payload.
However those fetchers are different for given 2 payloads.

That's why we have 2 fetchers for 2 events

Src/DependencyCollector/Shared/Implementation/HttpDiagnosticSourceListener.cs
lmolkova: With the PropertyFetcher, reflection is slow only for for the first time you fetcvh property from the same kind on anonymous object.
It creates a delegate that fetches the property and caches it. The next fetching from the same kind of the object is fast.

DiagnosticSource sends anonymous objects and reflection is the only way to get property value.

Before this change is accidentally worked because the code that created anonymous object lived in the same assembly and was accessible. So `dynamic` worked.
It's not the case now when HttpHandlerDiagnosticListener is gone to DiagnosticSource

lmolkova:
@dnduffy I replied to your reflection concerns, please check

Src/DependencyCollector/Shared/Implementation/HttpDiagnosticSourceListener.cs
dnduffy: I understand why there are two, what I was wondering is why you were grabbing the Request property and not the Response property from the response.

Src/DependencyCollector/Shared/Implementation/HttpDiagnosticSourceListener.cs
dnduffy: Reflection is much, much, better than dynamic!
I understand what is happening now, I completely misread it earlier and thought you already had the HttpWebRequest object and were getting properties from it.
This change is a big improvement over dynamic. :)

Src/DependencyCollector/Shared/Implementation/HttpDiagnosticSourceListener.cs
dnduffy: More specifically, it seems odd to me that this line: "(HttpWebResponse)this.responseFetcher.Fetch(value.Value);" is getting an HttpWebResponse object from a property named "Request" instead of a property named "Response".

lmolkova:
```
 case "System.Net.Http.Request":
                    {
                        var request = (HttpWebRequest)this.requestFetcherRequestEvent.Fetch(value.Value);
                        this.httpProcessingFramework.OnRequestSend(request);
                        break;
                    }

                    case "System.Net.Http.Response":
                    {
                        var request = (HttpWebRequest)this.requestFetcherResponseEvent.Fetch(value.Value);
                        var response = (HttpWebResponse)this.responseFetcher.Fetch(value.Value);
                        this.httpProcessingFramework.OnResponseReceive(request, response);
                        break;
                    }
```
* requestFetcher**Request**Event reads `Request` property from the `System.Net.Http.Request` event payload
* requestFetcher**Response**Event reads `Request` property from `System.Net.Http.Response` event payload
* responseFetcher reads `Response` property from the `System.Net.Http.Response` event payload.

If you want to propose better names for the fetchers, I'll be happy to do so :)

PR #462, opened 2017-04-15
Workaround for Activity RootId generation

lmolkova:
When there is no parent request id received from upstream service, Activity generates non random Id that does not work with sampling. See more details in https://github.com/Microsoft/ApplicationInsights-dotnet/issues/538

This change feeds OperationTelemetry.Id to the Activity to avoid it's native generation.

In the long term (for the stable 2.4), Activity generation should be revised: https://github.com/dotnet/corefx/issues/18418


lmolkova:
@cijothomas 
There are a lot of tests implemented for this feature. This change breaks none of them.

I will add some checks that verify Activity.Id generation is not used and OperationTelemetry.Id is used as a root id.

Src/DependencyCollector/Shared/Implementation/ClientServerDependencyTracker.cs
```diff
@@ -32,6 +32,18 @@ internal static DependencyTelemetry BeginTracking(TelemetryClient telemetryClien
             // but every operation must have it's own Activity and Id must be set accordingly
             // basically we repeat TelemetryClientExtensions.StartOperation here
             var activity = new Activity(DependencyActivityName);
+
+            // This is workaround for the issue https://github.com/Microsoft/ApplicationInsights-dotnet/issues/538
+            // if there is no parent Activity, ID Activity generates is not random enough to work well with 
+            // ApplicationInsights sampling algorithm
+            // This code should go away when Activity is fixed: https://github.com/dotnet/corefx/issues/18418
+            if (Activity.Current == null)
```
SergeyKanzhelev: Can we move this logic into `Start` method in Base Application Insights SDK so everybody using `Start` & `Stop` will have sampling working as expected?

lmolkova: It's a temporary workaround. Right way would be to solve it in Activity.

Also, are you talking about StartOperation?

It's not always possible and convenient to  call it because:
- you  don't always want to keep OperationHolder object: like in the ClientServerDependencyTracker
- you don't always want to initialize telemetry immediately: like in incoming request handling - because of delayed operation name calculation.

Also, AI don't always start activity: HttpModule does it and AI just 'patches' that activity before it is started, but does not start it

lmolkova:
@SergeyKanzhelev  do you have any other questions/issues?

Src/DependencyCollector/Shared/Implementation/ClientServerDependencyTracker.cs
SergeyKanzhelev: so when I write code like below - will the resulting operations be sampled correctly or not? When the fix in Activity can make it in DiagnosticsSource package?

```
while (true)
{
  using (var op = client.StartOperation("test"))
  {
      Sleep(10);
  }
}
```

lmolkova: StartOperation in fact uses ([here](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/151509eac5e23fcbd09f3a8989a35d1099703957/src/Core/Managed/Shared/TelemetryClientExtensions.cs#L97) and later [here](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/151509eac5e23fcbd09f3a8989a35d1099703957/src/Core/Managed/Shared/TelemetryClientExtensions.cs#L129)) OperationTelemetry.Id if there was no parent Activity. So, StartOperation is free from the sampling problems.

If the root, top-most Activity was created with StartOperation, everythin is fine, if it was created in  Web SDK or AspNetCore SDK, we make sure parent activity is fine in the SDKs.

If user created Activity manually, sampling may not work well, but it is the same as if user passed custom operationId in the StartOperation or set it manually in the telemetry.

PR #469, opened 2017-04-19
Make MetricManager Private

kstevenham:


Src/WindowsServer/WindowsServer.Shared/MetricManager.cs
```diff
@@ -0,0 +1,1571 @@
+﻿namespace Microsoft.ApplicationInsights.WindowsServer
```
SergeyKanzhelev: please move this class under `Implementation` folder

Src/WindowsServer/WindowsServer.Shared/MetricManager.cs
```diff
@@ -0,0 +1,1571 @@
...
+            {
+                public static class Moniker
+                {
+                    public const string Key = MetricPropertiesNamePrefix + ".ProcessedByMetricExtractors";
+                    public const string ExtractorInfoTemplate = "(Name:'{0}', Ver:'{1}')";      // $"(Name:'{ExtractorName}', Ver:'{ExtractorVersion}')"
+                }
+            }
+        }
+
+        public static class Autocollection
```
SergeyKanzhelev: You do not need this class

SergeyKanzhelev: It's fine to keep. I just realized it's just constants

Src/WindowsServer/WindowsServer.Shared/MetricManager.cs
```diff
@@ -0,0 +1,1571 @@
...
+
+        public const long VerboseFailure = 0x4;
+
+        public const long ErrorFailure = 0x8;
+
+        public const long ReservedUserKeywordBegin = 0x10;
+    }
+
+    [EventSource(Name = "Microsoft-ApplicationInsights-Core")]
+    internal sealed class CoreEventSource : EventSource
```
SergeyKanzhelev: Please rename to avoid naming conflicts

Src/WindowsServer/WindowsServer.Shared/MetricManager.cs
kstevenham: I didn't go through each class that I copied to determine if parts of it could be removed. The Autocollection class is part of the MetricTerms class that I copied. So I'd rather not start down this path of removing what was there. Worst case, the code is unused.

PR #470, opened 2017-04-19
merge dev to master for release 2.4.0-beta1

cijothomas:


Src/Web/Web.Nuget/Package.nuspec
```diff
@@ -25,7 +25,7 @@
         <dependency id="Microsoft.ApplicationInsights" version="[$coresdkversion$]" />
         <dependency id="Microsoft.ApplicationInsights.WindowsServer" version="$version$" />
         <dependency id="Microsoft.AspNet.TelemetryCorrelation" version="1.0.0-preview1-10407-02" />
```
SergeyKanzhelev: @lmolkova NuGet hosts the version `1.0.0-preview1-10414-01` of `TelemetryCorrelation`. Should we update the nuspec or it works all right?

lmolkova: It works since nupec specifies minimum version.
I agree it  would be cleaner to change it, but I assume it's to late for beta1. Isn't it?

I will clean it up later today.

Issue #467, opened 2017-04-19
Feature proposal: enable listening to HttpClient DiagnosticSource on net45


lmolkova:
**Problem:**

In the DependencyCollector 4.5 we have following instrumentation for outgoing Http requests:
- Profiler
- EventSource
- HttpClient on net46 with specific DiagnosticListener

DependenycCollector 4,5 is used in AspNetCore SDK for NET451 target.

Applications that run on AspNetCore are also likely to run on .net core (and AspNetCore 2.0 always brings .net core), even if they target NET Desktop framework. 
So the actual System.Net.Http dll and other .NET packages are in fact .NET core implementation.

Net core HttpClient  could neither be instrumented with profiler nor with event source. 
The only instrumentation it has is [DiagnosticSource](https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/src/System/Net/Http/DiagnosticsHandler.cs)

Listener to this instrumentation is implemented in DependencyCollector.NetCore (that targets netstandard 1.6).

**Solution**:
Let's compile listener for DiagnosticSource into DependencyCollector NET45, so regardless of HttpClient dll used in runtime, DependecyCollector would work.

dnduffy:
Agreed.

lmolkova:
Thanks for the feedback,  I'll submit PR later today

PR #472, opened 2017-04-19
Add Core Http DiagnosticListener to DependencyCollector.NET45

lmolkova:
Motivation is described in #467.

This change 
1. Compiles Http Core DiagnosticListener to DependenycCollector.NET4.5 assembly
2. Previously Microsoft.Extenstions.DiagnosticAdapter package was used to hide subscription details, but it is not available for NET45, so subscription is done with pure DiagnosticSource API
3. Moves listener tests to 4.5 tests, adds a couple core specific tests
4. Unfortunately stylecop was off for NetCore projects and there is a lot of noise in the PR to fix stylecop issues


Src/DependencyCollector/NetCore/DependencyCollector.NetCore.csproj
```diff
@@ -19,12 +19,11 @@
...
+    <Compile Include="..\Shared\Implementation\HttpCoreDiagnosticSourceListener.cs;..\Shared\Implementation\HttpHeadersUtilities.cs;..\Shared\Implementation\PropertyFetcher.cs;..\Shared\DependencyTrackingTelemetryModule.cs;..\Shared\SanitizedHostList.cs;..\Shared\Implementation\DependencyCollectorEventSource.cs;..\Shared\Implementation\ApplicationInsightsUrlFilter.cs;..\Shared\Implementation\RemoteDependencyConstants.cs;..\..\Common\CorrelationIdLookupHelper.cs;..\..\Common\CrossComponentCorrelationEventSource.cs;..\..\Common\ExceptionUtilities.cs;..\..\Common\ICorrelationIdLookupHelper.cs;..\..\Common\RequestResponseHeaders.cs;..\..\Common\HeadersUtilities.cs;..\..\Web\Web.Net45\Implementation\SdkVersionUtils.cs" Exclude="bin\**;obj\**;**\*.csproj;packages\**" />
```
SergeyKanzhelev: it is easier to read when every item split in its own `<Compile` node. Also you can include entire `..\Shared` folder instead of individual files. Consider change

lmolkova: done in #484

lmolkova: I do not want to include whole Shared folder for several reasons
1. It contains code that depend on System.Http -related code (WebRequests)
2. It adds A LOT of unused code into the lib that will just make it bigger

Src/DependencyCollector/NetCore.Tests/DependencyTrackingTelemetryModuleTestNetCore.cs
```diff
@@ -0,0 +1,169 @@
...
+                Assert.AreEqual(RemoteDependencyConstants.HTTP, item.Type, "HttpAny has to be dependency kind as it includes http and azure calls");
+                Assert.IsTrue(
+                    DateTime.UtcNow.Subtract(item.Timestamp.UtcDateTime).TotalMilliseconds <
+                    TimeSpan.FromMinutes(1).TotalMilliseconds,
+                    "timestamp < now");
+                Assert.IsTrue(
+                    item.Timestamp.Subtract(DateTime.UtcNow).TotalMilliseconds >
+                    -TimeSpan.FromMinutes(1).TotalMilliseconds,
+                    "now - 1 min < timestamp");
+                Assert.AreEqual("200", item.ResultCode);
```
Dmitry-Matveev: Minor: Here we assume that bing.com does not fail, should we use local endpoint instead if we'd like to check the status code or should we skip status code validation here? It might've been the same in the other tests of ours already, though... but we can start a good habit here, instead :)

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
```diff
@@ -173,11 +185,11 @@ protected virtual void Dispose(bool disposing)
             {
                 if (disposing)
                 {
-#if !NET40 && !NETCORE
+#if NET45
```
Dmitry-Matveev: Is it equivalent ("if NET45" and "if NET40 && !NETCORE")? `httpDesktopDiagnosticSourceListener` seems to be defined under the second condition but released under the fist condition.

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
```diff
@@ -189,6 +201,17 @@ protected virtual void Dispose(bool disposing)
                     {
                         this.sqlEventListener.Dispose();
                     }
+
+                    if (this.httpDesktopDiagnosticSourceListener != null)
```
Dmitry-Matveev: It looks like it was disposed above unless there is a slight difference in the variable name I didn't spot.

Src/DependencyCollector/NetCore.Tests/DependencyTrackingTelemetryModuleTestNetCore.cs
lmolkova: I agree that we should use a local endpoint. 
Skipping status code validation is not an option since we also may get any kind of network issue.

Such tests become almost functional, however they are still quite small and run fast, so I would also prefer to keep them here, so it easier to detect issue.

There are 2 options
1. Do not fail if case of non 200 status OR any network issues (exception). Add small timeout, so test would not run for default 90 sec http timeout in case of error. Then functionality is still tested and rare issues do not make CI broken, however some implementation errors may be swallowed and since we are creating good pattern, it's not the best solution anyway.

2. Add local server with Anna or Nancy.
3. Change those tests to functional.

So, I would vote for 2nd option. What do you think?

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
lmolkova: since there are only 40, 45 and CORE, those are equivalents, but I will change definition condition to be consistent

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
lmolkova: indeed, nice catch!

Src/DependencyCollector/NetCore.Tests/DependencyTrackingTelemetryModuleTestNetCore.cs
lmolkova: I created #494, we can take it separately

lmolkova:
@cijothomas this is also important for 2.4.0. Would it be merged?

cijothomas:
Yes , this will be merged.. (doing  merges one by one..)

Issue #394, opened 2017-03-31
[Exception Statistics] Report Operation Name as a standard context property, not a dimension


SergeyKanzhelev:
@AlexBulankou is this important for 2.4 release?

AlexBulankou:
I believe we implemented this for 2.4 Beta1. @SergeyKanzhelev  can you please confirm that close?

SergeyKanzhelev:
Nope, we just renamed the property to `operationName`.

SergeyKanzhelev:
fixed in #475

PR #475, opened 2017-04-20
Set operationName on context and emit example exception per thread

kstevenham:
Sets operationName on the context and will emit an example exception for different threads.

Unit tests will be coming.

### DIFF
--- a/Src/WindowsServer/WindowsServer.Shared.Tests/FirstChanceExceptionStatisticsTelemetryModuleTest.cs
+++ b/Src/WindowsServer/WindowsServer.Shared.Tests/FirstChanceExceptionStatisticsTelemetryModuleTest.cs
@@ -6,6 +6,7 @@
     using System.Globalization;
     using System.Runtime.CompilerServices;
     using System.Runtime.ExceptionServices;
+    using System.Threading.Tasks;
     using DataContracts;
     using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.Extensibility;
@@ -17,6 +18,8 @@
     [TestClass]
     public class FirstChanceExceptionStatisticsTelemetryModuleTest : IDisposable
     {
+        internal static readonly string TestOperationName = "FirstChanceExceptionTestOperation";
+
         private TelemetryConfiguration configuration;
         private IList<ITelemetry> items;
 
@@ -141,7 +144,7 @@
             {
                 OnInitialize = (item) =>
                 {
-                    item.Context.Operation.Name = "operationName";
+                    item.Context.Operation.Name = TestOperationName;
                 }
             });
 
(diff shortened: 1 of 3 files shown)


SergeyKanzhelev:
#394

Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -169,7 +163,7 @@ private static bool IsTracked(Exception exception)
         {
             if (exception.Data != null)
             {
-                return exception.Data.Contains(ExceptionIsTracked);
+                return exception.Data.Contains(Thread.CurrentThread.ManagedThreadId);
```
SergeyKanzhelev: I'd suggest to take the thread ID once and pass as a parameter.

Src/WindowsServer/WindowsServer.Shared/Implementation/MetricManager.cs
```diff
@@ -409,7 +409,14 @@ private static MetricTelemetry CreateAggregatedMetricTelemetry(Metric metric, Si
             {
                 foreach (KeyValuePair<string, string> property in metric.Dimensions)
                 {
-                    telemetry.Properties.Add(property);
+                    if (string.Compare(property.Key, "operationName") == 0)
```
SergeyKanzhelev: thinking about it - maybe use `ai.operation,name` as we do here: https://github.com/Microsoft/ApplicationInsights-dotnet/blob/de488729c28b721a758e0024ea2d4f79909ff797/src/Core/Managed/Shared/Extensibility/Implementation/External/ContextTagKeys_types.cs#L174

Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs
```diff
@@ -159,7 +153,7 @@ internal bool WasExceptionTracked(Exception exception)
             if (exception.Data != null && !exception.Data.IsReadOnly)
             {
                 // mark exception as tracked
-                exception.Data[ExceptionIsTracked] = null; // The value is unimportant. It's just a sentinel.
+                exception.Data[Thread.CurrentThread.ManagedThreadId] = null; // The value is unimportant. It's just a sentinel.
```
Dmitry-Matveev: Minor: Is something [like this](https://msdn.microsoft.com/en-us/library/dd293602.aspx) possible in .NET code or maybe present in Managed/Unmanaged mix? If exception can be transferred across threads we may detect it couple times.

Src/WindowsServer/WindowsServer.Shared.Tests/FirstChanceExceptionStatisticsTelemetryModuleTest.cs
```diff
@@ -475,6 +478,69 @@ public void FirstChanceExceptionStatisticsTelemetryModuleDoNotIncrementOnRethrow
...
+                            throw;
+                        }
+                    }
+                    catch (Exception)
+                    {
+                    }
+                }
+            }
+
+            Assert.Equal(30, metrics.Count);
```
SergeyKanzhelev: it will also be nice to check that `ExceptionTelemetry` objects will have a different call stacks

SergeyKanzhelev:
comments non blocking. Feel free to merge

PR #477, opened 2017-04-20
Add System.Diagnostics.StackTrace to DependencyCollector nuspec

lmolkova:


SergeyKanzhelev:
Can you please do the same for Base SDK

PR #481, opened 2017-04-21
Fix for Http Dektop DiagnosticSource instrumentation

### LINKED ISSUE #480: Breaking DiagnosticSource changes for Http on NET 4.6
Heads up on DiagnosticSource [change](https://github.com/dotnet/corefx/commit/28b6bbff681b956fbe1e0a5e2133e99dc280b17b?diff=split) in PR [18502](https://github.com/dotnet/corefx/pull/18502).

This change affects reflection hook that instruments WebRequests on .NET46 with DiganosticSource.
It changes events names and starts Activity/injects headers itself.

The rationale for this change: events sent with this hook should be consistent with HttpClient instrumentation on .NET Core. If we don't do it right now, before it became public, we would have to deal with deprecation hell later and we know it's going to be painful for everyone: event producer and consumers (including AI).

What needs to be fixed in AI **2.4.0** Web SDK:
- event names
- AI should not start Activity for dependency if the hook was called.

lmolkova:
Fixes #480

### DIFF
--- a/Src/DependencyCollector/Shared/Implementation/ClientServerDependencyTracker.cs
+++ b/Src/DependencyCollector/Shared/Implementation/ClientServerDependencyTracker.cs
@@ -28,24 +28,38 @@
             telemetry.Start();
             telemetryClient.Initialize(telemetry);
 #if NET45
+            Activity activity;
+            Activity currentActivity = Activity.Current;
+
+            // On .NET46 without profiler, outgoing requests are instrumented with reflection hook in DiagnosticSource 
+            //// see https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/HttpHandlerDiagnosticListener.cs
+            // it creates an Activity and injects standard 'Request-Id' and 'Correlation-Context' headers itself, so we should not start Activity in this case
+            if (currentActivity != null && currentActivity.OperationName == "System.Net.Http.Desktop.HttpRequestOut")
+            {
+                activity = currentActivity;
+            }
+            else
+            {
+                activity = new Activity(DependencyActivityName);
+
+                // This is workaround for the issue https://github.com/Microsoft/ApplicationInsights-dotnet/issues/538
+                // if there is no parent Activity, ID Activity generates is not random enough to work well with 
+                // ApplicationInsights sampling algorithm
+                // This code should go away when Activity is fixed: https://github.com/dotnet/corefx/issues/18418
+                if (Activity.Current == null)
+                {
+                    activity.SetParentId(telemetry.Id);
+                }
+
+                //// end of workaround
+
+                activity.Start();
+                activity.Stop();
+            }
+
             // telemetry is initialized from current Activity (but not the Id)
             // but every operation must have it's own Activity and Id must be set accordingly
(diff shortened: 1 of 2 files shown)


Src/DependencyCollector/Shared/Implementation/HttpDiagnosticSourceListener.cs
```diff
@@ -47,14 +47,16 @@ public void OnNext(KeyValuePair<string, object> value)
             {
                 switch (value.Key)
                 {
-                    case "System.Net.Http.Request":
+                    case "System.Net.Http.Desktop.HttpRequestOut.Start": 
+                    case "System.Net.Http.Request": // remove in 2.5.0
```
SergeyKanzhelev: please expand the comment with the link to the issue or just write in words why it should be removed.

Src/DependencyCollector/Shared/Implementation/ClientServerDependencyTracker.cs
```diff
@@ -28,40 +28,51 @@ internal static DependencyTelemetry BeginTracking(TelemetryClient telemetryClien
...
+                {
+                    activity.SetParentId(telemetry.Id);
+                }
 
-            //// end of workaround
+                //// end of workaround
 
-            activity.Start();
-            
+                activity.Start();
```
SergeyKanzhelev: why do we need to start and stop it?

lmolkova: In case dependency was tracked with profiler, we need to generate new hierarchical Id for it. We use Activity to generate it.

We can also manually append AI-generated DependencyTelemetry.Id to the Activity.Current.Id (if any).
So the code would look a bit more simple.

``` 
            if (currentActivity != null && currentActivity.OperationName == "System.Net.Http.Desktop.HttpRequestOut")
            {
                telemetry.Id = currentActivity.Id;
            }            
            else 
            {
                var parentId = currentActivity?.Id ?? '|';
                telemetry.Id = parentId  + telemetry.Id + '.';
            }

            if (string.IsNullOrEmpty(telemetry.Context.Operation.Id))
            {
                telemetry.Context.Operation.Id = currentActivity?.RootId ??  telemetry.Id 
            }
```
 I'm fine with doing it this way

lmolkova: After careful consideration I think it's  a bad idea to manually generate Id even if simplifies code: it would make Id grow too fast (Activity will append 1 digit, while telemetry.Id is much longer)
Also, if for some reason Id generation changes in Activity, it may cause grief in AI

SergeyKanzhelev: just put an extensive comment there why you are doing it.

SergeyKanzhelev: Do you think profiler need to use Activity instead of "manual" tracking?

lmolkova: Yes, it need to inject proper Id into Http header

lmolkova:
I provided more detailed comments, please have a lokk and merge if there are no other objections

PR #483, opened 2017-04-21
Removed unused app from functional tests

cijothomas:
Each ununsed app folder is 20MB of unwantes files copied over to test machines. This saves some time in functional test builds.


cijothomas:
This also has a fix to properly run the functional tests when running all the tests (sql,http,aspnetcore) together.

Dmitry-Matveev:
If 4.0 test app folder was unused - does it mean that SDK was not tested with .NET 4.0 apps for a while?
Sounds concerning based on the statistics about the amount of .NET 4.0 users.

cijothomas:
@Dmitry-Matveev yes - it appears 4.0 test app was not tested for a very long time.

cijothomas:
This turned out to be more than just removing dead code. This fixes the issues introduced when functional tests were split into individual classes. (mainly around 'initialize' code)

Dmitry-Matveev:
Let's make sure we fully understand situation with 4.0 before we remove it, please keep it for a while to sort things out - we might want to add the tests back that use that app rather than delete it.

PR #484, opened 2017-04-21
HttpDependenciesParsingTelemetryInitializer in DependenycCollector.NetCore

### LINKED ISSUE #482: Azure Http Dependencies do not have correct type on AspNetCore
HttpDependenciesParsingTelemetryInitializer sets `DependencyTelemetry.Type` for Azure storage:

```
        public const string AzureBlob = "Azure blob";
        public const string AzureTable = "Azure table";
        public const string AzureQueue = "Azure queue";
        public const string AzureDocumentDb = "Azure DocumentDB";
        public const string AzureServiceBus = "Azure Service Bus";
        public const string AzureIotHub = "Azure IoT Hub";
```

It's not compiled into the DependencyCollector.NetCore that breaks the feature on AspNetCore SDK: AppMap will show HTTP for such dependencies.

We need to change DependencyCollector.NetCore **2.4.0** to contain it. AspNetCore will add it to the telemetry initializers list.

lmolkova:
Fixes #482

### DIFF
--- a/Src/DependencyCollector/Shared.Tests/HttpDependenciesParsingTelemetryInitializerTest.cs
+++ b/Src/DependencyCollector/Shared.Tests/HttpDependenciesParsingTelemetryInitializerTest.cs
@@ -1,5 +1,6 @@
 ﻿namespace Microsoft.ApplicationInsights.DependencyCollector
 {
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
     using System;
     using System.Collections.Generic;
     using DataContracts;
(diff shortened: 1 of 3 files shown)


Src/DependencyCollector/Shared.Tests/HttpDependenciesParsingTelemetryInitializerTest.cs
```diff
@@ -1,5 +1,6 @@
 ﻿namespace Microsoft.ApplicationInsights.DependencyCollector
 {
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
```
SergeyKanzhelev: yeah... gotta remove this check for test projects

SergeyKanzhelev:
Created https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/401 to track it on aspnetcore side

PR #496, opened 2017-04-25
Update to 2.4.0-beta2

cijothomas:


lmolkova:
Guys, is there any reason why #472 was not merged? It will be used in AspNetCore SDK and if we miss the 2.4.0 train, it would be quite unfortunate.

cijothomas:
Yes , this will be merged.. (doing  merges one by one..)

lmolkova:
@cijothomas  cool, thanks!

PR #500, opened 2017-04-25
QuickPulse ignores first chance exceptions.

tokaplan:


cijothomas:
Closing this PR. Exceptions are not enabled by default, and we can revisit later on should QP ignore exceptions.

PR #502, opened 2017-04-25
QuickPulse sends nested Metrics along with full telemetry documents.

tokaplan:
Nested metrics were previously left out, only dimensions (properties) were being sent.

### DIFF
--- a/Src/PerformanceCollector/Shared/QuickPulseTelemetryProcessor.cs
+++ b/Src/PerformanceCollector/Shared/QuickPulseTelemetryProcessor.cs
@@ -345,10 +345,9 @@
 
             if (telemetry.Properties != null && telemetry.Properties.Count > 0)
             {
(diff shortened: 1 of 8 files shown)


Src/PerformanceCollector/Shared/QuickPulseTelemetryProcessor.cs
```diff
@@ -364,7 +368,29 @@ private static IEnumerable<string> FlattenMessages(Exception exception)
...
+            return properties?.ToArray();
+        }
+
+        private static KeyValuePair<string, double>[] GetMetrics(ISupportMetrics telemetry)
+        {
+            Dictionary<string, double> metrics = null;
+
+            if (telemetry.Metrics != null && telemetry.Metrics.Count > 0)
+            {
+                metrics = new Dictionary<string, double>(MaxPropertyCount + 1);
```
AlexKlimovMS: Comparer

PR #505, opened 2017-04-26
Run only one fetch app id task at a time for a given instrumenation key.

nizarq:
We do this by storing the task id in a concurrent dictionary. If the task is already added, it is not added again, instead we return immediately.

The motivation to do this is to minimize the load when querying for app id. Once app id is retrieved it is stored in a cache and returned from there. But initially when the cache doesn't contain it, user's requests are going to trigger a fetch. User's often have 500 or more requests per second. This change makes sure we do not create 500 or more tasks  (one for each request) in that case.

### DIFF
--- a/Src/Common/CorrelationIdLookupHelper.cs
+++ b/Src/Common/CorrelationIdLookupHelper.cs
@@ -39,6 +39,8 @@
 
         private ConcurrentDictionary<string, string> knownCorrelationIds = new ConcurrentDictionary<string, string>();
 
+        private ConcurrentDictionary<string, int> FetchTasks = new ConcurrentDictionary<string, int>();
+
         // Stores failed instrumentation keys along with the time we tried to retrieve them.
         private ConcurrentDictionary<string, FailedResult> failingInstrumenationKeys = new ConcurrentDictionary<string, FailedResult>();
 
@@ -119,29 +121,49 @@
                         }
                     }
 
-                    // We wait for <getAppIdTimeout> seconds (which is 0 at this point) to retrieve the appId. If retrieved during that time, we return success setting the correlation id.
-                    // If we are still waiting on the result beyond the timeout - for this particular call we return the failure but queue a task continuation for it to be cached for next time.
-                    Task<string> getAppIdTask = this.provideAppId(instrumentationKey.ToLowerInvariant());           
-                    if (getAppIdTask.Wait(GetAppIdTimeout))
+                    // We only want one task to be there to fetch the ikey. If initial requests come in a bunch, only one of them gets to take responsibility of creating the fetch task. Rest return.
+                    if (this.FetchTasks.TryAdd(instrumentationKey, int.MinValue))
                     {
-                        this.GenerateCorrelationIdAndAddToDictionary(instrumentationKey, getAppIdTask.Result);
-                        correlationId = this.knownCorrelationIds[instrumentationKey];
-                        return true;
+                        try
+                        {
+                            // We wait for <getAppIdTimeout> seconds (which is 0 at this point) to retrieve the appId. If retrieved during that time, we return success setting the correlation id.
+                            // If we are still waiting on the result beyond the timeout - for this particular call we return the failure but queue a task continuation for it to be cached for next time.
+                            Task<string> getAppIdTask = this.provideAppId(instrumentationKey.ToLowerInvariant());
+                            this.FetchTasks[instrumentationKey] = getAppIdTask.Id;
+
+                            if (getAppIdTask.Wait(GetAppIdTimeout))
+                            {
+                                this.GenerateCorrelationIdAndAddToDictionary(instrumentationKey, getAppIdTask.Result);
+                                correlationId = this.knownCorrelationIds[instrumentationKey];
+                                return true;
+                            }
+                            else
+                            {
+                                getAppIdTask.ContinueWith((appId) =>
+                                {
+                                    try
+                                    {
+                                        this.GenerateCorrelationIdAndAddToDictionary(instrumentationKey, appId.Result);
+                                    }
+                                    catch (Exception ex)
+                                    {
+                                        this.RegisterFailure(instrumentationKey, ex);
+                                    }
+                                });
+
+                                return false;
+                            }
+                        }
+                        finally
+                        {
+                            int taskId;
+                            this.FetchTasks.TryRemove(instrumentationKey, out taskId);
+                        }
                     }
                     else
                     {
-                        getAppIdTask.ContinueWith((appId) =>
-                        {
-                            try
-                            {
-                                this.GenerateCorrelationIdAndAddToDictionary(instrumentationKey, appId.Result);
-                            }
-                            catch (Exception ex)
-                            {
-                                this.RegisterFailure(instrumentationKey, ex);
-                            }
-                        });
-
+                        // Fetch tasks are scheduled - don't queue a task.
+                        correlationId = string.Empty;
                         return false;
                     }
                 }
@@ -227,11 +249,6 @@
                             return reader.ReadToEnd();
                         }
                     }
-                }
-                catch (Exception ex)
-                {
-                    AppMapCorrelationEventSource.Log.FetchAppIdFailed(this.GetExceptionDetailString(ex));
-                    return null;
                 }
                 finally
                 {
--- a/Src/Common/CorrelationIdLookupHelper.cs
+++ b/Src/Common/CorrelationIdLookupHelper.cs
@@ -152,8 +152,7 @@
                                 }
                                 finally
                                 {
-                                    int taskId;
-                                    this.fetchTasks.TryRemove(instrumentationKey, out taskId);
+                                    this.fetchTasks.TryRemove(instrumentationKey, out int taskId);
                                 }
                             });
 
--- a/Src/Common/CorrelationIdLookupHelper.cs
+++ b/Src/Common/CorrelationIdLookupHelper.cs
@@ -2,15 +2,16 @@
 {
     using System;
     using System.Collections.Concurrent;
+    using System.Collections.Generic;
     using System.Globalization;
     using System.IO;
     using System.Net;
-    using System.Threading.Tasks;
-    using Extensibility;
-    using Extensibility.Implementation.Tracing;
 #if NETCORE
     using System.Net.Http;
 #endif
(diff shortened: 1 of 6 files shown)


Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -119,29 +121,49 @@ public bool TryGetXComponentCorrelationId(string instrumentationKey, out string
...
+                    if (this.FetchTasks.TryAdd(instrumentationKey, int.MinValue))
                     {
-                        getAppIdTask.ContinueWith((appId) =>
+                        try
                         {
-                            try
+                            // We wait for <getAppIdTimeout> seconds (which is 0 at this point) to retrieve the appId. If retrieved during that time, we return success setting the correlation id.
+                            // If we are still waiting on the result beyond the timeout - for this particular call we return the failure but queue a task continuation for it to be cached for next time.
+                            Task<string> getAppIdTask = this.provideAppId(instrumentationKey.ToLowerInvariant());
+                            this.FetchTasks[instrumentationKey] = getAppIdTask.Id;
```
SergeyKanzhelev: why do we store some special value here? Is it needed here?

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -119,29 +121,49 @@ public bool TryGetXComponentCorrelationId(string instrumentationKey, out string
...
+
+                            if (getAppIdTask.Wait(GetAppIdTimeout))
                             {
-                                this.GenerateCorrelationIdAndAddToDictionary(instrumentationKey, appId.Result);
+                                this.GenerateCorrelationIdAndAddToDictionary(instrumentationKey, getAppIdTask.Result);
+                                correlationId = this.knownCorrelationIds[instrumentationKey];
```
SergeyKanzhelev: You can take it as  `getAppIdTask.Result`. Why query the dictionary?

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -119,29 +121,49 @@ public bool TryGetXComponentCorrelationId(string instrumentationKey, out string
...
+                                return true;
                             }
-                            catch (Exception ex)
+                            else
                             {
-                                this.RegisterFailure(instrumentationKey, ex);
+                                getAppIdTask.ContinueWith((appId) =>
```
SergeyKanzhelev: can the task be finished by now? There is no synchronization to ensure that you'll add continuation early enough

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -119,29 +121,49 @@ public bool TryGetXComponentCorrelationId(string instrumentationKey, out string
...
-                            try
+                            // We wait for <getAppIdTimeout> seconds (which is 0 at this point) to retrieve the appId. If retrieved during that time, we return success setting the correlation id.
+                            // If we are still waiting on the result beyond the timeout - for this particular call we return the failure but queue a task continuation for it to be cached for next time.
+                            Task<string> getAppIdTask = this.provideAppId(instrumentationKey.ToLowerInvariant());
+                            this.FetchTasks[instrumentationKey] = getAppIdTask.Id;
+
+                            if (getAppIdTask.Wait(GetAppIdTimeout))
                             {
-                                this.GenerateCorrelationIdAndAddToDictionary(instrumentationKey, appId.Result);
+                                this.GenerateCorrelationIdAndAddToDictionary(instrumentationKey, getAppIdTask.Result);
```
SergeyKanzhelev: why this call is not under `try{}catch` as the next one?

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -119,29 +121,49 @@ public bool TryGetXComponentCorrelationId(string instrumentationKey, out string
...
-                    }
-                    else
+                    // We only want one task to be there to fetch the ikey. If initial requests come in a bunch, only one of them gets to take responsibility of creating the fetch task. Rest return.
+                    if (this.FetchTasks.TryAdd(instrumentationKey, int.MinValue))
                     {
-                        getAppIdTask.ContinueWith((appId) =>
+                        try
                         {
-                            try
+                            // We wait for <getAppIdTimeout> seconds (which is 0 at this point) to retrieve the appId. If retrieved during that time, we return success setting the correlation id.
```
SergeyKanzhelev: what the motivation of this complication?

nizarq: I've just forwarded you a thread which is forcing me to come up with a fix.

Src/Common/CorrelationIdLookupHelper.cs
nizarq: Not really, just how I created the mental map of the dictionary. But just the presence of the key is really enough. Removing.

Src/Common/CorrelationIdLookupHelper.cs
nizarq: It is - Under https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/505/files#diff-a69629ac993eebf6313f7860a792cd57L109

Src/Common/CorrelationIdLookupHelper.cs
nizarq: No .Result will give you appID, not correlationID. The method above adds it to the dictionary. I can change the method to return the correlationId and then use that, but this piece of code only runs once for an instrumentation key. Not too worried about performance here.

Src/Common/CorrelationIdLookupHelper.cs
nizarq: That is totally fine. You can schedule a task with status=RanToCompletion and the continuation would run. I just did a quick test to double check as well.

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -119,29 +121,48 @@ public bool TryGetXComponentCorrelationId(string instrumentationKey, out string
...
+
+                                return false;
                             }
-                        });
-
+                        }
+                        finally
+                        {
+                            int taskId;
+                            this.FetchTasks.TryRemove(instrumentationKey, out taskId);
```
SergeyKanzhelev: I do not understand. If timeout is zero - it will never has appId ready and you will schedule continuation as well as removing ikey from the list of fetch tasks. So the very next request will schedule another one.

Shouldn't you do it inside the `ContinueWith` instead?

SergeyKanzhelev:
Please add unit test that will ensure there are only one task will be started

Src/Common/CorrelationIdLookupHelper.cs
nizarq: good point - that is definitely an issue. the finally block should be in the continuation - my bad.

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -122,42 +137,27 @@ public bool TryGetXComponentCorrelationId(string instrumentationKey, out string
...
+                                    this.GenerateCorrelationIdAndAddToDictionary(instrumentationKey, appId.Result);
+                                }
+                                catch (Exception ex)
+                                {
+                                    this.RegisterFailure(instrumentationKey, ex);
+                                }
+                                finally
+                                {
+                                    int taskId;
+                                    this.fetchTasks.TryRemove(instrumentationKey, out taskId);
```
SergeyKanzhelev: You can write `out int taskId` now instead of defining it above

Src/DependencyCollector/Shared.Tests/Implementation/ProfilerHttpProcessingTest.cs
```diff
@@ -938,12 +953,12 @@ private void Initialize(string instrumentationKey)
...
-            var correlationIdLookupHelper = new CorrelationIdLookupHelper((string ikey) =>
+            var correlationIdLookupHelper = new CorrelationIdLookupHelper(new Dictionary<string, string>
             {
-                // Pretend App Id is the same as Ikey
-                var tcs = new TaskCompletionSource<string>();
-                tcs.SetResult(ikey);
-                return tcs.Task;
+                {
+                    instrumentationKey,
+                    "cid-v1:" + instrumentationKey
```
SergeyKanzhelev: Ideally - do `"cid-v1:" + instrumentationKey + "-appid"` so test will validate that code doesn't use instrumentationKey mistakenly

Src/Web/Web.Shared.Net.Tests/RequestTrackingTelemetryModuleTest.cs
```diff
@@ -495,10 +521,10 @@ private string GetActivityRootId(string telemetryId)
             return telemetryId.Substring(1, telemetryId.IndexOf('.') - 1);
         }
 
-        private RequestTrackingTelemetryModule RequestTrackingTelemetryModuleFactory(TelemetryConfiguration config = null)
+        private RequestTrackingTelemetryModule RequestTrackingTelemetryModuleFactory(TelemetryConfiguration config = null, CorrelationIdLookupHelper correlationHelper = null)
```
SergeyKanzhelev: did you expose this override dictionary as a public property on telemetry modules? THis is how people will override it

Src/Common/CorrelationIdLookupHelper.cs
nizarq: resolved.

Src/Web/Web.Shared.Net.Tests/RequestTrackingTelemetryModuleTest.cs
nizarq: No I haven't done that. That's more like a feature request, not critical for build. If you would like to have that exposed, I think we can open a new enhancement issue and track it through that.

What situations would customers want to have this dictionary provided? Some internal testing / PPE thing? I don't think bulk of the customers running on prod care.

SergeyKanzhelev: this will resolve the problem of initial requests. You can have an automation that will specify appID in configuration during the deployment

nizarq: For a service of serious scale - do you really think it matters if the first few calls are missing the x-component correlation? 
If would say - let someone ask for it before we give it to them.

Issue #509, opened 2017-04-27
AppId is not parsed with DiagnosticSource HTTP hook on net46


lmolkova:
### Problem
DiagnosticSoure HTTP hook sends Begin/End request for outgoing HTTP request.

If request is redirected (and automatic redirects are allowed), there could be multiple End requests for each redirect.
DependecyCollector relies on the EventSource EndRequest to determine when request ends.

According to @nizarq and @xiaomi7732 experiments, EnventSource EndRequest happens **before** DiagnosticSource Http hook EndRequest. And event source event does not have response object to parse headers from.
 
**So the dependency is tracked without appId from the response headers.**

**Another problem is that we do not have tests to cover this issue**

### Solution1: do not support redirects
In case there are no redirects it's simple: there is just one End.
In case there is a redirect, we report only first response: it has wrong duration (only first response duration) and probably no appId.

**I wonder, how profiler treats redirects? It should have exactly the same problem**

### Solution2 (probably a workaround)
We have to wait for the last End from the DiagnosticSource Http hook and track dependency there, not relying on EventSource to do it.

Following code may be used to determine if the End is the last one (see [HttpWebRequest](http://referencesource.microsoft.com/#System/net/System/Net/HttpWebRequest.cs) code):

```C#
        static bool IsLastResponse(HttpWebRequest request, HttpWebResponse response, int redirectNumber)
        {
            if (request.AllowAutoRedirect && 
                IsAutoRedirectStatus(response.StatusCode) &&
                redirectNumber < request.MaximumAutomaticRedirections)
            {
                return false;
            }

            return true;
        }

        static bool IsAutoRedirectStatus(HttpStatusCode responseStatusCode)
        {
            return responseStatusCode == HttpStatusCode.Ambiguous ||   // 300
                   responseStatusCode == HttpStatusCode.Moved ||   // 301
                   responseStatusCode == HttpStatusCode.Redirect ||   // 302
                   responseStatusCode == HttpStatusCode.RedirectMethod ||   // 303
                   responseStatusCode == HttpStatusCode.RedirectKeepVerb;   // 307
        }
```

The only problem is that we have to count number of redirects to understand where to stop.

### Solution3 (probably a long term solution)

We can probably do the trick from solution 2 in DiagnosticSource, however there we must not count redirects manually and we have to use private reflection to get current redirect count from HttpWebRequest, which we should clarify with legals first. It would probably take at least a week just to have an agreement.

[...]

Issue #494, opened 2017-04-24
Use local endpoint in unit tests


lmolkova:
Some unit tests make outgoing http call to external endpoints
* DependencyCollector.NetCore.Tests DependencyTrackingTelemetryModuleTestNetCore
* DependencyCollector.Net46.Tests DependencyTrackingTelemetryModuleTestNet46

To avoid test failures on http issues, or external service unavailability, they should call local endpoint.
* On Net46, it's easy to start local server with [Anna](https://github.com/Anna-Rx/Anna)
* On NetCore, Kestrel should be used (there are examples in AspNetCore repo).

There is a value in keeping those as unit tests, since they are small, fast and easy to run.

With mocked localhost server tests must also verify cross-component correlation feature by mocking Request-Context

lmolkova:
both [Anna](https://github.com/Anna-Rx/Anna) and [Nancy](https://github.com/NancyFx/Nancy) do not have signed dlls and all DependencyCollector unit test dlls are signed (because they use internal classes from the SDk).

That means such tests should go to functional tests and dlls do not need to be signed.

SergeyKanzhelev:
We can just use the `HttpListener` like we do to fake the telemetry [channel](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/75373e57dcf8d646c54ee188461c373f2cc98939/Test/PerformanceCollector/FunctionalTests/Helpers/HttpListenerObservable.cs#L24) No need to run a framework for this.

PR #516, opened 2017-04-28
Track dependency instrumented with HttpDesktop DiagnosticSource in Response event (#509)

lmolkova:
See #509 for the details.

Please note:
1. Before this fix, all dependencies are reported, but none of the dependencies instrumented with DiagnosticSource have proper Target with appId.

2. After this fix
  * dependency has proper target if AppId was returned in response header
  * for requests with redirects, only first call is tracked (so the duration and status are wrong). This we should fix in DiagnosticSource. (**Update** [fixed and merged](https://github.com/dotnet/corefx/pull/19195))
  * some dependency calls may be missed if they are done with WebRequest.GetResponse() AND **response is not disposed**. This is fundamental flaw in the reflection hook and there is no know solution for this issue. 
E.g.
```C#
            HttpWebRequest request = WebRequest.CreateHttp("http://localhost:3000");
            HttpWebResponse response = (HttpWebResponse)request.GetResponse();
```
Will never be tracked... Application that does not dispose WebRequests may run short on the available connections, so in most cases uses will dispose them.

```C#
            HttpWebRequest request = WebRequest.CreateHttp("http://localhost:3000");
            using( HttpWebResponse response = (HttpWebResponse)request.GetResponse())
            {
            }
```
Will be tracked.
**This problem does not affect requests done with HttpClient as responses are disposed.** 

------------------------------------
**Considering that some dependencies may not be collected and it does not look fixable, please decide whether this fix should be taken at all.
Otherwise, just keep current implementation (and sacrifice AppId for all dependencies).**

### DIFF
--- a/Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
+++ b/Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Diagnostics;
+    using System.IO;
     using System.Net;
 
     using Microsoft.ApplicationInsights.Channel;
@@ -54,9 +55,13 @@
                 module.ProfileQueryEndpoint = FakeProfileApiEndpoint;
                 module.Initialize(config);
 
-                var url = new Uri("http://bing.com");
+                var url = new Uri("https://www.bing.com/");
                 HttpWebRequest request = WebRequest.CreateHttp(url);
-                request.GetResponse();
+                HttpWebResponse response = (HttpWebResponse)request.GetResponse();
+                using (var reader = new StreamReader(response.GetResponseStream()))
+                {
+                    reader.ReadToEnd();
+                }
 
                 Assert.IsNotNull(sentTelemetry);
                 var item = (DependencyTelemetry)sentTelemetry;
@@ -118,11 +123,16 @@
                 module.ProfileQueryEndpoint = FakeProfileApiEndpoint;
                 module.Initialize(config);
 
-                var url = new Uri("http://bing.com");
+                var url = new Uri("https://www.bing.com/");
                 HttpWebRequest request = WebRequest.CreateHttp(url);
 
                 var parent = new Activity("parent").AddBaggage("k", "v").SetParentId("|guid.").Start();
-                request.GetResponse();
+                HttpWebResponse response = (HttpWebResponse)request.GetResponse();
+                using (var reader = new StreamReader(response.GetResponseStream()))
+                {
+                    reader.ReadToEnd();
+                }
+
                 parent.Stop();
 
                 Assert.IsNotNull(sentTelemetry);
--- a/Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
+++ b/Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
@@ -1,14 +1,16 @@
 ﻿namespace Microsoft.ApplicationInsights.DependencyCollector
 {
     using System;
+    using System.Collections.Generic;
     using System.Diagnostics;
-    using System.IO;
+    using System.Linq;
     using System.Net;
 
-    using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.DataContracts;
     using Microsoft.ApplicationInsights.DependencyCollector.Implementation;
     using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+    using Microsoft.ApplicationInsights.TestFramework;
     using Microsoft.ApplicationInsights.Web.TestFramework;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
 
@@ -20,14 +22,16 @@
     {
         private const string IKey = "F8474271-D231-45B6-8DD4-D344C309AE69";
         private const string FakeProfileApiEndpoint = "http://www.microsoft.com";
+        private StubTelemetryChannel channel;
+        private TelemetryConfiguration config;
+        private List<DependencyTelemetry> sentTelemetry;
 
-        [TestMethod]
-        [Timeout(5000)]
-        public void TestDependencyCollectionNoParentActivity()
+        [TestInitialize]
+        public void Initialize()
         {
-            ITelemetry sentTelemetry = null;
+            this.sentTelemetry = new List<DependencyTelemetry>();
 
-            var channel = new StubTelemetryChannel
+            this.channel = new StubTelemetryChannel
             {
                 OnSend = telemetry =>
                 {
@@ -36,35 +40,38 @@
                     if (depTelemetry != null &&
                         !depTelemetry.Data.StartsWith(FakeProfileApiEndpoint, StringComparison.OrdinalIgnoreCase))
                     {
-                        Assert.IsNull(sentTelemetry);
-                        sentTelemetry = telemetry;
+                        this.sentTelemetry.Add(depTelemetry);
                     }
                 }
             };
 
-            var config = new TelemetryConfiguration
+            this.config = new TelemetryConfiguration
             {
                 InstrumentationKey = IKey,
-                TelemetryChannel = channel
+                TelemetryChannel = this.channel
             };
 
-            config.TelemetryInitializers.Add(new OperationCorrelationTelemetryInitializer());
+            this.config.TelemetryInitializers.Add(new OperationCorrelationTelemetryInitializer());
+            DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated = false;
+        }
 
+        [TestMethod]
+        [Timeout(5000)]
+        public void TestDependencyCollectionDiagnosticSourceNoParentActivity()
+        {
             using (var module = new DependencyTrackingTelemetryModule())
             {
+                module.EnableDiagnosticSourceInstrumentation = true;
                 module.ProfileQueryEndpoint = FakeProfileApiEndpoint;
-                module.Initialize(config);
+                module.Initialize(this.config);
 
                 var url = new Uri("https://www.bing.com/");
                 HttpWebRequest request = WebRequest.CreateHttp(url);
-                HttpWebResponse response = (HttpWebResponse)request.GetResponse();
-                using (var reader = new StreamReader(response.GetResponseStream()))
+                using (request.GetResponse())
                 {
-                    reader.ReadToEnd();
                 }
 
-                Assert.IsNotNull(sentTelemetry);
-                var item = (DependencyTelemetry)sentTelemetry;
+                var item = this.sentTelemetry.Single();
                 Assert.AreEqual(url, item.Data);
                 Assert.AreEqual(url.Host, item.Target);
                 Assert.AreEqual("GET " + url.AbsolutePath, item.Name);
@@ -79,6 +86,9 @@
                     -TimeSpan.FromMinutes(1).TotalMilliseconds, 
                     "now - 1 min < timestamp");
                 Assert.AreEqual("200", item.ResultCode);
+                Assert.AreEqual(
+                    SdkVersionHelper.GetExpectedSdkVersion(typeof(DependencyTrackingTelemetryModule), "rdddsd:"),
+                    item.Context.GetInternalContext().SdkVersion);
 
                 var requestId = item.Id;
                 Assert.AreEqual(requestId, request.Headers["Request-Id"]);
@@ -91,43 +101,59 @@
 
         [TestMethod]
         [Timeout(5000)]
-        public void TestNoDependencyCollectionNoResponseStreamReading()
+        public void TestDependencyCollectionEventSource()
         {
-            ITelemetry sentTelemetry = null;
-
-            var channel = new StubTelemetryChannel
-            {
-                OnSend = telemetry =>
-                {
-                    // The correlation id lookup service also makes http call, just make sure we skip that
-                    DependencyTelemetry depTelemetry = telemetry as DependencyTelemetry;
-                    if (depTelemetry != null &&
-                        !depTelemetry.Data.StartsWith(FakeProfileApiEndpoint, StringComparison.OrdinalIgnoreCase))
-                    {
-                        Assert.IsNull(sentTelemetry);
-                        sentTelemetry = telemetry;
-                    }
-                }
-            };
-
-            var config = new TelemetryConfiguration
-            {
-                InstrumentationKey = IKey,
-                TelemetryChannel = channel
-            };
-
-            config.TelemetryInitializers.Add(new OperationCorrelationTelemetryInitializer());
-
             using (var module = new DependencyTrackingTelemetryModule())
             {
                 module.ProfileQueryEndpoint = FakeProfileApiEndpoint;
-                module.Initialize(config);
+                module.Initialize(this.config);
+
+                var url = new Uri("https://www.bing.com/");
+                HttpWebRequest request = WebRequest.CreateHttp(url);
+                using (request.GetResponse())
+                {
+                }
+
+                var item = this.sentTelemetry.Single();
+                Assert.AreEqual(url, item.Data);
+                Assert.AreEqual(url.Host, item.Target);
+                Assert.AreEqual(url.AbsolutePath, item.Name);
+                Assert.IsTrue(item.Duration > TimeSpan.FromMilliseconds(0), "Duration has to be positive");
+                Assert.AreEqual(RemoteDependencyConstants.HTTP, item.Type, "HttpAny has to be dependency kind as it includes http and azure calls");
+                Assert.IsTrue(
+                    DateTime.UtcNow.Subtract(item.Timestamp.UtcDateTime).TotalMilliseconds <
+                    TimeSpan.FromMinutes(1).TotalMilliseconds,
+                    "timestamp < now");
+                Assert.IsTrue(
+                    item.Timestamp.Subtract(DateTime.UtcNow).TotalMilliseconds >
+                    -TimeSpan.FromMinutes(1).TotalMilliseconds,
+                    "now - 1 min < timestamp");
+                Assert.AreEqual("200", item.ResultCode);
+                Assert.AreEqual(
+                    SdkVersionHelper.GetExpectedSdkVersion(typeof(DependencyTrackingTelemetryModule), "rddf:"),
+                    item.Context.GetInternalContext().SdkVersion);
+
+                Assert.IsNull(request.Headers["Request-Id"]);
+                Assert.IsNull(request.Headers["x-ms-request-id"]);
+                Assert.IsNull(request.Headers["Correlation-Context"]);
+            }
+        }
+
+        [TestMethod]
+        [Timeout(5000)]
+        public void TestNoDependencyCollectionDiagnosticSourceNoResponseClose()
+        {
+            using (var module = new DependencyTrackingTelemetryModule())
+            {
+                module.EnableDiagnosticSourceInstrumentation = true;
+                module.ProfileQueryEndpoint = FakeProfileApiEndpoint;
+                module.Initialize(this.config);
 
                 var url = new Uri("https://www.bing.com/");
                 HttpWebRequest request = WebRequest.CreateHttp(url);
                 HttpWebResponse response = (HttpWebResponse)request.GetResponse();
 
-                Assert.IsNull(sentTelemetry);
+                Assert.IsFalse(this.sentTelemetry.Any());
                 var requestId = request.Headers["Request-Id"];
                 var rootId = request.Headers["x-ms-request-root-id"];
                 Assert.IsNotNull(requestId);
@@ -139,52 +165,25 @@
 
         [TestMethod]
         [Timeout(5000)]
-        public void TestDependencyCollectionWithParentActivity()
+        public void TestDependencyCollectionDiagnosticSourceWithParentActivity()
         {
-            ITelemetry sentTelemetry = null;
-
-            var channel = new StubTelemetryChannel
-            {
-                OnSend = telemetry =>
-                {
-                    // The correlation id lookup service also makes http call, just make sure we skip that
-                    DependencyTelemetry depTelemetry = telemetry as DependencyTelemetry;
-                    if (depTelemetry != null &&
-                        !depTelemetry.Data.StartsWith(FakeProfileApiEndpoint, StringComparison.OrdinalIgnoreCase))
-                    {
-                        Assert.IsNull(sentTelemetry);
-                        sentTelemetry = telemetry;
-                    }
-                }
-            };
-
-            var config = new TelemetryConfiguration
-            {
-                InstrumentationKey = IKey,
-                TelemetryChannel = channel
-            };
-
-            config.TelemetryInitializers.Add(new OperationCorrelationTelemetryInitializer());
-
             using (var module = new DependencyTrackingTelemetryModule())
             {
+                module.EnableDiagnosticSourceInstrumentation = true;
                 module.ProfileQueryEndpoint = FakeProfileApiEndpoint;
-                module.Initialize(config);
+                module.Initialize(this.config);
 
                 var url = new Uri("https://www.bing.com/");
                 HttpWebRequest request = WebRequest.CreateHttp(url);
 
                 var parent = new Activity("parent").AddBaggage("k", "v").SetParentId("|guid.").Start();
-                HttpWebResponse response = (HttpWebResponse)request.GetResponse();
-                using (var reader = new StreamReader(response.GetResponseStream()))
+                using (request.GetResponse())
                 {
-                    reader.ReadToEnd();
                 }
 
                 parent.Stop();
 
-                Assert.IsNotNull(sentTelemetry);
-                var item = (DependencyTelemetry)sentTelemetry;
+                var item = this.sentTelemetry.Single();
                 Assert.AreEqual("200", item.ResultCode);
 
                 var requestId = item.Id;
--- a/Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
+++ b/Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
@@ -55,6 +55,12 @@
             DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated = false;
         }
 
+        [TestCleanup]
+        public void Cleanup()
+        {
+            DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated = false;
+        }
+
         [TestMethod]
         [Timeout(5000)]
         public void TestDependencyCollectionDiagnosticSourceNoParentActivity()
@@ -64,6 +70,7 @@
                 module.EnableDiagnosticSourceInstrumentation = true;
                 module.ProfileQueryEndpoint = FakeProfileApiEndpoint;
                 module.Initialize(this.config);
+                Assert.IsTrue(DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated);
 
                 var url = new Uri("https://www.bing.com/");
                 HttpWebRequest request = WebRequest.CreateHttp(url);
@@ -107,6 +114,7 @@
             {
                 module.ProfileQueryEndpoint = FakeProfileApiEndpoint;
                 module.Initialize(this.config);
+                Assert.IsFalse(DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated);
 
                 var url = new Uri("https://www.bing.com/");
                 HttpWebRequest request = WebRequest.CreateHttp(url);
--- a/Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
+++ b/Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
@@ -91,6 +91,54 @@
 
         [TestMethod]
         [Timeout(5000)]
+        public void TestNoDependencyCollectionNoResponseStreamReading()
+        {
+            ITelemetry sentTelemetry = null;
+
+            var channel = new StubTelemetryChannel
+            {
+                OnSend = telemetry =>
+                {
+                    // The correlation id lookup service also makes http call, just make sure we skip that
+                    DependencyTelemetry depTelemetry = telemetry as DependencyTelemetry;
+                    if (depTelemetry != null &&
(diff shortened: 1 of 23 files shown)


Src/DependencyCollector/Shared.Tests/Implementation/FrameworkHttpProcessingTest.cs
```diff
@@ -506,13 +305,13 @@ public void Dispose()
...
 
-        private static void ValidateTelemetryPacketForOnRequestSend(
+/*        private static void ValidateTelemetryPacketForOnRequestSend(
             DependencyTelemetry remoteDependencyTelemetryActual, Uri url, string kind, bool? success, double valueMin, string statusCode)
         {
             Assert.AreEqual("GET " + url.AbsolutePath, remoteDependencyTelemetryActual.Name, true, "Resource name in the sent telemetry is wrong");
             string expectedVersion = SdkVersionHelper.GetExpectedSdkVersion(typeof(DependencyTrackingTelemetryModule), prefix: "rddfd:");
             ValidateTelemetryPacket(remoteDependencyTelemetryActual, url, kind, success, valueMin, statusCode, expectedVersion);
-        }
+        }*/
```
dnduffy: Just delete unneeded code.

Src/DependencyCollector/Shared/Implementation/DesktopDiagnosticSourceHttpProcessing.cs
```diff
@@ -0,0 +1,81 @@
...
+
+    /// <summary>
+    /// Concrete class with all processing logic to generate RDD data from the callbacks received from HttpDesktopDiagnosticSourceListener.
+    /// </summary>
+    internal sealed class DesktopDiagnosticSourceHttpProcessing : HttpProcessing
+    {
+        internal ObjectInstanceBasedOperationHolder TelemetryTable;
+
+        internal DesktopDiagnosticSourceHttpProcessing(TelemetryConfiguration configuration, ObjectInstanceBasedOperationHolder telemetryTupleHolder, bool setCorrelationHeaders, ICollection<string> correlationDomainExclusionList, string appIdEndpoint)
+            : base(configuration, SdkVersionUtils.GetSdkVersion("rdd" + RddSource.FrameworkAndDiagnostic + ":"), null, setCorrelationHeaders, correlationDomainExclusionList, appIdEndpoint)
```
dnduffy: Please check with Sergey if "rdd" should be some other prefix code for uniqueness or not.

Src/DependencyCollector/Shared.Tests/Implementation/FrameworkHttpProcessingTest.cs
lmolkova: nice catch!
forgot to remove it

Src/DependencyCollector/Shared/Implementation/DesktopDiagnosticSourceHttpProcessing.cs
lmolkova: it was already there (`RddSource.FrameworkAndDiagnostic`) for this source

lmolkova:
Dan Taylor suggestion is to have a flag that would let users choose what they want to sacrifice :)

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
```diff
@@ -54,9 +55,13 @@ public void TestDependencyCollectionNoParentActivity()
...
 
-                var url = new Uri("http://bing.com");
+                var url = new Uri("https://www.bing.com/");
                 HttpWebRequest request = WebRequest.CreateHttp(url);
-                request.GetResponse();
+                HttpWebResponse response = (HttpWebResponse)request.GetResponse();
+                using (var reader = new StreamReader(response.GetResponseStream()))
+                {
+                    reader.ReadToEnd();
+                }
```
cijothomas: Please retain the test for 'GetResponse() but no subsequent stream read' to test that we wont collect dependencies in that case, and having a test with comments will be easier for future reference.

Src/DependencyCollector/Shared.Tests/Implementation/DesktopDiagnosticSourceHttpProcessingTests.cs
```diff
@@ -0,0 +1,306 @@
...
+        }
+        #endregion //TestInitiliaze
+
+        /// <summary>
+        /// Validates that OnRequestSend and OnResponseReceive sends valid telemetry.
+        /// </summary>
+        [TestMethod]
+        public void RddTestHttpProcessingFrameworkUpdateTelemetryName()
+        {
+            var stopwatch = Stopwatch.StartNew();
```
cijothomas: start stopwatch right before this line 'this.httpProcessingFramework.OnRequestSend(request);  Don't want to include the time taken (its too small but still!) for WebRequest.Create.

cijothomas: same for all other tests in this class.

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
lmolkova: done

cijothomas:
@dnduffy How important is to get this 2.4.0-beta2 for you to refer in aspnet core?  This change is non-trivial, and dont want to include this in beta2 without routine manual tests. Can this wait for 2.4 stable release ?

@lmolkova - Will need to review this in detail+ check functional tests before merging this in.

lmolkova:
@cijothomas This is for Web SDK, it is quite important for the //build scenarios and stable 2.4.0.

lmolkova:
Update on the issue with not-tracked calls: they will be called if the response is closed/disposed properly.

```C#
            HttpWebRequest request = WebRequest.CreateHttp("http://localhost:3000");
            using (HttpWebResponse response = (HttpWebResponse)request.GetResponse())
            {
                Console.WriteLine(response.StatusCode);
            }
```

If app does not dispose WebRequests properly, it could experience a lot of other issues: e.g. reach connection limit.

So I do not consider this is a critical problem any more even fr WebRequests

lmolkova:
Hey guys, please merge it to develop if possible.
@cijothomas @SergeyKanzhelev

Src/DependencyCollector/Shared.Tests/Implementation/DesktopDiagnosticSourceHttpProcessingTests.cs
```diff
@@ -0,0 +1,307 @@
...
+        #region TestInitialize
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            this.configuration = new TelemetryConfiguration();
+            this.sendItems = new List<ITelemetry>();
+            this.configuration.TelemetryChannel = new StubTelemetryChannel { OnSend = item => this.sendItems.Add(item) };
+            this.configuration.InstrumentationKey = Guid.NewGuid().ToString();
+            this.httpProcessingFramework = new DesktopDiagnosticSourceHttpProcessing(this.configuration, new ObjectInstanceBasedOperationHolder(), /*setCorrelationHeaders*/ true, new List<string>(), RandomAppIdEndpoint);
```
cijothomas: httpProcessingFramework - this name is misleading. Please change to reflect the new class name.

Src/DependencyCollector/Shared/Implementation/DesktopDiagnosticSourceHttpProcessing.cs
cijothomas: From this PR, it looks like DiagnosticSource and Framework cannot be active at the same time. So RddSource.FrameworkAndDiagnostic does not make sense here as it was originally intended to tell us dependency was collected by both Framework and Diagnostic source.

cijothomas:
need to discuss about the version prefix. RddSource.FrameworkAndDiagnostic does not seem logical here, rddd maybe okay, but DependencyCollector in Core already uses that, so we need to distinguish that as well.

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
```diff
@@ -221,9 +221,23 @@ protected virtual void Dispose(bool disposing)
         /// </summary>
         private void InitializeForFrameworkEventSource()
         {
+#if NET45
```
nizarq: What is the difference between #if NET 45 here and #if !NET40 below.

lmolkova:
> need to discuss about the version prefix. RddSource.FrameworkAndDiagnostic does not seem logical here, rddd maybe okay, but DependencyCollector in Core already uses that, so we need to distinguish that as well.

 Core and Desktop will never be active at the same time, but it could be useful to distinguish them anyway. What about 'rdddd' (for rdd desktop diagnostics)? However it seems too similar to 'rddd' used in core.
Please advise what prefix is appropriate. I'm fine with anything.

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
lmolkova: you are right, there is no difference

lmolkova:
> Some comments left. Most importantly - can you add/modify Functional tests for the changes? There are a bunch of Functional tests already for dependency collector , please make use of that to validate things end to end.

There are several difficult aspects of adding/changing functional tests
1. There is no NET46 app and this instrumentation only works for NET46. 
2. In order to test it end-to-end, we need 2 apps: 1st app that tracks dependency, second app that responds with Request-Context headers.

I would prefer another solution described in #494, so we can have local mocked server and track dependencies done to it. It requires a bit of investment to set up localhost. There are other questions where those tests should live: in unit-tests projects or in functional tests. I would argue to keep them as unit tests. 
Anyway, I agree that there should be e2e tests, I agree to add them, however I would take it separately as additional PR, which I can do not earlier than a week from now. For now, these changes were tested manually.

Src/DependencyCollector/Shared/Implementation/DesktopDiagnosticSourceHttpProcessing.cs
```diff
@@ -0,0 +1,81 @@
...
+    using Microsoft.ApplicationInsights.DependencyCollector.Implementation.Operation;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Web.Implementation;
+
+    /// <summary>
+    /// Concrete class with all processing logic to generate RDD data from the callbacks received from HttpDesktopDiagnosticSourceListener.
+    /// </summary>
+    internal sealed class DesktopDiagnosticSourceHttpProcessing : HttpProcessing
+    {
+        internal ObjectInstanceBasedOperationHolder TelemetryTable;
```
nizarq: Why make this internal? For tests? Should we make this a private field and expose an internal property instead? It's best practice to have fields private only - properties give more fine grained control on exposing what you want to expose to the outside world - for example may be you only want to give get permissions not set.

Src/DependencyCollector/Shared.Tests/Implementation/DesktopDiagnosticSourceHttpProcessingTests.cs
cijothomas: This is not changed. DesktopDiagnosticSourceHttpProcessingTests is not testing HttpProcessingFramework, but the variable name is telling otherwise. (its difficult to read code with consuing names :( )

Src/DependencyCollector/Shared.Tests/Implementation/DesktopDiagnosticSourceHttpProcessingTests.cs
```diff
@@ -0,0 +1,307 @@
...
+        public void Cleanup()
+        {
+        }
+        #endregion //TestInitiliaze
+
+        /// <summary>
+        /// Validates that OnRequestSend and OnResponseReceive sends valid telemetry.
+        /// </summary>
+        [TestMethod]
+        public void RddTestHttpProcessingFrameworkUpdateTelemetryName()
```
cijothomas: can you modify all the test names to reflect what they are testing? HttpProcessing vs DesktopDiagnosticsSourceHttpProcessing?

Src/DependencyCollector/Shared/Implementation/DesktopDiagnosticSourceHttpProcessing.cs
lmolkova: copy/paste issue

lmolkova:
Update: redirect issue is [fixed](https://github.com/dotnet/corefx/pull/19195) in DiagnosticSource

Src/DependencyCollector/Shared/Implementation/HttpDesktopDiagnosticSourceListener.cs
```diff
@@ -53,7 +53,7 @@ public void OnNext(KeyValuePair<string, object> value)
                     case "System.Net.Http.Request": 
                     {
                         var request = (HttpWebRequest)this.requestFetcherRequestEvent.Fetch(value.Value);
```
cijothomas: We need to add DependencyCollectorEventSource.Log.Callback received type of logging here.

Src/DependencyCollector/Shared/Implementation/HttpProcessing.cs
```diff
@@ -249,8 +249,7 @@ internal object OnBegin(object thisObj, bool skipIfNotNew)
         /// <param name="exception">The exception object if any.</param>
         /// <param name="thisObj">This object.</param>                
         /// <param name="returnValue">Return value of the function if any.</param>
-        /// <param name="endTracking">Whether this method should end the tracking or not.</param>
-        internal void OnEnd(object exception, object thisObj, object returnValue, bool endTracking)
+        internal void OnEnd(object exception, object thisObj, object returnValue)
```
cijothomas: HttpProcessing class should not log about callbacks. Do that in the corresponding listeners.

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
```diff
@@ -221,9 +221,21 @@ protected virtual void Dispose(bool disposing)
         /// </summary>
         private void InitializeForFrameworkEventSource()
         {
-#if !NET40
-            FrameworkHttpProcessing frameworkHttpProcessing = new FrameworkHttpProcessing(this.telemetryConfiguration, DependencyTableStore.Instance.WebRequestCacheHolder, this.SetComponentCorrelationHttpHeaders, this.ExcludeComponentCorrelationHttpHeadersOnDomains, this.EffectiveProfileQueryEndpoint);
-            this.httpDesktopDiagnosticSourceListener = new HttpDesktopDiagnosticSourceListener(frameworkHttpProcessing);
+#if NET45
+            DesktopDiagnosticSourceHttpProcessing desktopHttpProcessing = new DesktopDiagnosticSourceHttpProcessing(
```
cijothomas: we are now enabling both diagnostic source and framework event source. where is the logic to prevent double collection or only one will be active at a time?

Src/DependencyCollector/Shared/Implementation/DesktopDiagnosticSourceHttpProcessing.cs
```diff
@@ -0,0 +1,81 @@
...
+    using System.Net;
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.DependencyCollector.Implementation.Operation;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Web.Implementation;
+
+    /// <summary>
+    /// Concrete class with all processing logic to generate RDD data from the callbacks received from HttpDesktopDiagnosticSourceListener.
+    /// </summary>
+    internal sealed class DesktopDiagnosticSourceHttpProcessing : HttpProcessing
```
lmolkova: Could you please show me where FrameworkEventSource reports failure cases?
I don't see it here [here](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/0e5ac6e38b2e842d6b132beffbced90dc484ae73/Src/DependencyCollector/Shared/Implementation/FrameworkHttpEventListener.cs)

And in case of HTTP failure means exception because of e.g. DNS or network issue or cancellation.

Src/DependencyCollector/Shared/Implementation/HttpProcessing.cs
lmolkova: Callbacks were not introduced in this PR, moving them goes far beyond this pull request,. This PR a critical issue fix and I doubt we should do more refactoring here

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
lmolkova: it's [here](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/4f7d1518545b8a6e88d7455b49fb0bc0887e0e8d/Src/DependencyCollector/Shared/Implementation/FrameworkHttpProcessing.cs#L44)

In runtime, we do not know what happens first: initialization of Http Desktop DiagnosticSource or initialization of the DependencyTrackingTelemetryModule, so we keep listening to both, but ignoring the EventSource if DiagnosticSource is active.

Src/DependencyCollector/Shared/Implementation/HttpProcessing.cs
lmolkova: I added callback logs into the listener and created an issue to consistently change profiler/httpprocessing logs #537

Src/DependencyCollector/Shared/Implementation/HttpDesktopDiagnosticSourceSubscriber.cs
```diff
@@ -48,6 +48,7 @@ public void OnNext(DiagnosticListener value)
                 if (value.Name == "System.Net.Http.Desktop")
                 {
                     this.sourceSubscription = value.Subscribe(this.parent, (Predicate<string>)null);
+                    DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated = true;
```
cijothomas: Does it need to set this back to false on Dispose?

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
```diff
@@ -74,6 +86,9 @@ public void TestDependencyCollectionNoParentActivity()
                     -TimeSpan.FromMinutes(1).TotalMilliseconds, 
                     "now - 1 min < timestamp");
                 Assert.AreEqual("200", item.ResultCode);
+                Assert.AreEqual(
+                    SdkVersionHelper.GetExpectedSdkVersion(typeof(DependencyTrackingTelemetryModule), "rdddsd:"),
```
cijothomas: This test uses DependencyTrackingTelemetryModule() with the config you created above. The config by default should have EnableDiagnosticSourceInstrumentation as false right? If thats the case, then how did we end up with 'rddsd' as version prefix?

cijothomas: never mind - i didn't see you are setting EnableDiagnosticSourceInstrumentation = true at the beginning.

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
```diff
@@ -35,31 +40,38 @@ public void TestDependencyCollectionNoParentActivity()
...
+            this.config = new TelemetryConfiguration
             {
                 InstrumentationKey = IKey,
-                TelemetryChannel = channel
+                TelemetryChannel = this.channel
             };
 
-            config.TelemetryInitializers.Add(new OperationCorrelationTelemetryInitializer());
+            this.config.TelemetryInitializers.Add(new OperationCorrelationTelemetryInitializer());
+            DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated = false;
```
cijothomas: Why test initialize need to explicitly mark this false?

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
```diff
@@ -35,31 +40,38 @@ public void TestDependencyCollectionNoParentActivity()
                     if (depTelemetry != null &&
                         !depTelemetry.Data.StartsWith(FakeProfileApiEndpoint, StringComparison.OrdinalIgnoreCase))
                     {
```
cijothomas: Why this check "!depTelemetry.Data.StartsWith(FakeProfileApiEndpoint" - DependencyCollector by default will not be tracking dependency calls made to any AI endpoints.

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
```diff
@@ -219,11 +224,26 @@ protected virtual void Dispose(bool disposing)
...
+                DesktopDiagnosticSourceHttpProcessing desktopHttpProcessing = new DesktopDiagnosticSourceHttpProcessing(
+                    this.telemetryConfiguration,
+                    DependencyTableStore.Instance.WebRequestConditionalHolder,
+                    this.SetComponentCorrelationHttpHeaders,
+                    this.ExcludeComponentCorrelationHttpHeadersOnDomains,
+                    this.EffectiveProfileQueryEndpoint);
+                this.httpDesktopDiagnosticSourceListener = new HttpDesktopDiagnosticSourceListener(desktopHttpProcessing);
+            }
+
+            FrameworkHttpProcessing frameworkHttpProcessing = new FrameworkHttpProcessing(
```
cijothomas: if EnableDiagnosticSourceInstrumentation is true, we enable it. So why do we need to initialize Framework any more? Even if you enable it, its doing to ignore all the call backs as it has check 'DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated'. This is not desirable. (correct me if i am missing something."

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
```diff
@@ -255,21 +275,21 @@ private void InitializeForRuntimeInstrumentationOrFramework()
...
                 {
                     // if config is set to disable runtime instrumentation then default to framework event source
-                    this.InitializeForFrameworkEventSource();
+                    this.InitializeForDiagnosticAndFrameworkEventSource();
                     DependencyCollectorEventSource.Log.RemoteDependencyModuleVerbose("Runtime instrumentation is set to disabled. Initialize with framework event source instead.");
                 }
             }
             else
             {
                 // if profiler is not attached then default to framework event source
```
cijothomas: minor - change comment to indicate eventsouce and/or diagnostic listene.

Src/DependencyCollector/Shared/Implementation/ClientServerDependencyTracker.cs
```diff
@@ -165,7 +165,7 @@ internal static void AddTupleForWebDependencies(WebRequest webRequest, Dependenc
 
             Tuple<DependencyTelemetry, bool> telemetryTuple = null;
 
-            if (DependencyTableStore.Instance.IsProfilerActivated || PretendProfilerIsAttached)
+            if (DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated || DependencyTableStore.Instance.IsProfilerActivated || PretendProfilerIsAttached)
             {
```
cijothomas: This does not look right. This means that, even when Profiler is not installed, tuple will be looked up in SQLRequestConditionalHolder - this is incorrect, as without Profiler SQL Dependencies should be stored in SqlRequestCacheHolder.

(More concerning here is  - did any functional tests for sql fail? if not there is a test gap)

Src/DependencyCollector/Shared/Implementation/ClientServerDependencyTracker.cs
```diff
@@ -198,7 +198,7 @@ internal static void AddTupleForSqlDependencies(SqlCommand sqlRequest, Dependenc
             }
 
             var telemetryTuple = new Tuple<DependencyTelemetry, bool>(telemetry, isCustomCreated);
```
cijothomas: same comment as above related to SQL without profiler.

Src/DependencyCollector/Shared/Implementation/HttpDesktopDiagnosticSourceSubscriber.cs
cijothomas: We are counting on HttpDesktopDiagnosticSourceSubscriber OnNext() to set IsDesktopHttpDiagnosticSourceActivated = true. We definitely need unit test to check this.

Also, is there a guarantee that OnNext will be called before any callbacks to Framework event source?? If not, the IsDesktopHttpDiagnosticSourceActivated will be False, and FrameworkEventSource will start collecting dependencies, which might be later collected by DiagnosticSource. (this may be artificial scenario, but i'd like to know if thats even possible)

Src/DependencyCollector/Shared/Implementation/HttpDesktopDiagnosticSourceListener.cs
```diff
@@ -52,8 +53,10 @@ public void OnNext(KeyValuePair<string, object> value)
                     // event was temporarily introduced in DiagnosticSource and removed before stable release
                     case "System.Net.Http.Request": 
                     {
+                        // request is never null
                         var request = (HttpWebRequest)this.requestFetcherRequestEvent.Fetch(value.Value);
-                        this.httpProcessingFramework.OnRequestSend(request);
+                        DependencyCollectorEventSource.Log.BeginCallbackCalled(request.GetHashCode(), value.Key);
```
cijothomas: Please use a different log message. I understand there is a task opened for fixing the calls backs, but having a new callback log message for 3-4 switch cases in DiagnosticSourceListener should be very easy fix.

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
```diff
@@ -219,11 +224,26 @@ protected virtual void Dispose(bool disposing)
...
+            if (this.EnableDiagnosticSourceInstrumentation)
+            {
+                DesktopDiagnosticSourceHttpProcessing desktopHttpProcessing = new DesktopDiagnosticSourceHttpProcessing(
+                    this.telemetryConfiguration,
+                    DependencyTableStore.Instance.WebRequestConditionalHolder,
+                    this.SetComponentCorrelationHttpHeaders,
+                    this.ExcludeComponentCorrelationHttpHeadersOnDomains,
+                    this.EffectiveProfileQueryEndpoint);
+                this.httpDesktopDiagnosticSourceListener = new HttpDesktopDiagnosticSourceListener(desktopHttpProcessing);
+            }
```
lmolkova: Let's add log that is WAS actually activated.

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
lmolkova: IsDesktopHttpDiagnosticSourceActivated is not a runtime constant, it set to true when the DiagnosticListener is subscribed to the particular source. Potentially it could happen AFTER module is initialized. It could also be the case that particular source is not available at all, or fails to initialize.

lmolkova: Moreover, in the next PR we will continue listening to EventSource to get failed dependencies

Src/DependencyCollector/Shared/Implementation/FrameworkHttpProcessing.cs
```diff
@@ -49,6 +41,12 @@ public void OnBeginHttpCallback(long id, string resourceName)
             try
             {
                 DependencyCollectorEventSource.Log.BeginCallbackCalled(id, resourceName);
+                if (DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated)
+                {
+                    // request is handled by Desktop DiagnosticSource Listener
+                    DependencyCollectorEventSource.Log.TrackingAnExistingTelemetryItemVerbose();
```
lmolkova: I've already done that in the next PR #538, there was a special event added there

Src/DependencyCollector/Shared/Implementation/HttpDesktopDiagnosticSourceListener.cs
lmolkova: Why it is important? I pass name of the callback in the parameter

Src/DependencyCollector/Shared/Implementation/HttpDesktopDiagnosticSourceSubscriber.cs
lmolkova: Added checks in DependencyTrackingTelemetryModuleTestNet46.

There is no guarantee that subscription will happen before module initialization. It must happen before the first HTTP outgoing call or will never happen at all.
So if there is a HTTP call, in EventSource we will know for sure if diag source is activated or not

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
lmolkova: because it's a fake endpoint here: microsoft.com

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
lmolkova: to clean up previous tests results. Added the same in Cleanup

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
lmolkova: this test explicitly enables instrumentation

Src/DependencyCollector/Shared/Implementation/ClientServerDependencyTracker.cs
lmolkova: removed this check for SQL deps, IsDesktopHttpDiagnosticSourceActivated is only for http

Src/DependencyCollector/Shared/Implementation/ClientServerDependencyTracker.cs
lmolkova: nothing has failed. may be there is a gap

Src/DependencyCollector/Shared/Implementation/HttpDesktopDiagnosticSourceListener.cs
cijothomas: ok callback name should be sufficient for now.

Issue #514, opened 2017-04-27
DependencyCollector for .NET Core: memory leak on netcoreapp1.1 and prior


lmolkova:
DependencyCollector keeps pending telemetry list.
In BeginRequest event it adds telemetry to the concurrent dictionary
In EndRequest event it removes telemetry from the list.

However on netcoreapp1.1 and prior, EndRequest is not guaranteed. It's not fired for canceled and faulted (DNS, network issues, etc..) requests.

In every long running service this list will always grow. In case of massive network issues service may die because of OOM.

PR #517, opened 2017-04-28
Use ConditionalWeakTable to store dependencies

lmolkova:
See #514.

Note that pendingExceptions are not a problem, since in netcore 2.0 Stop event (where we delete exception from the dictionary) is guaranteed.

### DIFF
--- a/Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs
+++ b/Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs
@@ -8,6 +8,7 @@ namespace Microsoft.ApplicationInsights.DependencyCollector.Implementation
     using System.Linq;
     using System.Net.Http;
     using System.Net.Http.Headers;
+    using System.Runtime.CompilerServices;
     using System.Threading.Tasks;
     using Microsoft.ApplicationInsights.Common;
     using Microsoft.ApplicationInsights.DataContracts;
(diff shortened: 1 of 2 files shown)


Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs
```diff
@@ -36,11 +37,11 @@ internal class HttpCoreDiagnosticSourceListener : IObserver<KeyValuePair<string,
         private readonly PropertyFetcher deprecatedResponseFetcher = new PropertyFetcher("Response");
         private readonly PropertyFetcher deprecatedRequestGuidFetcher = new PropertyFetcher("LoggingRequestId");
         private readonly PropertyFetcher deprecatedResponseGuidFetcher = new PropertyFetcher("LoggingRequestId");
-        
+
         #endregion
 
-        private readonly ConcurrentDictionary<string, IOperationHolder<DependencyTelemetry>> pendingTelemetry =
-            new ConcurrentDictionary<string, IOperationHolder<DependencyTelemetry>>();
+        private readonly ConditionalWeakTable<HttpRequestMessage, IOperationHolder<DependencyTelemetry>> pendingTelemetry =
```
SergeyKanzhelev: can this change be tested somehow?

lmolkova: the same data type is used in profiler to store dependencies for webrequests. ConditionalWeakTable does not have any API to check if it empty. It's possible with some reflection only. 
But why do you think we should test ConditionalWeakTable?

SergeyKanzhelev: I was thinking to call Begin callback many times and check somehow that the objects will be gone after `GC.Collect()`. However I cannot think of easy way to check that those operation holders are disposed. 

BTW, will dispose be called on `IOperationHolder`? Will it send telemetry on dispose?

lmolkova: WE call StopOperation that [disposes](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/src/Core/Managed/Shared/TelemetryClientExtensions.cs#L178) operation holder and it [tracks](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/src/Core/Managed/Shared/Extensibility/Implementation/OperationHolder.cs#L110) operation.

Oh, but it would mean that dependency will be tracked anyway when the HttpRequestMessage (the key) is disposed, which is a good side effect

lmolkova: You are right, we can call begin and then null request and after Gc.Collect we should expect the table to be empty. But my point: what are we trying to test here? ConditionalWeakTable? The fact that HttpRequestMessage is collected at some point? Those are not part of AppInsights.

lmolkova: > Oh, but it would mean that dependency will be tracked anyway when the HttpRequestMessage (the key) is disposed, which is a good side effect

This is not correct, because at the time when HttpRequestMessage will be disposed, there will be no matching activity for the operation and OperationHolder.Disponse will do nothing

Issue #504, opened 2017-04-25
Indicating "tracked component" for first few dependency calls for which app id is not yet available.


nizarq:
For cross component correlation, we add headers containing app id. This indicates the receiving component, that the component that generated the headers is a tracked component. 

When retrieving the app id for any given ikey, the first bunch of dependency calls to a given component, may not have the app ids available just yet (we retrieve app ID in the background but do not block the program execution while we do so). Hence outgoing headers for such dependencies do not contain the Request-Context header. This results in such dependencies being logged simply as HTTP dependencies rather than HTTP **tracked** dependencies. 

We feel there is value in indicating, that a dependency is tracked dependency event when app ID could not be fetched and added in time. This will among other things, make it easy for us to ignore showing dependencies in two places - one with the other server nodes and one under the HTTP node group, where both blocks really represent the same kind of url request and should be aggregated once. As per @soubhagyaDash, showing every dependency in two places, just kills the user confidence in us and gives the impression that our cross component correlation is broken. We intend to ignore the first few dependencies for which app_id could not be retrieved if there is a node in app map that aggregates the data from subsequent calls where app id is known.

Hence, the proposal here is to have a special value in the header, which indicates, that the current dependency call is to a tracked component but appId could not be retrieved yet. 

Here are a couple of proposals: `Request-Context=cid-v1:00000000-0000-0000-0000-000000000000` or `Request-Context=cid-retrieval_in_progress` (or something like that). I prefer the 2nd one. We can then just look for this header and still mark the dependency of type (track dependency).

nizarq:
@SergeyKanzhelev @cijothomas @Dmitry-Matveev

zhencui:
`Request-Context=cid-v1:` is also an option(leave an empty string as App Id). From the caller perspective, using string.IsNullOrEmpty to check the exceptional case is more natural than Guid.TryParse or comparing with a magic string.

nizarq:
I am warming up to this option as well. My only concern being - from a contract perspective it doesn't appear to be very clean. But it is no worse than Note that `Request-Context=cid-v1:00000000-0000-0000-0000-000000000000`.

To me, `cid-v1` stands for correlation id, version 1 and I expect it always have a value of a certain format as designated by `v1`.

Having said that, if others agree to having a null value, I don't think it's a bad option.

xiaomi7732:
[...]

PR #526, opened 2017-04-30
For first few calls when appId is not fetched, trying to get a cross …

nizarq:
…component correlation id - this change puts an empty correlation id in the headers. On the receiving end when the empty headers are detected, we make the dependency of type "HTTP (tracked component)"


nizarq:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/504

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -95,6 +95,17 @@ public CorrelationIdLookupHelper(string endpointAddress)
...
         /// <summary>
+        /// Gets the empty correlation id based on the correlation id format. This is sent as a header to indicate that we are dealing with a tracked component, but we could not fetch the appId just yet.
+        /// </summary>
+        public string EmptyCorrelationId
+        {
+            get
+            {
+                return string.Format(CultureInfo.InvariantCulture, CorrelationIdFormat, string.Empty);
+            }
+        }
```
xiaomi7732: nit: Will this string.Format happen every-time it is read? Although won't be too much gain, I suggest we make this a static read-only field.

Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs
```diff
@@ -300,6 +300,10 @@ private void InjectRequestHeaders(HttpRequestMessage request, string instrumenta
                             {
                                 HttpHeadersUtilities.SetRequestContextKeyValue(requestHeaders, RequestResponseHeaders.RequestContextCorrelationSourceKey, sourceApplicationId);
                             }
+                            else
```
xiaomi7732: There we create a new implication that when TryGetXComponentCorrelationId returns false, it means it is empty appId, right? I am a bit worried about this assumption because there could be other cases. For example, when an invalid configuration leads to request failure. Can we check the header again to making sure it is the case of empty appId being returned?

Src/DependencyCollector/Shared/Implementation/HttpProcessing.cs
```diff
@@ -181,11 +181,14 @@ internal object OnBegin(object thisObj, bool skipIfNotNew)
                         if (!string.IsNullOrEmpty(telemetry.Context.InstrumentationKey)
                             && webRequest.Headers.GetNameValueHeaderValue(RequestResponseHeaders.RequestContextHeader, RequestResponseHeaders.RequestContextCorrelationSourceKey) == null)
                         {
-                            string appId;
-                            if (this.correlationIdLookupHelper.TryGetXComponentCorrelationId(telemetry.Context.InstrumentationKey, out appId))
+                            if (this.correlationIdLookupHelper.TryGetXComponentCorrelationId(telemetry.Context.InstrumentationKey, out string appId))
                             {
                                 webRequest.Headers.SetNameValueHeaderValue(RequestResponseHeaders.RequestContextHeader, RequestResponseHeaders.RequestContextCorrelationSourceKey, appId);
                             }
+                            else
```
xiaomi7732: Same concern for the as the last one.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -290,6 +291,10 @@ public void AddTargetHashForResponseHeader(HttpContext context)
                     {
                         context.Response.Headers.SetNameValueHeaderValue(RequestResponseHeaders.RequestContextHeader, RequestResponseHeaders.RequestContextCorrelationTargetKey, correlationId);
                     }
+                    else
+                    {
```
xiaomi7732: Same here.

Src/Web/Web.Shared.Net.Tests/RequestTrackingTelemetryModuleTest.cs
```diff
@@ -344,6 +344,28 @@ public void OnEndDoesNotAddSourceFieldForRequestForSameComponent()
         }
 
         [TestMethod]
+        public void OnEndDoesNotAddSourceFieldForRequestWithEmptyCorrelationId()
+        {
+            // ARRANGE
```
xiaomi7732: 👍

Src/Common/CorrelationIdLookupHelper.cs
nizarq: Good suggestion. I just made it a constant.

Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs
nizarq: Here are the full set of assumptions: 
- If we hit this code, it means we are inside a tracked component.
- We try to recieve appID - it failed.
- No matter why the above happened we want to let the calling component know that they are calling a tracked component.

I don't really care about distinguishing why we failed to get the appID. I think for all cases, tracked dependency should be fine. When instrumentation key is not correct, regardless of whether we log http or http tracked dependency, we have a problem anyways and it can be found with the same degree of ease / difficulty.

Src/DependencyCollector/Shared/Implementation/HttpProcessing.cs
nizarq: Same as above.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
nizarq: same reply

Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs
xiaomi7732: So, you think it is fine to say whatever reason it fails, we treat it as http track without appId.
Hmm, do we really not care why it fails? 1. I think for #504, we are trying to deal with the special case of appId didn't get in time. In that case, we mitigate the special case by building a contract with UI layer that end user for AI SDK need no worry about it.
But that contract shouldn't include other failures that they might need to pay attention to - either because of a bad configuration or any other reason I don't how what.
That is why I think we are over-handling the special case here. What you think?

nizarq: Talked to @xiaomi7732 in person. One of the examples that makes this concern concrete is when the appId cannot be fetched because the url for resolving the appID is wrong. 

We discussed that we can go either way - that is log the dependency as `HTTP` or `HTTP (tracked dependency)`. The only thing different in either case is the type. In either case the user has to understand what is going on to figure out a fix. While it is possible that, not seeing http tracked dependency will spur them in action to fix the issue. It is also possible that most customers won't even realize that they have a problem because they happily see an http dependency and in fact, seeing http tracked dependency that does not have an appID might make it more obvious to them that they have a problem. 

We feel like it's fine to go either way and decided to keep the change as is in this PR.

xiaomi7732: Yup. I agree with @nizarq that when there's something wrong, it is equally difficult, if not more, to just track it as `HTTP`. I think we agreed to go with it unless we got user feedback / push back.

AlexBulankou:
@nizarq can you please elaborate on how making dependency type name to be "HTTP (tracked component)" addresses the problem of missing correlation id. Since dependency collected will have different target field, it will be aggregated in a separate bucket from tracked component with initialized correlation id. Since data is no aggregated incorrectly (on SDK or in MDM), there's no way to fix it now on UI side. Apologies if I'm missing something.

nizarq:
@AlexBulankou - If the HTTP (tracked component) dependency node with target + cid (appId) exists, we plan to hide the with we plan to hide the node that represents the same with call (target) but does not have the cid value.

The intent is to not show two nodes on the application map for the same kind of call. We are hoping the nodes we will be hiding would only, usually be for the initial few calls and it is fine to not have them on the map.

AlexBulankou:
@nizarq, hiding is inherently hacky. What if one of those hidden calls fails - will you show it them. What if customers are viewing the breakdown of calls in charts by using Analytics query - cannot hide it from there. Also what about Application Map in Visual Studio - will we also duplicate the logic of hiding it there. 
Can you elaborate on "We feel generally you shouldn't see the Http (tracked dependency) group node once we are flighting multi-component app map at 100%" - perhaps we should take it offline to discuss? 
The other idea is that we allow customers to provide their mapping of ikey to appid in the config. For resolved ikeys no call to profile endpoint is necessary.

nizarq:
@AlexBulankou - I think regardless of whether or not we want to hide anything in the UI. This change should be taken, it makes it consistent with other sdk implementations (like asp.net core), where we send http tracked dependency regardless of whether or not we were able to find the appID as long as we are dealing with a tracked component.

Pinging other reviewers as well to get it in time for 2.4 stable - @SergeyKanzhelev @Dmitry-Matveev

Src/DependencyCollector/Shared.Tests/Implementation/FrameworkHttpProcessingTest.cs
```diff
@@ -395,6 +395,50 @@ public void RddTestHttpProcessingFrameworkOnBeginSkipsAddingSourceHeaderPerConfi
...
+            var request = WebRequest.Create(this.testUrl);
+            this.httpProcessingFramework.OnRequestSend(request);
+
+            // Header with empty app Id.
+            Assert.AreEqual(request.Headers[RequestResponseHeaders.RequestContextHeader], "appId=cid-v1:");
+
+            // Wait for the appId fetch to complete
+            while (correlationIdLookupHelper.IsFetchAppInProgress(this.configuration.InstrumentationKey))
+            {
+                Thread.Sleep(10); // wait 10 ms.
```
cijothomas: potentially infinite. use TestTimeout for the Testcase to avoid.

cijothomas:
@nizarq @AlexBulankou  - as per my understanding - this won't be included in 2.4 stable as we only want to merge critical fixes. Please let me know if this is not the case.

nizarq:
@cijothomas - we agreed that we would not include it for the sdk being stabilized for build in order not to shake things up for build. Since we decided to go with a beta sdk before build and that is behind us, there is no reason why this should not be taken for the first release after build. So if you ask me - I would say take it.

Now I'm not sure how strict the bar is for taking changes in 2.4 stable after the last beta. I would defer to @AlexBulankou for a final say.

SergeyKanzhelev:
Very strict. This goes to 2.5-beta1

Src/DependencyCollector/Shared.Tests/Implementation/FrameworkHttpProcessingTest.cs
MS-TimothyMothra: I added [Timeout(5000)]

MS-TimothyMothra:
Closing this PR as per the current app map design.

PR #538, opened 2017-05-04
Handle failed dependencies with EventSource and DiagnosticSource

lmolkova:
**DEPENDS ON #516**, do not merge until that one is merged

Http Desktop DiagnosticListener never reports failed requests (resulted in exception because of network issue, SSL cert issue , timed out,  etc).

However they are reported by FramewrokEventSource.

This change reports such failed requests based on EventSoucrce event.

### DIFF
--- a/Src/DependencyCollector/Shared/Implementation/DependencyCollectorEventSource.cs
+++ b/Src/DependencyCollector/Shared/Implementation/DependencyCollectorEventSource.cs
@@ -278,6 +278,15 @@
             this.WriteEvent(25, this.ApplicationName);
         }
 
+        [Event(
+            26,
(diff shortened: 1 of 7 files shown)


Src/DependencyCollector/Shared/Implementation/DependencyCollectorEventSource.cs
```diff
@@ -278,6 +278,15 @@ public void HttpDesktopDiagnosticSourceListenerIsDeactivated(string appDomainNam
             this.WriteEvent(25, this.ApplicationName);
         }
 
+        [Event(
+            26,
+            Message = "Telemetry for id = '{0}' is tracked with HttpDesktopDiagnosticSourceListener.",
```
cijothomas: :)

cijothomas:
THis can be closed as the substitue is already merged

PR #550, opened 2017-05-05
Properly initialize telemetry when Activity is created in DiagnosticSource Http hook

### LINKED ISSUE #549: Http Dependency has wrong parentId when facilitated by Http Desktop DiagnosticSource hook with not released DiagnosticSource
With the next version of DiagnosticSource (available on myget only yet) dependency tracked by Http**Desktop**DiagnosticSourceListener has wrong ParentId: it is the same as dependency.Id.

As a result, some portal experience is broken:

image

Please decide how critical it is for beta3.

DiagnosticSource with update that causes this will be released for sure by May 10th.

Note that is not a bug in DiagnosticSource, but issue in AppInsgihts (as described in #550)

lmolkova:
Fixes #549
DiagnosticSource source behavior has changed after preview1 released on nuget: it started to fire new events: 
`System.Net.Http.Desktop.HttpRequestOut.Start`
`System.Net.Http.Desktop.HttpRequestOut.Stop`

It also creates `Activity` for outgoing request himself and injects 'standard' `Request-Id` and `Correlation-Context` headers.

1. in case of instrumentation with Http Desktop hook in DiagnosticSource AppInsights does not need to create activity or inject headers. It also need to initialize dependency telemetry from the current activity (telemetry.Id is activity.Id, ParentId is activity,ParentId).

2. In case of profiler or EventSource instrumentation, AppInsights creates activity and initialization is done by Base SDK `OperationCorrelationTelemetryInititlizer`, it initializes all telemetry as a **child** of current activity.

This change properly initializes activity for case1.


lmolkova:
ping 

this is important for 2.4.0 stable

Issue #548, opened 2017-05-05
Failed HTTP requests done with HttpClient are never collected by FrameworkEventSource


lmolkova:
Reproducible with any 2.4 beta and **2.3 stable SDK**

Test:
```C#
    await new HttpClient().GetAsync(new Uri($"http://{Guid.NewGuid()}/"));
```

**Expected** result: failed dependency is collected.
**Actual**: not collected

Dependency is properly collected with WebRequest, sync response

``` C#
    HttpWebRequest request = WebRequest.CreateHttp(new Uri($"http://{Guid.NewGuid()}/"));
    var result = request.BeginGetResponse(null, null);
    using (request.EndGetResponse(result)) {}
```

It happens because with HttpClient End happens before Begin:

```
Event Name                                                                            	Time MSec	Process Name                      	Rest  
Microsoft-ApplicationInsights-Extensibility-DependencyCollector/EndCallbackCalled     	3,821.804	vstest.executionengine.x86 (29372)	ThreadID="5,336" FormattedMessage="End callback called for id = '9223372032586174927' " id="9223372032586174927" appDomainName="UnitTestAdapter: Running test" 
Microsoft-ApplicationInsights-Extensibility-DependencyCollector/EndCallbackWithNoBegin	3,822.229	vstest.executionengine.x86 (29372)	ThreadID="5,336" FormattedMessage="End callback - cannot find start of operation for id = '9223372032586174927' " id="9223372032586174927" appDomainName="UnitTestAdapter: Running test" 
Microsoft-ApplicationInsights-Extensibility-DependencyCollector/BeginCallbackCalled   	3,823.995	vstest.executionengine.x86 (29372)	ThreadID="5,336" FormattedMessage="Begin callback called for id = '9,223,372,032,586,174,927', name= 'http://f5cb0a72-dc17-4f0d-aa57-6e5d6fb75b31/' " id="9,223,372,032,586,174,927" name="http://f5cb0a72-dc17-4f0d-aa57-6e5d6fb75b31/" appDomainName="UnitTestAdapter: Running test" 
```

**There is no issue with HttpDesktopDiagnosticSource from the latest beta3 - all kinds on dependencies are properly collected: WebRequest and HttpClient, failed or not.
So I assume this is not critical for 2.4.0**

cijothomas:
We have this test TestRddForHttpAspx451WithHttpClient(). 

[...]

PR #553, opened 2017-05-09
Reject AI URLs in DiagnosticSource.IsEnabled call

lmolkova:
Before HTTP hook in DiagnosticSource (>= preview2) sends Start event, is calls IsEnabled method and passes request object there.

ApplicationInsgiths needs to reject AI url at this stage to prevent any further notifications regarding this request from DiagnosticSource.

### DIFF
--- a/Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
+++ b/Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
@@ -8,6 +8,7 @@
     using System.Net;
     using System.Net.Http;
 
+    using Microsoft.ApplicationInsights.Common;
     using Microsoft.ApplicationInsights.DataContracts;
     using Microsoft.ApplicationInsights.DependencyCollector.Implementation;
     using Microsoft.ApplicationInsights.Extensibility;
@@ -80,7 +81,7 @@
                 }
 
                 this.ValidateTelemetryForDiagnosticSource(this.sentTelemetry.Single(), url, request, true, "200");
-                Assert.IsNull(request.Headers["Correlation-Context"]);
+                Assert.IsNull(request.Headers[RequestResponseHeaders.CorrelationContextHeader]);
             }
         }
 
@@ -101,28 +102,7 @@
                 {
                 }
 
-                var item = this.sentTelemetry.Single();
-                Assert.AreEqual(url, item.Data);
-                Assert.AreEqual(url.Host, item.Target);
-                Assert.AreEqual(url.AbsolutePath, item.Name);
-                Assert.IsTrue(item.Duration > TimeSpan.FromMilliseconds(0), "Duration has to be positive");
-                Assert.AreEqual(RemoteDependencyConstants.HTTP, item.Type, "HttpAny has to be dependency kind as it includes http and azure calls");
-                Assert.IsTrue(
-                    DateTime.UtcNow.Subtract(item.Timestamp.UtcDateTime).TotalMilliseconds <
-                    TimeSpan.FromMinutes(1).TotalMilliseconds,
-                    "timestamp < now");
-                Assert.IsTrue(
-                    item.Timestamp.Subtract(DateTime.UtcNow).TotalMilliseconds >
-                    -TimeSpan.FromMinutes(1).TotalMilliseconds,
-                    "now - 1 min < timestamp");
-                Assert.AreEqual("200", item.ResultCode);
-                Assert.AreEqual(
-                    SdkVersionHelper.GetExpectedSdkVersion(typeof(DependencyTrackingTelemetryModule), "rddf:"),
-                    item.Context.GetInternalContext().SdkVersion);
-
-                Assert.IsNull(request.Headers["Request-Id"]);
-                Assert.IsNull(request.Headers["x-ms-request-id"]);
-                Assert.IsNull(request.Headers["Correlation-Context"]);
+                this.ValidateTelemetryForEventSource(this.sentTelemetry.Single(), url, request, true, "200");
             }
         }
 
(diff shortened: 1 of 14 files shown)


Src/DependencyCollector/Shared/Implementation/HttpDesktopDiagnosticSourceSubscriber.cs
```diff
@@ -47,7 +52,20 @@ public void OnNext(DiagnosticListener value)
...
-                    this.sourceSubscription = value.Subscribe(this.parent, (Predicate<string>)null);
+                    this.sourceSubscription = value.Subscribe(
+                        this.parent, 
+                        (evnt, r, _) =>
+                        {
+                            if (r != null && evnt == "System.Net.Http.Desktop.HttpRequestOut")
+                            {
+                                // request is never null
+                                var request = (HttpWebRequest)r;
+                                return !this.applicationInsightsUrlFilter.IsApplicationInsightsUrl(request.RequestUri.ToString());
```
SergeyKanzhelev: Since the performance test overrides the profile API and quick pulse endpoint - this change will not help much specifically to performance tests. It will help customers for sure.

I'd suggest an easy fix in `IsApplicationInsightsUrl` - do the check `if (SdkInternalOperationsMonitor.IsEntered())` before making url comparisons. Once you'd implement this fix - call to `ToString()` here will be unnecessary

lmolkova: @cijothomas told that perf tests also override TelemetryChannel.Endpoint so `IsApplicationInsightsUrl` works.

But anyway it seems `SdkInternalOperationsMonitor.IsEntered()` should give better performance than string comparison

Src/DependencyCollector/Shared/Implementation/ApplicationInsightsUrlFilter.cs
```diff
@@ -9,52 +9,33 @@ internal sealed class ApplicationInsightsUrlFilter
...
         private readonly TelemetryConfiguration telemetryConfiguration;
 
         private KeyValuePair<string, string> cachedEndpointLeftPart = new KeyValuePair<string, string>();
 
         public ApplicationInsightsUrlFilter(TelemetryConfiguration telemetryConfiguration)
         {
             this.telemetryConfiguration = telemetryConfiguration;
+
+            // cache EndpointLeftPart ahaed, before first dependency call
+            this.GetEndpointLeftPart();
```
lmolkova: Apparently caching endpoint is an expensive operation that takes ~200 microseconds in release build on my machine.

Src/DependencyCollector/Shared/Implementation/DependencyTableStore.cs
```diff
@@ -6,13 +6,13 @@
 
     internal class DependencyTableStore : IDisposable
     {
+        internal static bool IsDesktopHttpDiagnosticSourceActivated = false;
```
lmolkova: previously it was instance field and request it's value caused expensive creation of DependencyTableStore. It's not needed for diagnostic listener

lmolkova:
I discovered that with very fast async Http requests, End callback may happen while we still process Begin, so dependency will not be tracked.  More details are in #548. It happens with  a few first dependencies because of expensive lazy initialization.
I addressed all the perf issues I've found, but still there is a chance that Begin will finish (and store the dependency) AFTER end.
It should be fixed with DiagnosticSource, since callbacks are synchronous, however performance matters anyway.

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
```diff
@@ -126,6 +128,60 @@ public void TestDependencyCollectionEventSource()
...
+                Assert.AreEqual(url.AbsolutePath, item.Name);
+                Assert.IsTrue(item.Duration > TimeSpan.FromMilliseconds(0), "Duration has to be positive");
+                Assert.AreEqual(RemoteDependencyConstants.HTTP, item.Type, "HttpAny has to be dependency kind as it includes http and azure calls");
+                Assert.IsTrue(
+                    DateTime.UtcNow.Subtract(item.Timestamp.UtcDateTime).TotalMilliseconds <
+                    TimeSpan.FromMinutes(1).TotalMilliseconds,
+                    "timestamp < now");
+                Assert.IsTrue(
+                    item.Timestamp.Subtract(DateTime.UtcNow).TotalMilliseconds >
+                    -TimeSpan.FromMinutes(1).TotalMilliseconds,
```
dnduffy: The test timeout is 5 seconds but we are asserting that the duration is less than one minute?
If the test will fail after 5 seconds shouldn't this be checking for only 5 seconds worth of difference or less?

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
```diff
@@ -245,9 +336,12 @@ private void ValidateTelemetryForDiagnosticSource(DependencyTelemetry item, Uri
                 item.Context.GetInternalContext().SdkVersion);
 
             var requestId = item.Id;
-            Assert.AreEqual(requestId, request.Headers["Request-Id"]);
-            Assert.AreEqual(requestId, request.Headers["x-ms-request-id"]);
             Assert.IsTrue(requestId.StartsWith('|' + item.Context.Operation.Id + '.'));
+            if (request != null)
+            {
+                Assert.AreEqual(requestId, request.Headers["Request-Id"]);
+                Assert.AreEqual(requestId, request.Headers["x-ms-request-id"]);
```
dnduffy: nitpick: Aren't these headers in constants we can reuse?

Src/DependencyCollector/Shared/Implementation/ApplicationInsightsUrlFilter.cs
```diff
@@ -9,52 +9,33 @@ internal sealed class ApplicationInsightsUrlFilter
...
+        internal readonly Uri TelemetryServiceEndpointUri = new Uri(TelemetryServiceEndpoint);
         private readonly TelemetryConfiguration telemetryConfiguration;
 
         private KeyValuePair<string, string> cachedEndpointLeftPart = new KeyValuePair<string, string>();
 
         public ApplicationInsightsUrlFilter(TelemetryConfiguration telemetryConfiguration)
         {
             this.telemetryConfiguration = telemetryConfiguration;
+
+            // cache EndpointLeftPart ahaed, before first dependency call
```
dnduffy: nitpick: what is "ahaed" supposed to be?

Src/DependencyCollector/Shared/Implementation/ApplicationInsightsUrlFilter.cs
dnduffy: nitpick: it reads a little strange to call a getter and ignore the result, especially when the comment says "cache" but the return value is not "cached".

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
```diff
@@ -237,7 +242,7 @@ private void InitializeForDiagnosticAndFrameworkEventSource()
                     this.SetComponentCorrelationHttpHeaders,
                     this.ExcludeComponentCorrelationHttpHeadersOnDomains,
                     this.EffectiveProfileQueryEndpoint);
-                this.httpDesktopDiagnosticSourceListener = new HttpDesktopDiagnosticSourceListener(desktopHttpProcessing);
+                this.httpDesktopDiagnosticSourceListener = new HttpDesktopDiagnosticSourceListener(desktopHttpProcessing, new ApplicationInsightsUrlFilter(this.telemetryConfiguration));
```
Dmitry-Matveev: Minor: Does it mean that the only moment we assign telemetry configuration to Url filter is the initialize of the dependency tracking? Change to the telemetry channel endpoint after that, e.g. in the code will not be part of the Url filter and endpoint set in code might not be validated in the filter. It might've been like that before, though.

lmolkova: this is supported by the AppInsights Url filer: it checks if endpoint was changed and caches new value if needed

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
```diff
@@ -298,5 +303,19 @@ private void InitializeForRuntimeInstrumentationOrFramework()
             }
         }
 #endif
+
+#if !NET40
+        [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
+        private void PrepareActivity()
```
lmolkova: no, there is no check. If Activity static stuff was already initialized, this method costs nothing, considering it's done once per DependencyTrackingTelemetryModule initialization.

PR #555, opened 2017-05-10
Remove dependency from storage if there was no status code (NET 4.5.2)

lmolkova:
Fixing issue introduced in #542

### DIFF
--- a/Src/DependencyCollector/Shared/Implementation/FrameworkHttpProcessing.cs
+++ b/Src/DependencyCollector/Shared/Implementation/FrameworkHttpProcessing.cs
@@ -146,7 +146,11 @@ namespace Microsoft.ApplicationInsights.DependencyCollector.Implementation
                     // We never collected statusCode or success before 2.1.0-beta4
                     // We also had duplicates if runtime is also 4.5.2 (4.6 runtime has no such problem)
                     // So starting with 2.1.0-beta4 we are cutting support for HTTP dependencies in .NET 4.5.2.
-                    this.TelemetryTable.Remove(id);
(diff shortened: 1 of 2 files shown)


Src/DependencyCollector/Shared/Implementation/FrameworkHttpProcessing.cs
```diff
@@ -146,6 +146,7 @@ public void OnEndHttpCallback(long id, bool? success, bool synchronous, int? sta
                     // We never collected statusCode or success before 2.1.0-beta4
                     // We also had duplicates if runtime is also 4.5.2 (4.6 runtime has no such problem)
                     // So starting with 2.1.0-beta4 we are cutting support for HTTP dependencies in .NET 4.5.2.
+                    this.TelemetryTable.Remove(id);
```
Dmitry-Matveev: Any side effects of this in addition to the positive effect of letting Event Source to collect request for us?
E.g. any other way to end up with no result code other than the things you mentioned in the issue description?

lmolkova: Note: it does not allow to collect dependencies, it only allows to remove dependency from the MemoryCache: we are not reporting them here.

There is even a negative effect, that when HttpDesktopDiagSourceListener will start working on .NET 4.5 it will not be able to collect such dependencies at all even though it will have status code. I'll fix that now.

My knowledge of the cases when there is no status code is purely based on the comment above this line

lmolkova:
@cijothomas 
> Please add a unit test.
done

PR #556, opened 2017-05-10
Do not parse Framework EventSource payloads when DiagnosticSource is activated

lmolkova:
Small optimization: if IsDesktopHttpDiagnosticSourceActivated is true, we do not even need to parse event source payloads

### DIFF
--- a/Src/DependencyCollector/Shared/Implementation/FrameworkHttpEventListener.cs
+++ b/Src/DependencyCollector/Shared/Implementation/FrameworkHttpEventListener.cs
@@ -78,13 +78,23 @@ namespace Microsoft.ApplicationInsights.DependencyCollector.Implementation
                 switch (eventData.EventId)
                 {
                     case BeginGetResponseEventId:
-                        this.OnBeginGetResponse(eventData);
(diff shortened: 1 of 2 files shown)


Src/DependencyCollector/Shared/Implementation/FrameworkHttpEventListener.cs
```diff
@@ -78,13 +78,23 @@ protected override void OnEventWritten(EventWrittenEventArgs eventData)
                 switch (eventData.EventId)
                 {
                     case BeginGetResponseEventId:
-                        this.OnBeginGetResponse(eventData);
+                        if (!DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated)
```
SergeyKanzhelev: can you remind me - will this method check that it was actually enabled or just configured to be enabled?

lmolkova: if it is actually enabled (subscribed to DiagSource)

SergeyKanzhelev: thx

Issue #554, opened 2017-05-10
Memory leak with MemoryCache


cijothomas:
After PR #516 , there are multiple readers/writers to the MemoryCache - EventSource processing and DiagnosticListener processing. Its not fully understood, but there seems to be a huge memory leak and dumps show MemoryCache is the cause. The items count in cache never goes above the expected value, but their memory usage keeps increasing forever.

Replacing mem cache with own implementation is fixing this problem. Opening this GH issue to understand the issue better or to see if we can live with custom implementation.

cijothomas:
The issue was with DependencyTableStore creating new MemoryCache as a side effect during every Instance property access. This problem existed for months now, but was aggravated with 516, as it called Instance property for every callback.

PR #557, opened 2017-05-10
Make DependencyTableStore a real singleton

lmolkova:
Fixes beta3 memory issue #554 
image

### DIFF
--- a/Src/DependencyCollector/Shared/Implementation/DependencyTableStore.cs
+++ b/Src/DependencyCollector/Shared/Implementation/DependencyTableStore.cs
@@ -13,6 +13,8 @@
 
         internal bool IsProfilerActivated = false;
         internal bool IsDesktopHttpDiagnosticSourceActivated = false;
+
+        private static readonly object SyncRoot = new object();
         private static DependencyTableStore instance;
 
         private DependencyTableStore() 
@@ -29,8 +31,18 @@
         {
            get 
            {
-              Interlocked.CompareExchange<DependencyTableStore>(ref instance, new DependencyTableStore(), null);
-              return instance;
+               if (instance == null)
+               {
+                   lock (SyncRoot)
+                   {
+                       if (instance == null)
+                       {
+                           instance = new DependencyTableStore();
+                       }
+                   }
+               }
+
+               return instance;
            }
         }
 


Src/DependencyCollector/Shared/Implementation/DependencyTableStore.cs
```diff
@@ -29,8 +31,18 @@ internal static DependencyTableStore Instance
         {
            get 
            {
-              Interlocked.CompareExchange<DependencyTableStore>(ref instance, new DependencyTableStore(), null);
-              return instance;
+               if (instance == null)
+               {
+                   lock (SyncRoot)
+                   {
+                       if (instance == null)
```
dnduffy: I don't recall, does C# suffer from problems with doubled checked locking the way that Java does?
Is there a reason not to just create the instance in the declaration and let the class loader handle it?

dnduffy:
My apologies, the instance property already existed and I didn't need to panic merge this.
Sorry

cijothomas:
This solves memory leak, but introduces a lock. Since this method is called multiple times within tracking of a single dependency, this can be perf overhead. i will slightly tweak this and submit new PR on top of this.

lmolkova:
> This solves memory leak, but introduces a lock. Since this method is called multiple times within tracking of a single dependency, this can be perf overhead. 

I did not get this.

The first time Instance is called in DependencyTrackingTelemetryModule.InitializeForDiagnosticAndFrameworkEventSource it will create a new table. Event if there is some concurrency in creation, there will be no dead lock.

After that Instance will always be there and there will be no lock at  all. This is canonical multi-threaded singleton implementation.

Src/DependencyCollector/Shared/Implementation/DependencyTableStore.cs
lmolkova: C# needs doubled checking, see https://msdn.microsoft.com/en-us/library/ff650316.aspx

I do not know why initially this was implemented as singleton. If you say it's not needed, I'm fine to initialize instance once in crot

cijothomas:
Ignore what i said about lock- lock is not involved except for 1st time.

Src/DependencyCollector/Shared/Implementation/DependencyTableStore.cs
dnduffy: I'm not questioning the singleton, only the doubled checked locking for initialization instead of a static constructor call with the variable declaration.

lmolkova: Double checking is described in that article. Anyway I'm fine with @cijothomas PR #559. We should have it in the stable release

PR #559, opened 2017-05-10
Removed locking from DependencyTableStore

cijothomas:
Removed locking from DependencyTableStore as lock is perf deterrent in this hot path. Instead the instance in static constructed, so its thread safe and wont need locks.

### DIFF
--- a/Src/DependencyCollector/Shared/Implementation/DependencyTableStore.cs
+++ b/Src/DependencyCollector/Shared/Implementation/DependencyTableStore.cs
@@ -14,10 +14,9 @@
         internal bool IsProfilerActivated = false;
         internal bool IsDesktopHttpDiagnosticSourceActivated = false;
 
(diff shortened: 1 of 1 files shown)


cijothomas:
THis wont be needed as i realized the lock is not needed for every call in the PR #557 . So 557 alone is sufficient.

cijothomas:
Wont merge this PR for beta3 as this needs some thoughts..

dnduffy:
"...needs some thoughts."
What thoughts?  These are the exact changes I asked for in the other PR, I think you should just merge it.

cijothomas:
@dnduffy Did not want to include a last minute change into beta3.

cijothomas:
@Dmitry-Matveev @SergeyKanzhelev - Can you guys also take a quick peek to see if this looks okay?

PR #562, opened 2017-05-10
Add comment about know issue in MemoryCache

lmolkova:
MemoryCache has a know race-condition issue when items with the same key are added concurrently.
This does not affect ApplicationInsights scenarios now, but keeping a comment could help to avoid running into this issue later.

### DIFF
--- a/Src/DependencyCollector/Shared/Implementation/Operation/CacheBasedOperationHolder.cs
+++ b/Src/DependencyCollector/Shared/Implementation/Operation/CacheBasedOperationHolder.cs
@@ -41,6 +41,11 @@
             return this.memoryCache.Remove(id.ToString(CultureInfo.InvariantCulture)) != null;
         }
 
+        /// <summary>
+        /// Adds telemetry tuple to MemoryCache. DO NOT call it for the id that already exists in the cache.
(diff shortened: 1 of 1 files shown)


Src/DependencyCollector/Shared/Implementation/Operation/CacheBasedOperationHolder.cs
```diff
@@ -41,6 +41,11 @@ public bool Remove(long id)
             return this.memoryCache.Remove(id.ToString(CultureInfo.InvariantCulture)) != null;
         }
 
+        /// <summary>
+        /// Adds telemetry tuple to MemoryCache. DO NOT call it for the id that already exists in the cache.
+        /// This is a known Memory Cache race-condition issue when items with same id are added concurrently
+        /// and MemoryCache leaks memory. It should be fixed sometime AFTER .NET 4.7.1.
+        /// </summary>
```
dnduffy: Can/should we do some very lightweight spinlock or similar around preventing this situation since we will likely have to support older framework versions for a while?

lmolkova: now we do not have scenarios that write same id concurrently. If we will ever have a need for this, we will need to have some sync/lock logic.

PR #558, opened 2017-05-10
Revert "Do not parse Framework EventSource payloads when DiagnosticSource is activated"

cijothomas:
Reverts Microsoft/ApplicationInsights-dotnet-server#556

PR #563, opened 2017-05-10
Revert "Revert "Do not parse Framework EventSource payloads when DiagnosticSource is activated""

cijothomas:
Reverts Microsoft/ApplicationInsights-dotnet-server#558

### DIFF
--- a/Src/DependencyCollector/Shared/Implementation/FrameworkHttpEventListener.cs
+++ b/Src/DependencyCollector/Shared/Implementation/FrameworkHttpEventListener.cs
@@ -78,13 +78,23 @@ namespace Microsoft.ApplicationInsights.DependencyCollector.Implementation
                 switch (eventData.EventId)
                 {
                     case BeginGetResponseEventId:
-                        this.OnBeginGetResponse(eventData);
+                        if (!DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated)
+                        {
+                            // request is handled by Desktop DiagnosticSource Listener
+                            this.OnBeginGetResponse(eventData);
+                        }
+
                         break;
                     case EndGetResponseEventId:
                         this.OnEndGetResponse(eventData);
                         break;
                     case BeginGetRequestStreamEventId:
-                        this.OnBeginGetRequestStream(eventData);
+                        if (!DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated)
+                        {
+                            // request is handled by Desktop DiagnosticSource Listener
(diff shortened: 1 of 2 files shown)


cijothomas:
This is Revert of Revert..

Src/DependencyCollector/Shared/Implementation/FrameworkHttpEventListener.cs
```diff
@@ -78,13 +78,23 @@ protected override void OnEventWritten(EventWrittenEventArgs eventData)
                 switch (eventData.EventId)
                 {
                     case BeginGetResponseEventId:
-                        this.OnBeginGetResponse(eventData);
+                        if (!DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated)
+                        {
+                            // request is handled by Desktop DiagnosticSource Listener
+                            this.OnBeginGetResponse(eventData);
```
cijothomas: Lets keep the logging here to indicate begin is skipped due to DiagSource.
DependencyCollectorEventSource.Log.SkipTrackingTelemetryItemWithEventSource(id);

Src/DependencyCollector/Shared/Implementation/FrameworkHttpEventListener.cs
```diff
@@ -78,13 +78,23 @@ protected override void OnEventWritten(EventWrittenEventArgs eventData)
...
                         break;
                     case EndGetResponseEventId:
                         this.OnEndGetResponse(eventData);
                         break;
                     case BeginGetRequestStreamEventId:
-                        this.OnBeginGetRequestStream(eventData);
+                        if (!DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated)
+                        {
+                            // request is handled by Desktop DiagnosticSource Listener
+                            this.OnBeginGetRequestStream(eventData);
```
cijothomas: same comment as above - log here about skipping.

cijothomas:
@lmolkova Can you adress the 2 feedbacks.

Src/DependencyCollector/Shared/Implementation/FrameworkHttpEventListener.cs
lmolkova: In order to log something we need to get id from the payload, otherwise log will be useless. The whole point to avoid parsing payload. We will have log about callback from diagnostic source, it will be enough to understand that it was skipped by event source

cijothomas: got it.

Issue #540, opened 2017-05-04
Lack of stop event on system.data.sqlclient.sqlconnection.open breaks EventSource activity tracking


northtyphoon:
The following code in ProfileRuntimeInstrumenation.cs instrumented System.Data.SqlClient.SqlConnection.Open function. The instrumentation only emits Microsoft-ApplicationInsights-Data/Operation/Start event but not Stop event. It breaks the EventSource activity tracking used in Application Insights Profiler. It caused the subsequent sql activity be mis-correlated as child activity.

            // Decorates SqlConnection.Open, 0 params(+this)
            // Tracks dependency call only in case of failure
            Functions.Decorate(
                "System.Data",
                "System.Data.dll",
                "System.Data.SqlClient.SqlConnection.Open",
                sqlCallbacks.OnBeginForOneParameter,
                null,
                sqlCallbacks.OnExceptionForOneParameter,
                isStatic: false);

Microsoft-ApplicationInsights-Data/Operation/Start                                 	6,035.317		//2/1/3/  
Microsoft-ApplicationInsights-Data/Operation/Start                                 	6,035.713		//2/1/3/1/

SergeyKanzhelev:
What milestone are we tracking this bug for? I had impression that it is important. @northtyphoon we discussed the possible fix with you. Do you plan to send a PR?

northtyphoon:
We need the fix in 2.4. I will sync with @iusafaro .

PR #567, opened 2017-05-11
Lack of stop event on system.data.sqlclient.sqlconnection.open breaks EventSource activity tracking #540

iusafaro:
Lack of stop event on system.data.sqlclient.sqlconnection.open breaks EventSource activity tracking #540

### DIFF
--- a/Src/DependencyCollector/Shared.Tests/Implementation/ProfilerSqlProcessingTest.cs
+++ b/Src/DependencyCollector/Shared.Tests/Implementation/ProfilerSqlProcessingTest.cs
@@ -143,6 +143,22 @@
         }
 
         [TestMethod]
+        public void OnEndStopActivityOnlyDoesNotSendTelemetry_1ArgumentOverride()
+        {
+            var command = GetSqlCommandTestForStoredProc();
+            var returnObjectPassed = new object();
+
+            Stopwatch stopwatch = new Stopwatch();
+            stopwatch.Start();
+            var context = this.sqlCommandProcessingProfiler.OnBeginForOneParameter(command);
+            var objectReturned = this.sqlCommandProcessingProfiler.OnEndStopActivityOnlyForOneParameter(context, returnObjectPassed, command);
+            stopwatch.Stop();
+
+            Assert.AreSame(returnObjectPassed, objectReturned, "Object returned is not the same as expected return object");
+            Assert.AreEqual(0, this.sendItems.Count, "No telemetry item should be sent");
+        }
+
+        [TestMethod]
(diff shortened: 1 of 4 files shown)


Src/DependencyCollector/Shared.Tests/Implementation/ProfilerSqlProcessingTest.cs
```diff
@@ -143,6 +143,22 @@ public void OnEndSendsCorrectTelemetry_1ArgumentOverride()
         }
 
         [TestMethod]
```
SergeyKanzhelev: you need a test to validate that `End` for dependency telemetry will in fact be called

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
```diff
@@ -402,22 +413,29 @@ private void OnEndInternal(object exceptionObj, object thisObj)
...
+                        telemetry.Success = false;
+                        telemetry.Properties.Add("ErrorMessage", exception.Message);
+
+                        var sqlEx = exception as SqlException;
+                        telemetry.ResultCode = sqlEx != null ? sqlEx.Number.ToString(CultureInfo.InvariantCulture) : "0";
+                    }
+                    else
+                    {
+                        telemetry.Success = true;
+                        telemetry.ResultCode = "0";
```
SergeyKanzhelev: this is not required as `ResultCode` is optional. It's not relevant to this review though...

Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
```diff
@@ -371,7 +381,8 @@ private void OnEnd(object exceptionObj, object thisObj)
         /// </summary>
         /// <param name="exceptionObj">The exception object if any.</param>
         /// <param name="thisObj">This object.</param>
-        private void OnEndInternal(object exceptionObj, object thisObj)
+        /// <param name="processTelemetryItem">True if telemetry item should be processed.</param>
```
SergeyKanzhelev: name is not clear. Can you call it `sendTelemetryItem` instead. Description for `param` should also be more specific

Src/DependencyCollector/Shared.Tests/Implementation/ProfilerSqlProcessingTest.cs
iusafaro: done. also updated OnEndExceptionAsync to stop operation.

PR #570, opened 2017-05-12
Ignore deprecated events from HttpClient on .NET Core 2.0

### LINKED ISSUE #566: Net core 2.0 HttpClient will start to send new and deprecated events in the next release
Currently HttpClient on .net core 2.0 check if someone listens to 'System.Net.Http.HttpRequestOut*' event and then send 'new' evenst only.
Otherwise, for backward compatibility it also checks if someone listens to deprecated 'System.Net.Http.Request/Response' events.

This introduces issue described in https://github.com/dotnet/corefx/issues/18762 when 2 different listeners are interested in different events: e.g. Glimpse of old version and AI of new version.

In the future release of HttpClient, it will check for both events and will send both if they are enabled: https://github.com/dotnet/corefx/pull/19659

Http**Core**DiagnosticSourceListener listens to both sets of events (because it targets netstandard1.6 and does not know which runtime will be used). So with the future HttpClient, it will report the same telemetry twice.

AI Listener needs to ignore 'old' event on the new runtime similarly to how it's done in AspNetCore incoming request [listener](https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/develop/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs#L26), but old events should be rejected in IsEnabled callback, rather than in the event callback.

This needs to be done in 2.4.0 stable release

lmolkova:
1. It fixes #566 : ignores deprecated events on .net core 2.0 and appinsights internal urls in IsEnabled callback
2. Uses local endpoint in tests and addresses #494 for .net core tests

### DIFF
--- a/Src/DependencyCollector/NetCore.Tests/DependencyTrackingTelemetryModuleTestNetCore.cs
+++ b/Src/DependencyCollector/NetCore.Tests/DependencyTrackingTelemetryModuleTestNetCore.cs
@@ -2,16 +2,23 @@
 {
     using System;
     using System.Diagnostics;
+    using System.Collections.Generic;
     using System.Linq;
     using System.Net.Http;
     using System.Threading;
     using System.Threading.Tasks;
-    using Microsoft.ApplicationInsights.Channel;
+    using Microsoft.ApplicationInsights.Common;
     using Microsoft.ApplicationInsights.DataContracts;
     using Microsoft.ApplicationInsights.DependencyCollector;
     using Microsoft.ApplicationInsights.DependencyCollector.Implementation;
     using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+    using Microsoft.ApplicationInsights.TestFramework;
+    using Microsoft.AspNetCore.Builder;
+    using Microsoft.AspNetCore.Hosting;
+    using Microsoft.AspNetCore.Http;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
+
 
     /// <summary>
     /// .NET Core specific tests that verify Http Dependencies are collected for outgoing request
@@ -20,7 +27,43 @@
     public class DependencyTrackingTelemetryModuleTestNetCore
     {
         private const string IKey = "F8474271-D231-45B6-8DD4-D344C309AE69";
-        private const string FakeProfileApiEndpoint = ApplicationInsightsUrlFilter.TelemetryServiceEndpoint;
+        private const string FakeProfileApiEndpoint = "https://dc.services.visualstudio.com/v2/track";
+        private const string localhostUrl = "http://localhost:5050";
+
+        private StubTelemetryChannel channel;
+        private TelemetryConfiguration config;
+        private List<DependencyTelemetry> sentTelemetry;
+
+        /// <summary>
+        /// Initialize.
+        /// </summary>
+        [TestInitialize]
+        public void Initialize()
+        {
+            this.sentTelemetry = new List<DependencyTelemetry>();
+
+            this.channel = new StubTelemetryChannel
+            {
+                OnSend = telemetry =>
+                {
+                    // The correlation id lookup service also makes http call, just make sure we skip that
+                    DependencyTelemetry depTelemetry = telemetry as DependencyTelemetry;
+                    if (depTelemetry != null)
+                    {
+                        this.sentTelemetry.Add(depTelemetry);
+                    }
+                },
+                EndpointAddress = FakeProfileApiEndpoint
+            };
+
+            this.config = new TelemetryConfiguration
+            {
+                InstrumentationKey = IKey,
+                TelemetryChannel = this.channel
+            };
+
+            this.config.TelemetryInitializers.Add(new OperationCorrelationTelemetryInitializer());
+        }
 
         /// <summary>
         /// Cleans up.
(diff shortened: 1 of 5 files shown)


lmolkova:
I also want to improve tests a bit to have common verification with Net46 tests. 
I will do it after #553 is merged

Src/DependencyCollector/Shared/Implementation/DependencyCollectorEventSource.cs
```diff
@@ -307,6 +307,56 @@ public void HttpDesktopEndCallbackCalled(long id, string appDomainName = "Incorr
             this.WriteEvent(28, id, this.ApplicationName);
         }
 
+        [Event(
```
cijothomas: thnx for adding all the logs.

Src/DependencyCollector/NetCore.Tests/DependencyTrackingTelemetryModuleTestNetCore.cs
```diff
@@ -2,104 +2,104 @@
...
+    using Microsoft.AspNetCore.Hosting;
+    using Microsoft.AspNetCore.Http;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
 
+
     /// <summary>
     /// .NET Core specific tests that verify Http Dependencies are collected for outgoing request
     /// </summary>
     [TestClass]
     public class DependencyTrackingTelemetryModuleTestNetCore
```
cijothomas: Do we collect failed dependencies in core yet? If not just add a unit test or mention that in comment.

cijothomas:
Wait for Davids approval before merge.
@dnduffy - can you also take a look as this affect aspnet core only

Src/DependencyCollector/Shared.Tests/Implementation/FrameworkHttpProcessingTest.cs
```diff
@@ -253,7 +253,7 @@ public void OnEndHttpCallbackWithoutStatusCodeDoesNotRemoveTelemetryFromCacheWhe
             this.httpProcessingFramework.OnBeginHttpCallback(100, this.testUrl.OriginalString);
             Assert.IsNotNull(this.cache.Get(100));
 
-            DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated = true;
+            DependencyTableStore.IsDesktopHttpDiagnosticSourceActivated = true;
```
dnduffy: .Instance?

dnduffy:
@lmolkova Does this PR need DependencyTableStore.IsDesktopHttpDiagnosticSourceActivated updated to DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated?

Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs
```diff
@@ -164,17 +174,16 @@ internal void OnException(Exception exception)
...
-            if (Activity.Current == null)
+            var currentActivity = Activity.Current;
+            if (currentActivity == null)
             {
                 DependencyCollectorEventSource.Log.CurrentActivityIsNull();
                 return;
             }
 
-            if (request != null &&
-                !this.applicationInsightsUrlFilter.IsApplicationInsightsUrl(request.RequestUri))
```
Dmitry-Matveev: Does it mean we want to have headers on AI Url or this check moved up/down a bit and is still performed? Is there a guarantee that request is not null?

lmolkova:
> Does this PR need DependencyTableStore.IsDesktopHttpDiagnosticSourceActivated updated to DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated

Nope, it was tmp change to deal with #555 merge issue fixed with #573

Src/DependencyCollector/NetCore.Tests/DependencyTrackingTelemetryModuleTestNetCore.cs
lmolkova: we collect them starting from  netcore2.0. Added comment, TODO and test for it.

Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs
lmolkova: This chack is [moved to](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/993cab25f0076ef6942ad6e1a640e18fcc14fac1/Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs#L451) IsEnabled callback

So AI internal calls are rejected a bit earlier, which allows to save a bit on HttpClient instrumentation and  DiagnosticSource payloads and  Activity objects creation.

At this stage of request processing in .netcore  (when the event is fired from HttpClient) request could not be null, HttpClient will throw before that.

Src/DependencyCollector/NetCore.Tests/DependencyTrackingTelemetryModuleTestNetCore.cs
```diff
@@ -2,104 +2,104 @@
...
                 var request = new HttpRequestMessage(HttpMethod.Get, url);
-                await new HttpClient().SendAsync(request);
 
-                // on netcoreapp1.0 DiagnosticSource event is fired asycronously, let's wait for it 
+                using (new LocalServer(localhostUrl))
+                {
+                    await new HttpClient().SendAsync(request);
+                }
+
+                // on netcoreapp1.0 DiagnosticSource event is fired asychronously, let's wait for it
```
dnduffy: spelling: "asynchronously"
nitpick: Sentence capitalization and punctuation.

Src/DependencyCollector/NetCore.Tests/DependencyTrackingTelemetryModuleTestNetCore.cs
```diff
@@ -110,60 +110,114 @@ public void Cleanup()
...
+                var url = new Uri(localhostUrl);
                 var request = new HttpRequestMessage(HttpMethod.Get, url);
-                await new HttpClient().SendAsync(request);
+                using (new LocalServer(localhostUrl))
+                {
+                    await new HttpClient().SendAsync(request);
+                }
 
-                // on netcoreapp1.0 DiagnosticSource event is fired asycronously, let's wait for it 
+                // on netcoreapp1.0 DiagnosticSource event is fired asychronously, let's wait for it
```
dnduffy: "Asynchronously" is still not quite right.
Same nitpick as above about sentence format.

PR #575, opened 2017-05-12
Use local server to test Http dependency collector

lmolkova:
Addresses .NET 46 part of #494

### DIFF
--- a/Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
+++ b/Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
@@ -120,7 +120,7 @@
 
         [TestMethod]
         [Timeout(5000)]
-        public async Task TestDependencyCollectionEventSource404()
+        public async Task TestDependencyCollectionEventSourceNonSuccessStatusCode()
         {
             using (var module = new DependencyTrackingTelemetryModule())
             {
@@ -151,7 +151,7 @@
 
         [TestMethod]
         [Timeout(5000)]
-        public async Task TestDependencyCollectionDiagnosticSource404()
+        public async Task TestDependencyCollectionDiagnosticSourceNonSuccessStatusCode()
         {
             using (var module = new DependencyTrackingTelemetryModule())
             {
@@ -185,6 +185,7 @@
                     request.GetResponse();
                 }
 
+                // HttpDesktopDiagnosticListener cannot collect dependencies if HttpWebResponse was not closed/disposed
                 Assert.IsFalse(this.sentTelemetry.Any());
                 var requestId = request.Headers[RequestResponseHeaders.RequestIdHeader];
                 var rootId = request.Headers[RequestResponseHeaders.StandardRootIdHeader];
@@ -225,24 +226,27 @@
 
         [TestMethod]
         [Timeout(5000)]
-        public async Task TestDependencyCollectionFailedRequestDiagnosticSource()
+        public async Task TestDependencyCollectionDnsIssueRequestDiagnosticSource()
         {
             using (var module = new DependencyTrackingTelemetryModule())
             {
                 module.ProfileQueryEndpoint = FakeProfileApiEndpoint;
                 module.Initialize(this.config);
 
-                var url = new Uri($"http://{Guid.NewGuid()}:123/");
+                var url = new Uri($"http://{Guid.NewGuid()}/");
                 HttpClient client = new HttpClient();
                 await client.GetAsync(url).ContinueWith(t => { });
 
+                // here the start of dependency is tracked with HttpDesktopDiagnosticSourceListener, 
+                // so the expected SDK version should have DiagnosticSource 'rdddsd' prefix. 
+                // however the end is tracked by FrameworkHttpEventListener
                 this.ValidateTelemetryForDiagnosticSource(this.sentTelemetry.Single(), url, null, false, string.Empty);
             }
         }
 
         [TestMethod]
         [Timeout(5000)]
-        public async Task TestDependencyCollectionFailedRequestEventSource()
+        public async Task TestDependencyCollectionDnsIssueRequestEventSource()
         {
             using (var module = new DependencyTrackingTelemetryModule())
             {
@@ -250,7 +254,7 @@
                 module.ProfileQueryEndpoint = FakeProfileApiEndpoint;
                 module.Initialize(this.config);
 
-                var url = new Uri($"http://{Guid.NewGuid()}:123/");
+                var url = new Uri($"http://{Guid.NewGuid()}/");
                 HttpClient client = new HttpClient();
                 await client.GetAsync(url).ContinueWith(t => { });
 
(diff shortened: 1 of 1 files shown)


Src/DependencyCollector/Shared.Tests/Implementation/FrameworkHttpProcessingTest.cs
```diff
@@ -253,7 +253,7 @@ public void OnEndHttpCallbackWithoutStatusCodeDoesNotRemoveTelemetryFromCacheWhe
             this.httpProcessingFramework.OnBeginHttpCallback(100, this.testUrl.OriginalString);
             Assert.IsNotNull(this.cache.Get(100));
 
-            DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated = true;
+            DependencyTableStore.IsDesktopHttpDiagnosticSourceActivated = true;
```
lmolkova: please ignore it, develop if broken by #555 and fixed in #573

Src/DependencyCollector/Shared/Implementation/FrameworkHttpProcessing.cs
```diff
@@ -140,7 +140,7 @@ public void OnEndHttpCallback(long id, bool? success, bool synchronous, int? sta
                     // We also had duplicates if runtime is also 4.5.2 (4.6 runtime has no such problem)
                     // So starting with 2.1.0-beta4 we are cutting support for HTTP dependencies in .NET 4.5.2.
                     // But we will let DesktopDiagnosticSourceListener collect dependency if it is activated 
-                    if (!DependencyTableStore.Instance.IsDesktopHttpDiagnosticSourceActivated)
+                    if (!DependencyTableStore.IsDesktopHttpDiagnosticSourceActivated)
```
lmolkova: please ignore it, develop if broken by #555 and fixed in #573

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
```diff
@@ -108,7 +120,7 @@ public void TestDependencyCollectionEventSource()
 
         [TestMethod]
         [Timeout(5000)]
-        public void TestDependencyCollectionEventSource404()
+        public async Task TestDependencyCollectionEventSource404()
```
cijothomas: Can you remove 404 from the test name and replace with just 'failed' - I used 404 earlier as i google.com/404 was the only reliable http call which returned non-sucess status code. Now that you control the 'server', you can return any fail status code. Removing 404 and naming it just 'fail' clears any confusion people may have in future about 404 being special in someway..

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
```diff
@@ -117,28 +129,42 @@ public void TestDependencyCollectionEventSource404()
...
-                var url = new Uri("http://google.com/404");
                 HttpClient httpClient = new HttpClient();
-                httpClient.GetStringAsync(url).ContinueWith(t => { }).Wait();
+                using (new LocalServer(LocalhostUrl, ctx => ctx.Response.StatusCode = 404))
+                {
+                    await httpClient.GetStringAsync(LocalhostUrl).ContinueWith(t => { });
+                }
 
-                this.ValidateTelemetryForDiagnosticSource(this.sentTelemetry.Single(), url, null, false, "404");
+                this.ValidateTelemetryForDiagnosticSource(this.sentTelemetry.Single(), new Uri(LocalhostUrl), null, false, "404");
```
cijothomas: can you add a comment to make it explicit that DiagnosticSource alone is not collecting dependency here, but a combined effort by DiagnosticSource and FrameworkES, but the expected prefix will be 'rdddsd' because we mark the prefix based on who started monitoring the dependency.  (This avoids any confusion to future readers about ability of DiagnosticSource to collect failures)

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
```diff
@@ -151,9 +177,13 @@ public void TestNoDependencyCollectionDiagnosticSourceNoResponseClose()
...
 
-                var url = new Uri("https://www.bing.com/");
+                var url = new Uri(LocalhostUrl);
                 HttpWebRequest request = WebRequest.CreateHttp(url);
-                HttpWebResponse response = (HttpWebResponse)request.GetResponse();
+
+                using (new LocalServer(LocalhostUrl))
+                {
+                    request.GetResponse();
+                }
```
cijothomas: Please add a comment indicating DiagnosticSource won't collect dependency if response is not disposed.

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
```diff
@@ -192,52 +225,91 @@ public void TestDependencyCollectionDiagnosticSourceWithParentActivity()
 
         [TestMethod]
         [Timeout(5000)]
-        public void TestDependencyCollectionFailedRequestDiagnosticSource()
+        public async Task TestDependencyCollectionFailedRequestDiagnosticSource()
```
cijothomas: Rename test to indicate the failure is at DNS Resolution itself , compared to failure due to status code > 400

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
```diff
@@ -192,52 +225,91 @@ public void TestDependencyCollectionDiagnosticSourceWithParentActivity()
...
+                HttpClient client = new HttpClient();
+                await client.GetAsync(url).ContinueWith(t => { });
+
+                this.ValidateTelemetryForDiagnosticSource(this.sentTelemetry.Single(), url, null, false, string.Empty);
+            }
+        }
+
+        [TestMethod]
+        [Timeout(5000)]
+        public async Task TestDependencyCollectionFailedRequestEventSource()
```
cijothomas: rename to indicate failure is at DNS resolution itself.

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
```diff
@@ -270,7 +342,7 @@ public void OnBeginOnEndAreNotCalledForAppInsightsUrl()
         private void ValidateTelemetryForDiagnosticSource(DependencyTelemetry item, Uri url, WebRequest request, bool success, string resultCode)
         {
             Assert.AreEqual(url, item.Data);
-            Assert.AreEqual(url.Host, item.Target);
+            Assert.AreEqual($"{url.Host}:{url.Port}", item.Target);
```
cijothomas: Add a comment what we collect port only if its non-standard port like 80,443, and for local server the port is non-standard.
If possible, make a test localhost:80 or 443, and validate that we don't collect port in those cases.

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
lmolkova: I think 404 is not even a failure, it's non 20* status code, and fail could mean that request was not send at all (e.g. because of DNS or network issue). Let's call in as NonSuccessStatusCode

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
lmolkova: Note that for 404 (where the response is available) DiagnosticSource tracks the end. Event source tracks dependency only for dns or other network issue (when there is no response). I'll update corresponding test

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
lmolkova: I'm afraid to use localhost:80 in tests, I suspect the port could be used in multiple cases on build and developers machines.
I'll use it in dns issue test, so we'll verify it.
It's checked with FrameworkHttpProcessingTest for event source

Src/DependencyCollector/Net46.Tests/DependencyTrackingTelemetryModuleTestNet46.cs
cijothomas: NonSuccessStatusCode sounds good.

lmolkova:
ping. I think I addressed all the comments, please let me know if I missed anything or merge it

Issue #510, opened 2017-04-27
"Extreme App" monitoring noise significantly


SergeyKanzhelev:
Extreme App is the app with sampling turned off with 5 traces in every request. Monitoring noise increased in the latest SDK (2.4.0-beta1) comparing to what it was in 2.3.

We have new features that might have contributed:
- new TelemetryCorrelation http module that subscribes to the new callback `PreRequestHandlerExecute`
- Pre-aggregated metrics - new locks on sync path
- exceptions statistics - additional data collection

@cijothomas and @kstevenham investigating

SergeyKanzhelev:
Possible improvements after running local perf analysis. Disclaimer here - I didn't configure endpoints correctly so some numbers are skewed. 

Degradation:
1. Cache the method we take with reflection: `MethodInfo addOnSendingHeadersMethod = httpApplication.Response.GetType().GetMethod("AddOnSendingHeaders");`. Also no need to instantiate the parameters array all the time - can only do it once. 
2. Why do we need to call `if (this.IsFirstRequest(httpApplication))` in OnEndRequest for non `NET40`? Why to subscribe on event at all?
3. `ClientIpHeaderTelemetryInitializer` takes the most on my test. It may be even worse as on local box verbose tracing got enabled when I turn on sampling profiling. It looks like most of this time is IP validation that can/should be moved down to the sanitize method and run asynchronously. Also no need to allocate `var requestWrapper = new HttpRequestWrapper(platformContext.Request);`
4. Second most expensive initializer - `SyntheticUserAgentTelemetryInitializer`. This initializer for some reason do not cache the value of SyntheticSource in RequestTelemetry. So parsing happening all the time - for every telemetry separately. It will be extremely visible in the app with a lot of request-related telemetry.  Also `foreach` is less efficient than `for` and dictionary do not contain empty elements - no need to check every time. One more - taking `request.UserAgent` on every iteration is super expensive:

  ```
  public string UserAgent
  {
	get
	{
		if (this._wr != null)
		{
			return this._wr.GetKnownRequestHeader(39);
		}
		return null;
	}
  }
  ```
5. User cookie - can replace `Split` to `Substring`. `var cookieParts = userCookie.Value.Split('|');` Same with session
6. Operation name initializer makes way too many concatenations. Maybe use `StringBuilder`. In the next version we can just take the routeTemplate instead of constructing manually

SergeyKanzhelev:
[...]

PR #579, opened 2017-05-12
Small perf improvments.

cijothomas:
1.Caches reflective lookup of method addOnSendingHeadersMethod to improve performance

2. Params for reflective call is a Action, which is now created once , and re-used vs creating new action every time. Again, this is to improve performance, and no functionality change is expected.

Address a portion of https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/510

### DIFF
--- a/Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
+++ b/Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
@@ -18,7 +18,9 @@
     {
         private readonly RequestTrackingTelemetryModule requestModule;
         private readonly ExceptionTrackingTelemetryModule exceptionModule;
-
+        private MethodInfo addOnSendingHeadersMethod;
+        private bool? addOnSendingHeadersMethodExists;
+        object[] paramsForReflectiveCall;
         /// <summary>
         /// Indicates if module initialized successfully.
         /// </summary>
@@ -47,6 +49,27 @@
                         }
                     }
                 }
+
+                this.addOnSendingHeadersMethod = null;
+                this.addOnSendingHeadersMethodExists = null;
+
+                this.paramsForReflectiveCall = new object[]
+                        {
+                            new Action<HttpContext>((httpContext) =>
+                            {
+                                try
+                                {
+                                    if (this.requestModule != null)
+                                    {
+                                        this.requestModule.AddTargetHashForResponseHeader(httpContext);
+                                    }
+                                }
+                                catch (Exception ex)
+                                {
+                                    WebEventSource.Log.AddTargetHeaderFailedWarning(ex.ToInvariantString());
+                                }
+                            })
+                        };
             }
             catch (Exception exc)
             {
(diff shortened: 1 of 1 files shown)


Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
```diff
@@ -121,29 +145,15 @@ private void AddCorreleationHeaderOnSendRequestHeaders(HttpApplication httpAppli
...
-                                }
-                                catch (Exception ex)
-                                {
-                                    WebEventSource.Log.AddTargetHeaderFailedWarning(ex.ToInvariantString());
-                                }
-                            })
-                        };
-
-                        addOnSendingHeadersMethod.Invoke(httpApplication.Response, parameters);
+                        this.addOnSendingHeadersMethod = httpApplication.Response.GetType().GetMethod("AddOnSendingHeaders");
```
SergeyKanzhelev: Why not in constructor?

Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
```diff
@@ -121,29 +145,15 @@ private void AddCorreleationHeaderOnSendRequestHeaders(HttpApplication httpAppli
...
+                        this.addOnSendingHeadersMethodExists = this.addOnSendingHeadersMethod != null;
+                    }
+                    
+                    if (this.addOnSendingHeadersMethod != null)
+                    {                        
+                        this.addOnSendingHeadersMethod.Invoke(httpApplication.Response, this.paramsForReflectiveCall);
```
SergeyKanzhelev: Invoke is more expensive than delegate. See https://gist.github.com/jmangelo/3173109#file-gistfile1-cs-L57-L66 for example use

Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
```diff
@@ -18,6 +18,9 @@ public sealed class ApplicationInsightsHttpModule : IHttpModule
     {
         private readonly RequestTrackingTelemetryModule requestModule;
         private readonly ExceptionTrackingTelemetryModule exceptionModule;
+        private MethodInfo addOnSendingHeadersMethod;
+        private bool? addOnSendingHeadersMethodExists;
+        private object[] paramsForReflectiveCall;
```
SergeyKanzhelev: Name should be reflective of content. This works better: `addOnSendingHeadersMethodParams`

Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
cijothomas: will do.

Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
cijothomas: done

Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
cijothomas: done

cijothomas:
@SergeyKanzhelev Addressed the comments. Made invocation using delegate - but only for !Net40 as there was this interface ISubscriptionToken, returned by addOnSendingHeaders(), which is not available in NET40 but is required for constructing the open delegate.

cijothomas:
Even without the recent fix (delegate vs invoke), the overhead from 9000 RPS app reduced from 39 to 38. (2 runs made).
Will do more runs with the  recent fix included.

PR #580, opened 2017-05-13
SyntheticUserAgentTelemetryInitializer perf optimization

cijothomas:
SyntheticUserAgentTelemetryInitializer optimize to store and re-use syntheticsource from a cache. This improves performance where there are multiple telemetry items from the same context - after the 1st item, syntheticsource is read from cache.

Attempt to address item 4 from https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/510

### DIFF
--- a/Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
+++ b/Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
@@ -11,8 +11,8 @@
     /// </summary>
     public class SyntheticUserAgentTelemetryInitializer : WebTelemetryInitializerBase
     {
-        private const string syntheticSourceNameKey = "operationsyntheticsource";
-        private const string syntheticSourceName = "Bot";
+        private const string SyntheticSourceNameKey = "Microsoft.ApplicationInsights.RequestTelemetry.SyntheticSource";
+        private const string SyntheticSourceName = "Bot";
         private string filters = string.Empty;
         private string[] filterPatterns;        
 
@@ -51,9 +51,9 @@
             {
                 if (platformContext != null)
                 {
-                    if (platformContext.Items.Contains(syntheticSourceNameKey))
+                    if (platformContext.Items.Contains(SyntheticSourceNameKey))
                     {
-                        telemetry.Context.Operation.SyntheticSource = platformContext.Items[syntheticSourceNameKey].ToString();
+                        telemetry.Context.Operation.SyntheticSource = platformContext.Items[SyntheticSourceNameKey].ToString();
                     }
                     else
                     {
@@ -67,8 +67,8 @@
                             {
                                 if (userAgent.IndexOf(this.filterPatterns[i], StringComparison.OrdinalIgnoreCase) != -1)
                                 {
-                                    telemetry.Context.Operation.SyntheticSource = syntheticSourceName;
-                                    platformContext.Items.Add(syntheticSourceNameKey, syntheticSourceName);
+                                    telemetry.Context.Operation.SyntheticSource = SyntheticSourceName;
+                                    platformContext.Items.Add(SyntheticSourceNameKey, SyntheticSourceName);
                                     return;
                                 }
                             }
--- a/Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
+++ b/Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
@@ -55,10 +55,10 @@
                     {
                         // We expect customers to configure telemetry initializer before they add it to active configuration
                         // So we will not protect fiterPatterns array with locks (to improve perf)
-                        foreach (string pattern in this.filterPatterns)
+                        string userAgent = request.UserAgent;
+                        for (int i = 0; i < this.filterPatterns.Length; i++)
                         {
-                            if (!string.IsNullOrWhiteSpace(pattern) &&
-                                request.UserAgent.IndexOf(pattern, StringComparison.OrdinalIgnoreCase) != -1)
+                            if (userAgent.IndexOf(this.filterPatterns[i], StringComparison.OrdinalIgnoreCase) != -1)
                             {
                                 telemetry.Context.Operation.SyntheticSource = "Bot";
                                 return;
--- a/Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
+++ b/Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
@@ -11,8 +11,10 @@
     /// </summary>
     public class SyntheticUserAgentTelemetryInitializer : WebTelemetryInitializerBase
     {
+        private const string syntheticSourceNameKey = "operationsyntheticsource";
+        private const string syntheticSourceName = "Bot";
         private string filters = string.Empty;
-        private string[] filterPatterns;
+        private string[] filterPatterns;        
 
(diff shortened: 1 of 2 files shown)


Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
```diff
@@ -49,19 +51,26 @@ protected override void OnInitializeTelemetry(HttpContext platformContext, Reque
...
-                        // So we will not protect fiterPatterns array with locks (to improve perf)
-                        foreach (string pattern in this.filterPatterns)
+                        var request = platformContext.GetRequest();
+                        if (request != null && !string.IsNullOrEmpty(request.UserAgent))
                         {
-                            if (!string.IsNullOrWhiteSpace(pattern) &&
-                                request.UserAgent.IndexOf(pattern, StringComparison.OrdinalIgnoreCase) != -1)
+                            // We expect customers to configure telemetry initializer before they add it to active configuration
+                            // So we will not protect fiterPatterns array with locks (to improve perf)
+                            string userAgent = request.UserAgent;
```
SergeyKanzhelev: take it earlier  - no need to take `request.UserAgent` above.

Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
```diff
@@ -49,19 +51,26 @@ protected override void OnInitializeTelemetry(HttpContext platformContext, Reque
             {
                 if (platformContext != null)
                 {
-                    var request = platformContext.GetRequest();
-
-                    if (request != null && !string.IsNullOrEmpty(request.UserAgent))
+                    if (platformContext.Items.Contains(SyntheticSourceNameKey))
```
SergeyKanzhelev: first thing to check is SyntheticSource from the `requestTelemetry`. Only store the fact that request is not synthetic in `platformContext.Items`

cijothomas: I couldn't find the reason how checking requestTelemetry will improve perf even. Even for the requestTelemetry, initializer must be called to populate the field, and by doing that, the value is cached to platformConext.Items.

Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
cijothomas: yes right.

Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
SergeyKanzhelev: it's not only about optimization. It is also scenario - when one wants to override the SyntheticSource for Request from code - we will pick it up by this initializer

SergeyKanzhelev: I agree that code looks simpler without that check, but from scenario perspective it will be better to store it in requesttelemetry

SergeyKanzhelev:
You can use that logic here: `WebTestTelemetryInitializer`. At least cache the value in `RequestTelemetry`

Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
```diff
@@ -49,19 +51,26 @@ protected override void OnInitializeTelemetry(HttpContext platformContext, Reque
...
-                        // We expect customers to configure telemetry initializer before they add it to active configuration
-                        // So we will not protect fiterPatterns array with locks (to improve perf)
-                        foreach (string pattern in this.filterPatterns)
+                        var request = platformContext.GetRequest();
+                        if (request != null && !string.IsNullOrEmpty(request.UserAgent))
                         {
-                            if (!string.IsNullOrWhiteSpace(pattern) &&
-                                request.UserAgent.IndexOf(pattern, StringComparison.OrdinalIgnoreCase) != -1)
+                            // We expect customers to configure telemetry initializer before they add it to active configuration
+                            // So we will not protect fiterPatterns array with locks (to improve perf)
```
pharring: typo ```fiterPatterns``` -> ```filterPatterns```

Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
```diff
@@ -11,8 +11,10 @@
     /// </summary>
     public class SyntheticUserAgentTelemetryInitializer : WebTelemetryInitializerBase
     {
+        private const string SyntheticSourceNameKey = "Microsoft.ApplicationInsights.RequestTelemetry.SyntheticSource";
```
pharring: Is there any need to have this be a human-readable string? If not, consider a static object:
```C#
private static readonly object SyntheticSourceNameKey = new object();
```
It's GetHashCode implementation will be faster and, therefore, look-ups and inserts will be too.

Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
```diff
@@ -49,19 +51,26 @@ protected override void OnInitializeTelemetry(HttpContext platformContext, Reque
             {
                 if (platformContext != null)
                 {
-                    var request = platformContext.GetRequest();
-
-                    if (request != null && !string.IsNullOrEmpty(request.UserAgent))
+                    if (platformContext.Items.Contains(SyntheticSourceNameKey))
+                    {
+                        telemetry.Context.Operation.SyntheticSource = platformContext.Items[SyntheticSourceNameKey].ToString();
```
pharring: Why isn't the right hand side just ```SyntheticSourceName``` ?
(Frankly, the value stored in the ```.Items``` dictionary is irrelevant. It could just be ```null```)

Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
cijothomas: Current code does not seem to be reading anything from RequestTelemetry - changing that would be a behavior change instead of perf optimization attempt - do we want that for stabilization time?

Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
cijothomas: agree. will change.

cijothomas: chatted with sergey about this - he prefer to keep the same user readable name.

Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
cijothomas: yes right.

Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
cijothomas: as agreed, i will adress it in separate PR, after stable.

Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
cijothomas: fixed

PR #593, opened 2017-05-23
Check for ApplicationInsights URL on .net core 2.0 in case there are multiple subscribers with different filters

### LINKED ISSUE #592: Protect from side-effects of multiple DiagnosticSource listeners
AppInsights Web SDK listens to outgoing Http events with DiagnosticListener.

Http instrumentation (core or desktop) first calls `IsEnabled(eventName, request)` to determine if the request needs to be instrumented.
It checks whether request is done to ApplicationInsights internal URI and rejects it in `isEnabled` callback.

It's a DiagnosticListener cliche to guard every Write event with IsEnabled call.

When there are multiple subscribers to the same source, they could use different implementations of `IsEnabled`, DiagnosticListener determines if event should be fired to **ALL** subscribers if at least **ONE** of IsEnabled returned true.

Customers may want to write their own listeners to parse/inject custom headers or do any other kind of tracing. Or other tracing systems may listen to the same events (e.g. Glimpse).

In this case `HttpCoreDiagnosticSourceListener` on .net core 2.0 may start tracking internal ApplicationInsights calls that would cause stack overflow exception and dead lock for the dependency call.

So, every event callback must check for ApplicationInsights URL before proceeding with event processing.

lmolkova:
Fixes #592

### DIFF
--- a/Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs
+++ b/Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs
@@ -183,7 +183,12 @@ namespace Microsoft.ApplicationInsights.DependencyCollector.Implementation
 
             DependencyCollectorEventSource.Log.HttpCoreDiagnosticSourceListenerStart(currentActivity.Id);
 
-            this.InjectRequestHeaders(request, this.configuration.InstrumentationKey);
+            // even though we have IsEnabled filter to reject ApplicationInsights URLs before any events are fired,
+            // in case of multiple subscribers, if one subscriber returns true to IsEnabled, all subscribers will receive event
+            if (!this.applicationInsightsUrlFilter.IsApplicationInsightsUrl(request.RequestUri))
+            {
+                this.InjectRequestHeaders(request, this.configuration.InstrumentationKey);
+            }
         }
 
         //// netcoreapp 2.0 event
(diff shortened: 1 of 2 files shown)


lmolkova:
Please consider to include in stable 2.4.0 release.

Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs
```diff
@@ -163,8 +166,13 @@ internal void OnException(Exception exception)
 
             DependencyCollectorEventSource.Log.HttpCoreDiagnosticSourceListenerException(currentActivity.Id);
 
-            this.pendingExceptions.TryAdd(currentActivity.Id, exception);
-            this.client.TrackException(exception);
+            // even though we have IsEnabled filter to reject ApplicationInsights URLs before any events are fired,
+            // Exceptions are special and fired even if reuqest instrumentation is disabled
```
dnduffy: nitpick: "request" and please add punctuation.

Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs
```diff
@@ -183,7 +191,12 @@ internal void OnActivityStart(HttpRequestMessage request)
 
             DependencyCollectorEventSource.Log.HttpCoreDiagnosticSourceListenerStart(currentActivity.Id);
 
-            this.InjectRequestHeaders(request, this.configuration.InstrumentationKey);
+            // even though we have IsEnabled filter to reject ApplicationInsights URLs before any events are fired,
```
dnduffy: nitpick: Sentence capitalization and punctuation.

PR #602, opened 2017-05-24
Fixed references.

dnduffy:


lmolkova:
I suggest to understand this scenario better.

The problem is that PerformanceCollector Unit Tests target .NET45
However they also reference Web.NET40 tests and Tests dll is copied to the PerfCollector/UnitTests output folder. Since PerfCollector/UnitTests target NET45, they bring ApplicationInsights SDK .NET 45, that is not compatible with Web40 tests.

So, Web.Tests.Net40 run twice. Once from PerfCollector folder, Another time from Web.Net40.Tests.
Is it needed at all?
If so, should instead Web.NET45 tests run. I recall I've tried to change reference to NET45 tests and it did not work.

PR #594, opened 2017-05-23
Catch and log potential exceptions during DiagnosticSource initialization errors

lmolkova:
Based on SDK telemetry findings, there is issue, that reproduces in some customer environment.

The issue has never reproduced in any AppInsights/.NET tests and assumption is that issue is quite rare or specific to some setup.

When the issue reproduces, neither DiagnosticListener nor EventListener collect dependencies.
Mitigation for now is to gracefully fallback to event source and log issue.

The nature of issue is not known so far, I will continue looking for the root cause.
```
AI (Internal): [msg=RemoteDependencyModule failed];[msg=System.TypeInitializationException: The type initializer for 'System.Diagnostics.HttpHandlerDiagnosticListener' threw an exception. ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Diagnostics.DiagnosticListener..ctor(String name)
   at System.Diagnostics.HttpHandlerDiagnosticListener..cctor()
   --- End of inner exception stack trace ---
   at System.Diagnostics.DiagnosticListener.get_AllListeners()
   at Microsoft.ApplicationInsights.DependencyCollector.Implementation.HttpDesktopDiagnosticSourceSubscriber..ctor(HttpDesktopDiagnosticSourceListener parent)
   at Microsoft.ApplicationInsights.DependencyCollector.Implementation.HttpDesktopDiagnosticSourceListener..ctor(DesktopDiagnosticSourceHttpProcessing httpProcessing)
   at Microsoft.ApplicationInsights.DependencyCollector.DependencyTrackingTelemetryModule.InitializeForDiagnosticAndFrameworkEventSource()
   at Microsoft.ApplicationInsights.DependencyCollector.DependencyTrackingTelemetryModule.InitializeForRuntimeInstrumentationOrFramework()
at Microsoft.ApplicationInsights.DependencyCollector.DependencyTrackingTelemetryModule.Initialize(TelemetryConfiguration configuration)];[fwv=4.0.30319.42000];

AI (Internal): [msg=RemoteDependencyModule failed];[msg=System.TypeInitializationException: The type initializer for 'System.Diagnostics.HttpHandlerDiagnosticListener' threw an exception. ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Diagnostics.DiagnosticListener..ctor(String name)
   at System.Diagnostics.HttpHandlerDiagnosticListener..ctor()
   at System.Diagnostics.HttpHandlerDiagnosticListener..cctor()
[...]

PR #606, opened 2017-06-01
Update DiagnosticSource to preview2

lmolkova:
It updates DiagnosticSource to the preview2 version

Changes in DiagnosticSource preview2:
1. Activity Id generation (fixes sampling issue): https://github.com/Microsoft/ApplicationInsights-dotnet/issues/538
2. No machine name in Activity Id (fixes security issue https://github.com/dotnet/corefx/pull/18881)
3. Http Desktop DiagnosticSource on .net 4.5
4. Real fix for #594 (ServiceProfiler + AI do not work together on .net 46)

There are also changes in functional and unit tests as Http Desktop DiagnosticSource became available on .NET45.

-------------

Please do NOT merge until following matter is clarified: **(UPDATE: resolved - projects need to have a reference as it should be installed with AI Base SDK, nupecs do not need to be updated )**

 @dnduffy please confirm that you intentionally added (#600) dependency to DiagnosticSource to all projects
- HostingStartup
- PerformnaceCollector
- WindowsServer
(and their tests)

Do we also need to update nuspecs for these packages.


SergeyKanzhelev:
@lmolkova NuGet automatically adds reference to the closure of all NuGet references. So adding base SDK adds it's dependencies as well. No need to update the NuSpec though

Src/DependencyCollector/Net45.Tests/DependencyTrackingTelemetryModuleHttpTest.cs
```diff
@@ -345,6 +344,196 @@ public void OnBeginOnEndAreNotCalledForAppInsightsUrl()
...
+                HttpWebRequest request = WebRequest.CreateHttp(url);
+
+                using (new LocalServer(LocalhostUrl))
+                {
+                    using (request.GetResponse())
+                    {
+                    }
+                }
+
+                this.ValidateTelemetryForEventSource(this.sentTelemetry.Single(), url, request, true, "200");
```
dnduffy: nitpick: It would be nice if duplicate code was in helper functions instead of copy/paste duplicated.

lmolkova: @dnduffy Not sure I understood you correctly: do you mean that helper methods (`ValidateTelemetryForEventSource` and `ValidateTelemetryForDiagnosticSource`) has a lot in common? Or you see that there is other copy-paste code that could be moved to helper method?

dnduffy: I mean that TestDependencyCollectorPostRequestsAreCollectedDiagnosticSource and TestDependencyCollectorPostRequestsAreCollectedEventSource have a lot in common.
TestHttpRequestsWithQueryStringAreCollectedDiagnosticSource and TestHttpRequestsWithQueryStringAreCollectedEventSource have a lot in common.
And TestDependencyCollectionDiagnosticSourceRedirect and TestDependencyCollectionEventSourceRedirect have a lot in common.
Basically there are three tests each duplicated once with the only difference being whether or not module.DisableDiagnosticSourceInstrumentation is set to true.  That setting could be parameterized and the Assert on IsDesktopHttpDiagnosticSourceActivated changed to match the parameter value and then you could have three "long" test helper functions that you call twice (once with true and once with false) rather than two copies of each of the three tests.

lmolkova: I see your point now.
From one side I agree with you and if it were production code it would for sure make sense.

From the other side I think bringing more logic into unit tests does not pay-off and complicates things for someone would debug it one year later.
It would have make sense to have parameterized tests with simple switch between eventsource/ diagnosticsource, however mstest does not have good support for it and it would be difficult to understand which test has failed (we already have this problem for NET40/NET45 tests).

So I'm hesitant to make this change. But If you really want it and think it makes sense, I will do it.

dnduffy: I just mean do this:
``` csharp
[TestMethod]
[Timeout(5000)]
public void TestDependencyCollectorPostRequestsAreCollectedDiagnosticSource()
{
    DoTestDependencyCollectorPostRequestsAreCollected(true);
}

[TestMethod]
[Timeout(5000)]
public void TestDependencyCollectorPostRequestsAreCollectedEventSource()
{
    DoTestDependencyCollectorPostRequestsAreCollected(false);
}
```
Even in test code copy/paste makes it harder to maintain.

lmolkova: done, please review

cijothomas:
@lmolkova  do we need a similar PR for base sdk as well to update to the newest DiagSource?

lmolkova:
I wanted to have this PR (in Web SDK)
1) to run func tests
2) make changes in tests

In base SDK there is no code/behavior that changes with DiagnosticSource update, so there I thought to update it to stable right away. Are you ok with this plan?

cijothomas:
yes we are good.. will merge now.

Issue #522, opened 2017-04-28
Add support for EntityFramework diagnostic source listener


AlexBulankou:
This PR was never completed but it can be used as example: https://github.com/Microsoft/ApplicationInsights-aspnetcore/pull/297

SergeyKanzhelev:
CC: @TomMilos who will be working on expanding the technologies coverage for dependency monitoring

SergeyKanzhelev:
This will be handled by @anpete next week. 

Roadmap: https://github.com/aspnet/EntityFramework/wiki/Roadmap#ef-core-20-under-development
Issue: https://github.com/aspnet/EntityFramework/issues/8272

Dmitry-Matveev:
Addressed with #607 .

PR #607, opened 2017-06-01
Adds a new DiagnosticSource based Dependency Collector for SqlClient.

anpete:
Part of #522 - Add support for EntityFramework diagnostic source listener

**Update:** Pushed a new version that contains full support for SqlClient events.

### DIFF
--- a/Src/DependencyCollector/NetCore/DependencyCollector.NetCore.csproj
+++ b/Src/DependencyCollector/NetCore/DependencyCollector.NetCore.csproj
@@ -13,7 +13,7 @@
   <Import Project="..\..\..\..\packages\MicroBuild.Core.0.2.0\build\MicroBuild.Core.targets" Condition="Exists('..\..\..\..\packages\MicroBuild.Core.0.2.0\build\MicroBuild.Core.targets')" />
 
   <PropertyGroup>
-    <Copyright>Copyright © Microsoft. All Rights Reserved.</Copyright>
+    <Copyright>Copyright Â© Microsoft. All Rights Reserved.</Copyright>
     <VersionPrefix>2.5.0-beta1</VersionPrefix>
     <Authors>Microsoft</Authors>
     <TargetFramework>netstandard1.6</TargetFramework>
@@ -34,6 +34,7 @@
   <ItemGroup>
     <Compile Include="..\Shared\DependencyTrackingTelemetryModule.cs" Link="DependencyTrackingTelemetryModule.cs" />
     <Compile Include="..\Shared\HttpDependenciesParsingTelemetryInitializer.cs" Link="HttpDependenciesParsingTelemetryInitializer.cs" />
+    <Compile Include="..\Shared\Implementation\Operation\ObjectInstanceBasedOperationHolder.cs" Link="ObjectInstanceBasedOperationHolder.cs" />
     <Compile Include="..\Shared\SanitizedHostList.cs" Link="SanitizedHostList.cs" />
     <Compile Include="..\Shared\Implementation\DependencyCollectorEventSource.cs" Link="DependencyCollectorEventSource.cs" />
     <Compile Include="..\Shared\Implementation\ApplicationInsightsUrlFilter.cs" Link="ApplicationInsightsUrlFilter.cs" />
--- a/Src/DependencyCollector/NetCore/DependencyCollector.NetCore.csproj
+++ b/Src/DependencyCollector/NetCore/DependencyCollector.NetCore.csproj
@@ -39,6 +39,7 @@
     <Compile Include="..\Shared\Implementation\ApplicationInsightsUrlFilter.cs" Link="ApplicationInsightsUrlFilter.cs" />
     <Compile Include="..\Shared\Implementation\RemoteDependencyConstants.cs" Link="RemoteDependencyConstants.cs" />
     <Compile Include="..\Shared\Implementation\HttpCoreDiagnosticSourceListener.cs" Link="HttpCoreDiagnosticSourceListener.cs" />
+    <Compile Include="..\Shared\Implementation\SqlClientDiagnosticSourceListener.cs" Link="SqlClientDiagnosticSourceListener.cs" />
     <Compile Include="..\Shared\Implementation\HttpHeadersUtilities.cs" Link="HttpHeadersUtilities.cs" />
     <Compile Include="..\Shared\Implementation\PropertyFetcher.cs" Link="PropertyFetcher.cs" />
     <Compile Include="..\Shared\Implementation\RDDSource.cs" Link="RDDSource.cs" />
@@ -66,6 +67,7 @@
     <PackageReference Include="Microsoft.Extensions.PlatformAbstractions" Version="1.1.0" />
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.4.0" />
     <PackageReference Include="System.Diagnostics.StackTrace" Version="4.3.0" />
+    <PackageReference Include="System.Data.SqlClient" Version="4.3.1" />
     <PackageReference Include="MicroBuild.Core" Version="0.2.0" developmentDependency="true" />
 
   </ItemGroup>
--- a/Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
+++ b/Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
@@ -1,0 +1,458 @@
+namespace Microsoft.ApplicationInsights.Tests
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Data;
+    using System.Data.SqlClient;
+    using System.Diagnostics;
+    using System.Linq;
+
+    using Microsoft.ApplicationInsights.Channel;
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using Microsoft.ApplicationInsights.DependencyCollector.Implementation;
+
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+
+    [TestClass]
+    public class SqlClientDiagnosticSourceListenerTests
+    {
+        private const string TestConnectionString = "Data Source=(localdb)\\MSSQLLocalDB;Database=master";
+        
+        private IList<ITelemetry> sendItems;
+        private StubTelemetryChannel stubTelemetryChannel;
+        private TelemetryConfiguration configuration;
+        private FakeSqlClientDiagnosticSource fakeSqlClientDiagnosticSource;
+        private SqlClientDiagnosticSourceListener sqlClientDiagnosticSourceListener;
+
+        [TestInitialize]
+        public void TestInit()
+        {
+            this.sendItems = new List<ITelemetry>();
+            this.stubTelemetryChannel = new StubTelemetryChannel {OnSend = item => this.sendItems.Add(item)};
+
+            this.configuration = new TelemetryConfiguration
+            {
+                InstrumentationKey = Guid.NewGuid().ToString(),
+                TelemetryChannel = stubTelemetryChannel
+            };
+
+            this.fakeSqlClientDiagnosticSource = new FakeSqlClientDiagnosticSource();
+            this.sqlClientDiagnosticSourceListener = new SqlClientDiagnosticSourceListener(configuration);
+        }
+
+        [TestCleanup]
+        public void Cleanup()
+        {
+            this.sqlClientDiagnosticSourceListener.Dispose();
+            this.fakeSqlClientDiagnosticSource.Dispose();
+            this.configuration.Dispose();
+            this.stubTelemetryChannel.Dispose();
+        }
+
+        [TestMethod]
+        public void TracksCommandExecuted()
+        {
+            var operationId = Guid.NewGuid();
+            var sqlConnection = new SqlConnection(TestConnectionString);
+            var sqlCommand = sqlConnection.CreateCommand();
+            sqlCommand.CommandText = "select * from orders";
+            
+            var beforeExecuteEventData = new
+            {
+                OperationId = operationId,
+                Timestamp = 1000000L
+            };
+
+            this.fakeSqlClientDiagnosticSource.Write(
+                SqlClientDiagnosticSourceListener.SqlBeforeExecuteCommand,
+                beforeExecuteEventData);
+
+            var afterExecuteEventData = new
+            {
+                OperationId = operationId,
+                Operation = "ExecuteReader",
+                ConnectionId = sqlConnection.ClientConnectionId,
+                Command = sqlCommand,
+                Statistics = sqlConnection.RetrieveStatistics(),
+                Timestamp = 2000000L
+            };
+
+            var now = DateTimeOffset.UtcNow;
+
+            this.fakeSqlClientDiagnosticSource.Write(
+                SqlClientDiagnosticSourceListener.SqlAfterExecuteCommand,
+                afterExecuteEventData);
+
+            var dependencyTelemetry = (DependencyTelemetry)this.sendItems.Single();
+
+            Assert.AreEqual(afterExecuteEventData.OperationId.ToString("N"), dependencyTelemetry.Id);
+            Assert.AreEqual(sqlCommand.CommandText, dependencyTelemetry.Data);
+            Assert.AreEqual("(localdb)\\MSSQLLocalDB | master", dependencyTelemetry.Name);
+            Assert.AreEqual("(localdb)\\MSSQLLocalDB | master", dependencyTelemetry.Target);
+            Assert.AreEqual(RemoteDependencyConstants.SQL, dependencyTelemetry.Type);
+            Assert.IsTrue((bool)dependencyTelemetry.Success);
+            Assert.AreEqual(1000000L, dependencyTelemetry.Duration.Ticks);
+            Assert.IsTrue(dependencyTelemetry.Timestamp.Ticks < now.Ticks);
+        }
+        
+        [TestMethod]
+        public void TracksCommandError()
+        {
+            var operationId = Guid.NewGuid();
+            var sqlConnection = new SqlConnection(TestConnectionString);
+            var sqlCommand = sqlConnection.CreateCommand();
+            sqlCommand.CommandText = "select * from orders";
+
+            var beforeExecuteEventData = new
+            {
+                OperationId = operationId,
+                Timestamp = 1000000L
+            };
+
+            this.fakeSqlClientDiagnosticSource.Write(
+                SqlClientDiagnosticSourceListener.SqlBeforeExecuteCommand,
+                beforeExecuteEventData);
+
+            var commandErrorEventData = new
+            {
+                OperationId = operationId,
+                Operation = "ExecuteReader",
+                ConnectionId = sqlConnection.ClientConnectionId,
+                Command = sqlCommand,
+                Exception = new Exception("Boom!"),
+                Timestamp = 2000000L
+            };
+
+            var now = DateTimeOffset.UtcNow;
+
+            this.fakeSqlClientDiagnosticSource.Write(
+                SqlClientDiagnosticSourceListener.SqlErrorExecuteCommand,
+                commandErrorEventData);
+
+            var exceptionTelemetry = (ExceptionTelemetry)this.sendItems.Single();
+
+            Assert.AreSame(commandErrorEventData.Exception, exceptionTelemetry.Exception);
+            Assert.AreEqual(commandErrorEventData.Exception.Message, exceptionTelemetry.Message);
+            Assert.IsTrue(exceptionTelemetry.Timestamp.Ticks < now.Ticks);
+        }
+
+        [TestMethod]
+        public void TracksConnectionOpened()
+        {
+            var operationId = Guid.NewGuid();
+            var sqlConnection = new SqlConnection(TestConnectionString);
+
+            var beforeOpenEventData = new
+            {
+                OperationId = operationId,
+                Timestamp = 1000000L
+            };
+
+            this.fakeSqlClientDiagnosticSource.Write(
+                SqlClientDiagnosticSourceListener.SqlBeforeOpenConnection,
+                beforeOpenEventData);
+
+            var afterOpenEventData = new
+            {
+                OperationId = operationId,
+                Operation = "Open",
+                ConnectionId = sqlConnection.ClientConnectionId,
+                Connection = sqlConnection,
+                Statistics = sqlConnection.RetrieveStatistics(),
+                Timestamp = 2000000L
+            };
+
+            var now = DateTimeOffset.UtcNow;
+            
+            this.fakeSqlClientDiagnosticSource.Write(
+                SqlClientDiagnosticSourceListener.SqlAfterOpenConnection,
+                afterOpenEventData);
+
+            var dependencyTelemetry = (DependencyTelemetry)this.sendItems.Single();
+
+            Assert.AreEqual(afterOpenEventData.OperationId.ToString("N"), dependencyTelemetry.Id);
+            Assert.AreEqual(afterOpenEventData.Operation, dependencyTelemetry.Data);
+            Assert.AreEqual("(localdb)\\MSSQLLocalDB | master | " + afterOpenEventData.Operation, dependencyTelemetry.Name);
+            Assert.AreEqual("(localdb)\\MSSQLLocalDB | master", dependencyTelemetry.Target);
+            Assert.AreEqual(RemoteDependencyConstants.SQL, dependencyTelemetry.Type);
+            Assert.IsTrue((bool)dependencyTelemetry.Success);
+            Assert.AreEqual(1000000L, dependencyTelemetry.Duration.Ticks);
+            Assert.IsTrue(dependencyTelemetry.Timestamp.Ticks < now.Ticks);
+        }
+        
+        [TestMethod]
+        public void TracksConnectionOpenedError()
+        {
+            var operationId = Guid.NewGuid();
+            var sqlConnection = new SqlConnection(TestConnectionString);
+            
+            var beforeOpenEventData = new
+            {
+                OperationId = operationId,
+                Timestamp = 1000000L
+            };
+
+            this.fakeSqlClientDiagnosticSource.Write(
+                SqlClientDiagnosticSourceListener.SqlBeforeOpenConnection,
+                beforeOpenEventData);
+
+            var errorOpenEventData = new
+            {
+                OperationId = operationId,
+                Operation = "Open",
+                ConnectionId = sqlConnection.ClientConnectionId,
+                Connection = sqlConnection,
+                Exception = new Exception("Boom!"),
+                Timestamp = 2000000L
+            };
+
+            var now = DateTimeOffset.UtcNow;
+
+            this.fakeSqlClientDiagnosticSource.Write(
+                SqlClientDiagnosticSourceListener.SqlErrorOpenConnection,
+                errorOpenEventData);
+
+            var exceptionTelemetry = (ExceptionTelemetry)this.sendItems.Single();
+
+            Assert.AreSame(errorOpenEventData.Exception, exceptionTelemetry.Exception);
+            Assert.AreEqual(errorOpenEventData.Exception.Message, exceptionTelemetry.Message);
+            Assert.IsTrue(exceptionTelemetry.Timestamp.Ticks < now.Ticks);
+        }
+
+        [TestMethod]
+        public void TracksConnectionClosed()
+        {
+            var operationId = Guid.NewGuid();
+            var sqlConnection = new SqlConnection(TestConnectionString);
+
+            var beforeCloseEventData = new
+            {
+                OperationId = operationId,
+                Timestamp = 1000000L
+            };
+
+            this.fakeSqlClientDiagnosticSource.Write(
+                SqlClientDiagnosticSourceListener.SqlBeforeCloseConnection,
+                beforeCloseEventData);
+
+            var afterCloseEventData = new
+            {
+                OperationId = operationId,
+                Operation = "Close",
+                ConnectionId = sqlConnection.ClientConnectionId,
+                Connection = sqlConnection,
+                Statistics = sqlConnection.RetrieveStatistics(),
+                Timestamp = 2000000L
+            };
+
+            var now = DateTimeOffset.UtcNow;
+
+            this.fakeSqlClientDiagnosticSource.Write(
+                SqlClientDiagnosticSourceListener.SqlAfterCloseConnection,
+                afterCloseEventData);
+
+            var dependencyTelemetry = (DependencyTelemetry)this.sendItems.Single();
+
+            Assert.AreEqual(afterCloseEventData.OperationId.ToString("N"), dependencyTelemetry.Id);
+            Assert.AreEqual(afterCloseEventData.Operation, dependencyTelemetry.Data);
+            Assert.AreEqual("(localdb)\\MSSQLLocalDB | master | " + afterCloseEventData.Operation, dependencyTelemetry.Name);
+            Assert.AreEqual("(localdb)\\MSSQLLocalDB | master", dependencyTelemetry.Target);
+            Assert.AreEqual(RemoteDependencyConstants.SQL, dependencyTelemetry.Type);
+            Assert.IsTrue((bool)dependencyTelemetry.Success);
+            Assert.AreEqual(1000000L, dependencyTelemetry.Duration.Ticks);
+            Assert.IsTrue(dependencyTelemetry.Timestamp.Ticks < now.Ticks);
+        }
+
+        [TestMethod]
+        public void TracksConnectionCloseError()
+        {
+            var operationId = Guid.NewGuid();
+            var sqlConnection = new SqlConnection(TestConnectionString);
+
+            var beforeOpenEventData = new
+            {
+                OperationId = operationId,
+                Timestamp = 1000000L
+            };
+
+            this.fakeSqlClientDiagnosticSource.Write(
+                SqlClientDiagnosticSourceListener.SqlBeforeCloseConnection,
+                beforeOpenEventData);
+
+            var errorOpenEventData = new
+            {
+                OperationId = operationId,
+                Operation = "Close",
+                ConnectionId = sqlConnection.ClientConnectionId,
+                Connection = sqlConnection,
+                Exception = new Exception("Boom!"),
+                Timestamp = 2000000L
+            };
+
+            var now = DateTimeOffset.UtcNow;
+
+            this.fakeSqlClientDiagnosticSource.Write(
+                SqlClientDiagnosticSourceListener.SqlErrorCloseConnection,
+                errorOpenEventData);
+
+            var exceptionTelemetry = (ExceptionTelemetry)this.sendItems.Single();
+
+            Assert.AreSame(errorOpenEventData.Exception, exceptionTelemetry.Exception);
+            Assert.AreEqual(errorOpenEventData.Exception.Message, exceptionTelemetry.Message);
+            Assert.IsTrue(exceptionTelemetry.Timestamp.Ticks < now.Ticks);
+        }
+
+        [TestMethod]
+        public void TracksTransactionCommitted()
+        {
+            var operationId = Guid.NewGuid();
+            var sqlConnection = new SqlConnection(TestConnectionString);
+
+            var beforeCommitEventData = new
+            {
+                OperationId = operationId,
+                Timestamp = 1000000L
+            };
+
+            this.fakeSqlClientDiagnosticSource.Write(
+                SqlClientDiagnosticSourceListener.SqlBeforeCommitTransaction,
+                beforeCommitEventData);
+
+            var afterCommitEventData = new
+            {
+                OperationId = operationId,
+                Operation = "Commit",
+                IsolationLevel = IsolationLevel.Snapshot,
+                Connection = sqlConnection,
+                Timestamp = 2000000L
+            };
+
+            var now = DateTimeOffset.UtcNow;
+
+            this.fakeSqlClientDiagnosticSource.Write(
+                SqlClientDiagnosticSourceListener.SqlAfterCommitTransaction,
+                afterCommitEventData);
+
+            var dependencyTelemetry = (DependencyTelemetry)this.sendItems.Single();
+
+            Assert.AreEqual(afterCommitEventData.OperationId.ToString("N"), dependencyTelemetry.Id);
+            Assert.AreEqual(afterCommitEventData.Operation, dependencyTelemetry.Data);
+            Assert.AreEqual(
+                "(localdb)\\MSSQLLocalDB | master | " + afterCommitEventData.Operation + " | " + afterCommitEventData.IsolationLevel, 
+                dependencyTelemetry.Name);
+            Assert.AreEqual("(localdb)\\MSSQLLocalDB | master", dependencyTelemetry.Target);
+            Assert.AreEqual(RemoteDependencyConstants.SQL, dependencyTelemetry.Type);
+            Assert.IsTrue((bool)dependencyTelemetry.Success);
+            Assert.AreEqual(1000000L, dependencyTelemetry.Duration.Ticks);
+            Assert.IsTrue(dependencyTelemetry.Timestamp.Ticks < now.Ticks);
+        }
+
+        [TestMethod]
+        public void TracksTransactionCommitError()
+        {
+            var operationId = Guid.NewGuid();
+            var sqlConnection = new SqlConnection(TestConnectionString);
+
+            var beforeCommitEventData = new
+            {
+                OperationId = operationId,
+                Timestamp = 1000000L
+            };
+
+            this.fakeSqlClientDiagnosticSource.Write(
+                SqlClientDiagnosticSourceListener.SqlBeforeCommitTransaction,
+                beforeCommitEventData);
+
+            var errorCommitEventData = new
+            {
+                OperationId = operationId,
+                Operation = "Commit",
+                IsolationLevel = IsolationLevel.Snapshot,
+                Connection = sqlConnection,
+                Exception = new Exception("Boom!"),
+                Timestamp = 2000000L
+            };
+
+            var now = DateTimeOffset.UtcNow;
+
+            this.fakeSqlClientDiagnosticSource.Write(
+                SqlClientDiagnosticSourceListener.SqlErrorCommitTransaction,
+                errorCommitEventData);
+
+            var exceptionTelemetry = (ExceptionTelemetry)this.sendItems.Single();
+
+            Assert.AreSame(errorCommitEventData.Exception, exceptionTelemetry.Exception);
+            Assert.AreEqual(errorCommitEventData.Exception.Message, exceptionTelemetry.Message);
+            Assert.IsTrue(exceptionTelemetry.Timestamp.Ticks < now.Ticks);
+        }
+
+        [TestMethod]
+        public void TracksTransactionRolledBack()
+        {
+            var operationId = Guid.NewGuid();
+            var sqlConnection = new SqlConnection(TestConnectionString);
+
+            var beforeRollbackEventData = new
+            {
+                OperationId = operationId,
+                Timestamp = 1000000L
+            };
+
+            this.fakeSqlClientDiagnosticSource.Write(
+                SqlClientDiagnosticSourceListener.SqlBeforeRollbackTransaction,
+                beforeRollbackEventData);
+
+            var afterRollbackEventData = new
+            {
+                OperationId = operationId,
+                Operation = "Rollback",
+                IsolationLevel = IsolationLevel.Snapshot,
+                Connection = sqlConnection,
+                Timestamp = 2000000L
+            };
+
+            var now = DateTimeOffset.UtcNow;
+
+            this.fakeSqlClientDiagnosticSource.Write(
+                SqlClientDiagnosticSourceListener.SqlAfterRollbackTransaction,
+                afterRollbackEventData);
+
+            var dependencyTelemetry = (DependencyTelemetry)this.sendItems.Single();
+
+            Assert.AreEqual(afterRollbackEventData.OperationId.ToString("N"), dependencyTelemetry.Id);
+            Assert.AreEqual(afterRollbackEventData.Operation, dependencyTelemetry.Data);
+            Assert.AreEqual(
+                "(localdb)\\MSSQLLocalDB | master | " + afterRollbackEventData.Operation + " | " + afterRollbackEventData.IsolationLevel,
+                dependencyTelemetry.Name);
+            Assert.AreEqual("(localdb)\\MSSQLLocalDB | master", dependencyTelemetry.Target);
+            Assert.AreEqual(RemoteDependencyConstants.SQL, dependencyTelemetry.Type);
+            Assert.IsTrue((bool)dependencyTelemetry.Success);
+            Assert.AreEqual(1000000L, dependencyTelemetry.Duration.Ticks);
+            Assert.IsTrue(dependencyTelemetry.Timestamp.Ticks < now.Ticks);
+        }
+
+        private class FakeSqlClientDiagnosticSource : IDisposable
+        {
+            private readonly DiagnosticListener listener;
+
+            public FakeSqlClientDiagnosticSource()
+            {
+                this.listener = new DiagnosticListener(SqlClientDiagnosticSourceListener.DiagnosticListenerName);
+            }
+
+            public void Write(string name, object value)
+            {
+                this.listener.Write(name, value);
+            }
+
+            public void Dispose()
+            {
+                this.listener.Dispose();
+            }
+        }
+    }
+
+#pragma warning restore CS1591 // Missing XML comment for publicly visible type or member
+}
--- a/Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
+++ b/Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
@@ -190,50 +190,6 @@ namespace Microsoft.ApplicationInsights.Tests
             Assert.AreEqual(commandErrorEventData.Exception.ToInvariantString(), dependencyTelemetry.Properties["Exception"]);
             Assert.IsFalse(dependencyTelemetry.Success.Value);
         }
-
-        [TestMethod]
-        public void TracksConnectionOpened()
-        {
-            var operationId = Guid.NewGuid();
-            var sqlConnection = new SqlConnection(TestConnectionString);
-
-            var beforeOpenEventData = new
-            {
-                OperationId = operationId,
-                Timestamp = 1000000L
-            };
-
-            this.fakeSqlClientDiagnosticSource.Write(
-                SqlClientDiagnosticSourceListener.SqlBeforeOpenConnection,
-                beforeOpenEventData);
-
-            var afterOpenEventData = new
-            {
-                OperationId = operationId,
-                Operation = "Open",
-                ConnectionId = sqlConnection.ClientConnectionId,
-                Connection = sqlConnection,
-                Statistics = sqlConnection.RetrieveStatistics(),
-                Timestamp = 2000000L
-            };
-
-            var now = DateTimeOffset.UtcNow;
-            
-            this.fakeSqlClientDiagnosticSource.Write(
-                SqlClientDiagnosticSourceListener.SqlAfterOpenConnection,
-                afterOpenEventData);
-
-            var dependencyTelemetry = (DependencyTelemetry)this.sendItems.Single();
-
-            Assert.AreEqual(afterOpenEventData.OperationId.ToString("N"), dependencyTelemetry.Id);
-            Assert.AreEqual(afterOpenEventData.Operation, dependencyTelemetry.Data);
-            Assert.AreEqual("(localdb)\\MSSQLLocalDB | master | " + afterOpenEventData.Operation, dependencyTelemetry.Name);
-            Assert.AreEqual("(localdb)\\MSSQLLocalDB | master", dependencyTelemetry.Target);
-            Assert.AreEqual(RemoteDependencyConstants.SQL, dependencyTelemetry.Type);
-            Assert.IsTrue((bool)dependencyTelemetry.Success);
-            Assert.AreEqual(1000000L, dependencyTelemetry.Duration.Ticks);
-            Assert.IsTrue(dependencyTelemetry.Timestamp.Ticks < now.Ticks);
-        }
         
         [TestMethod]
         public void TracksConnectionOpenedError()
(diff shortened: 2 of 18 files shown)


anpete:
cc SergeyKanzhelev. Please add any other relevant folks.

cijothomas:
@anpete - Can you open GH issue which describes this enhancement ? It'll be much easier to look at PR once the necessary context is already understood!
Thanks.

anpete:
@cijothomas  I added the existing issue to the description.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,149 @@
...
+
+            // TODO: Unify with ProfilerSqlConnectionProcessing?
+            telemetry.Name = string.Join(" | ", connection.DataSource, connection.Database, SqlConnectionCommandText);
+            telemetry.Type = RemoteDependencyConstants.SQL;
+            telemetry.Target = string.Join(" | ", connection.DataSource, connection.Database);
+            telemetry.Data = SqlConnectionCommandText;
+            telemetry.Timestamp = timestamp;
+            telemetry.Duration = duration;
+
+            this.client.Track(telemetry);
```
cijothomas: Do we get callback if SQL open() failed? We need to set telemetry.Sucess field based on if the call was sucess/failed.

lmolkova:
I have some comments, I'll add them later today. Please do not merge yet

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: Yes. Once we are happy with the general approach I will implement the rest. Take a look [here](https://github.com/dotnet/corefx/blob/master/src/System.Data.SqlClient/src/System/Data/SqlClient/SqlClientDiagnosticListenerExtensions.cs)for what is available.

anpete:
@lmolkova Thanks! Not planning to merge anytime soon (I can't anyway as I don't have rights 😄).

lmolkova:
@anpete 

 I've done changes in HttpClient instrumentation and participated in Activity API/DiagnosticSource implementation as well as supporting it across AppInsights SDK, so I hope I could be helpful here too. 
I don't have enough context on how EF works and the instrumentation, but I can provide some details on how to work with Activities in AI. I assume there are other events you want to support and I'd like to understand the scope of this work. It would be great to meet and discuss it offline.

I know there is some ongoing work/planning in EF Core to change the instrumentation: write different events and support Activity. Are you aware about it? I don't have much details but could find out. I assume that there will be some backward compatibility, so supporting current events will probably still work for a while anyway. @SergeyKanzhelev, @AlexBulankou (JFYI)

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        public const string SqlConnectionCommandText = "Open";
+
+        private void OnAfterOpenConnection(SqlConnection connection, DateTimeOffset timestamp, TimeSpan duration)
+        {
+            var telemetry = new DependencyTelemetry();
+
+            // TODO: Activities?
+
+            this.client.Initialize(telemetry); // TODO: Call this?
```
lmolkova: You don't need to call Initialize here as it will be called anyway in Track method right after it. 
In HTTP listener it's done for other reasons.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,149 @@
...
+            }
+        }
+
+        public const string SqlConnectionCommandText = "Open";
+
+        private void OnAfterOpenConnection(SqlConnection connection, DateTimeOffset timestamp, TimeSpan duration)
+        {
+            var telemetry = new DependencyTelemetry();
+
+            // TODO: Activities?
```
lmolkova: Since it's a 'After' event it probably does not make sense to create Activity here. Activity represents some operation: it's context and duration. And it makes sense if there is something interesting happens between start and stop.
However I probably miss some context on the EF and it's instrumentation and would like to have more info :)

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,149 @@
...
+
+            internal SqlClientDiagnosticSourceSubscriber(SqlClientDiagnosticSourceListener listener)
+            {
+                this.httpDiagnosticListener = listener;
+
+                try
+                {
+                    this.listenerSubscription = DiagnosticListener.AllListeners.Subscribe(this);
+                }
+                catch (Exception ex)
```
lmolkova: I believe we have added this try/catch in Http listener to workaround issue that is resolved now. Not sure if it's needed, as `DiagnosticListener.AllListeners.Subscribe` nust never throw. But I would let @cijothomas to decide

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,149 @@
...
+        {
+            if (this.subscriber != null)
+            {
+                this.subscriber.Dispose();
+            }
+        }
+
+        private sealed class SqlClientDiagnosticSourceSubscriber : IObserver<DiagnosticListener>, IDisposable
+        {
+            private readonly SqlClientDiagnosticSourceListener httpDiagnosticListener;
```
lmolkova: sqlDiagnosticListener?

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
```diff
@@ -124,6 +126,8 @@ public void Initialize(TelemetryConfiguration configuration)
                             // Net40 only supports runtime instrumentation
                             // Net45 supports either but not both to avoid duplication
                             this.InitializeForRuntimeInstrumentationOrFramework();
+#else
+                            this.sqlClientDiagnosticSourceListener = new SqlClientDiagnosticSourceListener(configuration);
```
lmolkova: please dispose it

lmolkova: We have previously discovered that even if application targets .NET Full Framework, it may install some package that targets netstandard1.*. That will bring the whole .NET Core stack into such application. In this case SqlClient may be .NET Core SqlClient.

For HTTP we decided to enable listener on NET45 as well (see `httpCoreDiagnosticSourceListener` initialization). It costs nothing to have no-op listener.

anpete:
cc @divega

anpete:
@lmolkova New version up.

cc @divega @ajcvickers

Src/DependencyCollector/NetCore/DependencyCollector.NetCore.csproj
```diff
@@ -66,6 +67,7 @@
     <PackageReference Include="Microsoft.Extensions.PlatformAbstractions" Version="1.1.0" />
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.4.0-preview2-25330-01" />
     <PackageReference Include="System.Diagnostics.StackTrace" Version="4.3.0" />
+    <PackageReference Include="System.Data.SqlClient" Version="4.3.1" />
```
lmolkova: please also DependencyCollector nuspec for netstandard1.6.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,488 @@
...
+        public const string SqlBeforeRollbackTransaction = SqlClientPrefix + "WriteTransactionRollbackBefore";
+        public const string SqlAfterRollbackTransaction = SqlClientPrefix + "WriteTransactionRollbackAfter";
+        public const string SqlErrorRollbackTransaction = SqlClientPrefix + "WriteTransactionRollbackError";
+
+        private const string SqlClientPrefix = "System.Data.SqlClient.";
+
+        private readonly TelemetryClient client;
+        private readonly SqlClientDiagnosticSourceSubscriber subscriber;
+
+        private readonly ConcurrentDictionary<Guid, long> operationStartTimestamps = new ConcurrentDictionary<Guid, long>();
```
lmolkova: are there any cases including faulty ones when items will be added, but not removed: exceptions, DB unavailability, anything?

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,488 @@
...
+        private void OnCommandErrorEvent(
+            Guid operationId,
+            string operation,
+            Guid connectionId,
+            Exception exception,
+            long endTimestamp)
+        {
+            var duration = this.DeriveDuration(operationId, endTimestamp);
+
+            // TODO: How should this be correlated to the parent dependency event?
```
lmolkova: ApplicationInsights will set telemetry.Operation.Context.Id and ParentId when telemetry is initialized.

However, since there was no Activity started for parent dependency, exception will be a sibling to the dependency.
You have some operationId in the event. Is it id of the query/transaction?
If so, you can just set telemetry.Operation.Context.Parent.Id to it and it will be correlated to the parent dependency.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: Nothing obvious. SqlClient is will either generate an After event or an Error event (inside a finally block), both of which will cause us to remove the entry. 

Doing this correlation is ugly. I think we could remove if SqlClient used Activities, correct?

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: Happy either way, just LMK.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,521 @@
...
+                        (SqlConnection)TransactionError.Connection.Fetch(evnt.Value),
+                        (Exception)TransactionError.Exception.Fetch(evnt.Value),
+                        (long)TransactionError.Timestamp.Fetch(evnt.Value));
+
+                    break;
+                }
+            }
+        }
+
+        private static void InitializeTelemetry(ITelemetry telemetry, Guid operationId)
```
anpete: @lmolkova Please check this. I may well have misunderstood your comments on error event correlation. Thanks!

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,521 @@
...
+                var commandName = command.CommandType == CommandType.StoredProcedure
+                    ? command.CommandText
+                    : string.Empty;
+
+                dependencyName = string.IsNullOrEmpty(commandName)
+                    ? string.Join(" | ", command.Connection.DataSource, command.Connection.Database)
+                    : string.Join(" | ", command.Connection.DataSource, command.Connection.Database, commandName);
+            }
+
+            var duration = this.DeriveDuration(operationId, endTimestamp);
```
Dmitry-Matveev: This seems to be the only place where `DeriveDuration()` call and subsequently `TryRemove()` on the `OperationStartTimestamps` collection is not in the beginning of the method. If some code that may throw is ever added on top, we may not reach `DeriveDuration()` and `OperationStartTimestamps` may keep growing.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
Dmitry-Matveev: In addition to Liudmila's question: is there any throttling on the diagnostic source so that we may not receive `Error` or `After` and keep increasing memory taken by `operationStartTimestamps`?

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,521 @@
...
+                Timestamp = DateTimeOffset.UtcNow - duration,
+                Success = true
+            };
+
+            InitializeTelemetry(telemetry, operationId);
+
+            this.client.Track(telemetry);
+        }
+
+        private void OnErrorConnectionEvent(
```
Dmitry-Matveev: Minor/Curious: Is there a place in EF source that may produce both `After` and `Error` for the same transaction? E.g. if `After` is sent from the `finally{}` block?

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
Dmitry-Matveev: Based on my history with "must never throw", the thing eventually throws one way or the other (e.g. new code in DS will be added by someone who does not know it should not throw and SDK dependency on DS is currently ">X" so new one will be picked up automatically) :).
I would err on a side of logging the error and gracefully degrade.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,521 @@
...
+        {
+            var activity = Activity.Current;
+
+            if (activity != null)
+            {
+                telemetry.Context.Operation.Id = activity.RootId;
+            }
+            else
+            {
+                telemetry.Context.Operation.Id = Convert.ToString(operationId, CultureInfo.InvariantCulture);
```
pharring: Convert.ToString with a GUID will box the GUID. Instead, simply do `operationId.ToString()`.
If you want to save a few bytes, then do `operationId.ToString("N")` which removes the dashes.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,521 @@
...
+            }
+        }
+
+        private static void CorrelateErrorTelemetry(ExceptionTelemetry telemetry, Guid operationId)
+        {
+            telemetry.Context.Operation.ParentId = Convert.ToString(operationId, CultureInfo.InvariantCulture);
```
pharring: `operationId.ToString("N")`

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
pharring: A way to deal with this is to have two tables (primary and secondary). For look-ups, check both tables (primary first, then secondary). For inserts: Use only the primary until it reaches a maximum capacity. When the primary is 'full', clear the secondary and swap it with the primary.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,521 @@
...
+                dependencyName = string.IsNullOrEmpty(commandName)
+                    ? string.Join(" | ", command.Connection.DataSource, command.Connection.Database)
+                    : string.Join(" | ", command.Connection.DataSource, command.Connection.Database, commandName);
+            }
+
+            var duration = this.DeriveDuration(operationId, endTimestamp);
+
+            var telemetry = new DependencyTelemetry()
+            {
+                Id = Convert.ToString(operationId, CultureInfo.InvariantCulture),
```
pharring: `operationId.ToString("N")`

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,521 @@
...
+                ? TimeSpan.FromTicks(endTimestamp - startTimestamp1)
+                : default(TimeSpan);
+        }
+        
+        #region Fetchers
+
+        // Fetchers for execute command before event
+        private static class CommandBefore
+        {
+            public static readonly PropertyFetcher OperationId = new PropertyFetcher("OperationId");
```
pharring: Suggest: `nameof(OperationId)`

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: Shouldn't be. [Here](https://github.com/dotnet/corefx/blob/83b606b5debad20536a581c425f66f4e9c396ed1/src/System.Data.SqlClient/src/System/Data/SqlClient/SqlCommand.cs#L809)is an example of the usage pattern they employ in SqlClient.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: In this case it seems fine just to move the call to the top of the method and add a comment.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: Do you mean on the producer side? If so, no, see [here](https://github.com/dotnet/corefx/blob/83b606b5debad20536a581c425f66f4e9c396ed1/src/System.Data.SqlClient/src/System/Data/SqlClient/SqlCommand.cs#L786) for an example of the publication pattern used by SqlClient

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,522 @@
...
+            }
+        }
+
+        private static void InitializeTelemetry(ITelemetry telemetry, Guid operationId)
+        {
+            var activity = Activity.Current;
+
+            if (activity != null)
+            {
+                telemetry.Context.Operation.Id = activity.RootId;
```
lmolkova: please also set telemetry.Context.Operation.ParentId to activity.ParentId

lmolkova: Please also add activity.Baggage to the telemetry.Context.Operation.Properties like in [HttpCoreDiagnosticSourceListener](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs#L235)

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
lmolkova: sure. perhaps if someone will add bug/code that throws, it would be other places that we do not expect to throw :)
Anyway, let's keep it for now if you'd like

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,522 @@
...
+
+        private void OnAfterCommandEvent(
+            Guid operationId,
+            string operation,
+            Guid connectionId,
+            SqlCommand command,
+            IDictionary statistics,
+            long endTimestamp)
+        {
+            // Call DeriveDuration first to ensure we remove any start timestamp from operationStartTimestamps
```
lmolkova: Could you please also add EventSource log for every callback with the message containing event name and operationId.
It helps a lot to debug things. 
@cijothomas ;)

anpete:
@lmolkova I think everything is addressed now. Let me know it there is anything else.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,565 @@
...
+        {
+        }
+
+        void IObserver<KeyValuePair<string, object>>.OnError(Exception error)
+        {
+        }
+
+        void IObserver<KeyValuePair<string, object>>.OnNext(KeyValuePair<string, object> evnt)
+        {
+            switch (evnt.Key)
```
cijothomas: do we need to wrap this method inside try..catch ?

anpete: That is not the case in HttpCoreDiagnosticSourceListener, which is what I based this impl. on.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,565 @@
...
+        private void OnCommandErrorEvent(
+            Guid operationId,
+            string operation,
+            Guid connectionId,
+            Exception exception,
+            long endTimestamp)
+        {
+            var duration = this.DeriveDuration(operationId, endTimestamp);
+
+            var telemetry = new ExceptionTelemetry(exception)
```
cijothomas: why there is no DependencyTelemetry (with success=false) ? This behavior is different from current SQL dependency monitoring.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,565 @@
...
+                case SqlAfterOpenConnection:
+                case SqlAfterCloseConnection:
+                {
+                    var operationId = (Guid)ConnectionAfter.OperationId.Fetch(evnt.Value);
+
+                    DependencyCollectorEventSource.Log.SqlClientDiagnosticSubscriberCallbackCalled(operationId, evnt.Key);
+                    
+                    this.OnAfterConnectionEvent(
+                        operationId,
+                        (string)ConnectionAfter.Operation.Fetch(evnt.Value),
```
cijothomas: what will be the content of this string? (operation)
Current profiler based sql dependencies use a fixed, hard-coded string 'Open'. Is this operation going to be same as 'Open' ?

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,565 @@
...
+            var telemetry = new DependencyTelemetry()
+            {
+                Id = operationId.ToString("N"),
+                Name = string.Join(" | ", connection.DataSource, connection.Database, operation),
+                Type = RemoteDependencyConstants.SQL,
+                Target = string.Join(" | ", connection.DataSource, connection.Database),
+                Data = operation,
+                Duration = duration,
+                Timestamp = DateTimeOffset.UtcNow - duration,
+                Success = true
```
cijothomas: Need to sync with @iusafaro . Current dependency for SQL monitors Open() calls only if the call is failing. This seem to be collecting both success/failure.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,565 @@
...
+
+        private void OnAfterConnectionEvent(
+            Guid operationId, 
+            string operation,
+            Guid connectionId,
+            SqlConnection connection, 
+            IDictionary statistics, 
+            long endTimestamp)
+        {
+            var duration = this.DeriveDuration(operationId, endTimestamp);
```
cijothomas: how precise are the duration? Does it have StopWatch precision? Open() is typically very fast and we need to ensure our precision can handle such fast calls.

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
```diff
@@ -31,12 +31,13 @@ public class DependencyTrackingTelemetryModule : ITelemetryModule, IDisposable
 
 #if NET45 || NETCORE
         private HttpCoreDiagnosticSourceListener httpCoreDiagnosticSourceListener;
+        private SqlClientDiagnosticSourceListener sqlClientDiagnosticSourceListener;
```
cijothomas: is SqlClientDiagnosticSourceListener present in .net core as well as full .net? I could not find any logic to prevent collision/duplication when Profiler is present as well.

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
```diff
@@ -134,6 +135,8 @@ public void Initialize(TelemetryConfiguration configuration)
                                 this.SetComponentCorrelationHttpHeaders,
                                 this.ExcludeComponentCorrelationHttpHeadersOnDomains, 
                                 null);
+
+                            this.sqlClientDiagnosticSourceListener = new SqlClientDiagnosticSourceListener(configuration);
```
cijothomas: if this is present in NET45/46 etc, then this code should be moved to InitializeForRuntimeInstrumentationOrFramework() method. Specifically - this should be enabled only if Profiler is determined to be NOT installed.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
cijothomas: All the other dependency collectors use some data structure like this to correlate between begin/end calls, but all of them have some protections against this growing infinitely (in case of bugs elsewhere..). We need to add some protection here as well..
Specifically - for CacheBasedOperationHolder. the cache has a expiration policy to ensure items gets removed after some time even if no matching End is called.
Similarly, ObjectInstanceBasedOperationHolder uses ConditionalWeakTable to ensure this dictionary won't grow infinitely. It is risky to assume that the external system (diagnostic source) always calls END.

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
anpete: If you mean, does full .net SQL Client publish DiagnosticSource events, then, no, it doesn't. This was enabled for both per feedback from @lmolkova

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
anpete: See comment above.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: SQL Client uses Stopwatch.GetTimestamp()

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: Values can be ["Open", "OpenAsync", "Close"]

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: Should that be instead of or as well as the exception telemetry?

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,565 @@
...
+            }
+            else
+            {
+                telemetry.Context.Operation.Id = operationId.ToString("N");
+            }
+        }
+
+        private static void CorrelateErrorTelemetry(ExceptionTelemetry telemetry, Guid operationId)
+        {
+            telemetry.Context.Operation.ParentId = operationId.ToString("N");
```
cijothomas: Couldn't find any unit tests for this ?

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,565 @@
...
+                        (SqlConnection)TransactionError.Connection.Fetch(evnt.Value),
+                        (Exception)TransactionError.Exception.Fetch(evnt.Value),
+                        (long)TransactionError.Timestamp.Fetch(evnt.Value));
+
+                    break;
+                }
+            }
+        }
+
+        private static void InitializeTelemetry(ITelemetry telemetry, Guid operationId)
```
cijothomas: could not find any tests for this?

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
cijothomas: Current dependency collectors only send DependencyTelemetry (success=false), and don't send any exception telemetry. This is major change and needs discussion with other stakeholders. We should follow the existing standards (i.e report DependencyTelemetry), and maybe add the exception code into the dependency telemetry itself, rather than sending a new ExceptionTelemetry. @SergeyKanzhelev

anpete: Happy to change it.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: I understand. However, in this case, we own, and have carefully reviewed the publication code and it seems impossible for it to fail to publish the end calls. But, perhaps we have missed something - Can you please show me, in the specific code I have linked to above, how we can hit this issue?

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: @iusafaro Please confirm that we do not want to track successful connection open calls. I will change it if that's the case.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
cijothomas: For all the cases, we own and reviewed code which produce call backs. (framework/ profiler).
The intention is not to find any bugs in the external code, but rather simply protecting SDK from any potential bugs. Just being on safe side.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: @cijothomas Just to confirm, I should still go ahead and change this to use DependencyTelemetry (success = false)?

cc @SergeyKanzhelev

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: Added

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
cijothomas: Yes. DependencyTelemetry (success = false), just like current modules do.

anpete: Added a commit that removes exception telemetry.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
cijothomas: Right. For SQLConnection these are the methods instrumented currently.
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared/Implementation/ProfilerRuntimeInstrumentation.cs#L318

And we dont track successfull calls.

anpete: Will remove. It's interesting that you don't want to track this as it is useful data - Specifically, knowing how long it took to open the connection.

anpete: What about connection close?

cijothomas: No close() either. Follow
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared/Implementation/ProfilerRuntimeInstrumentation.cs#L318

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,539 @@
...
+
+            var telemetry = new DependencyTelemetry()
+            {
+                Id = operationId.ToString("N"),
+                Name = string.Join(" | ", connection.DataSource, connection.Database, operation),
+                Type = RemoteDependencyConstants.SQL,
+                Target = string.Join(" | ", connection.DataSource, connection.Database),
+                Data = operation,
+                Duration = duration,
+                Timestamp = DateTimeOffset.UtcNow - duration,
```
cijothomas: We anyway store the precise timestamp when operation started, so we should just use that, instead of deriving, which wont be as accurate. (There is a UI showing dependency timeline, which could get affected if start time is not as precise as possible, esp. for fast calls.)

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,539 @@
...
+
+            var telemetry = new DependencyTelemetry()
+            {
+                Id = operationId.ToString("N"),
+                Name = dependencyName,
+                Type = RemoteDependencyConstants.SQL,
+                Target = target,
+                Data = command.CommandText,
+                Duration = duration,
+                Timestamp = DateTimeOffset.UtcNow - duration,
```
cijothomas: We anyway store the precise timestamp when operation started, so we should just use that, instead of deriving, which wont be as accurate. (There is a UI showing dependency timeline, which could get affected if start time is not as precise as possible, esp. for fast calls.)

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
cijothomas: What will happen if an exception occurs anywhere in the code here? Since there is no handling, will it propagate to customer code? 
@lmolkova do you know?

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,539 @@
...
+                        (Guid)CommandError.ConnectionId.Fetch(evnt.Value),
+                        (SqlCommand)CommandError.Command.Fetch(evnt.Value),
+                        (Exception)CommandError.Exception.Fetch(evnt.Value),
+                        (long)CommandError.Timestamp.Fetch(evnt.Value));
+
+                    break;
+                }
+                    
+                case SqlBeforeOpenConnection:
+                case SqlBeforeCloseConnection:
```
cijothomas: is there a 'After' close/open here? If not, who will remove the entry from the operationStartTimestamps dictionary?

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,539 @@
...
+                    
+                    this.operationStartTimestamps.TryAdd(
+                        operationId,
+                        (long)ConnectionBefore.Timestamp.Fetch(evnt.Value));
+
+                    break;
+                }
+
+                case SqlErrorOpenConnection:
+                case SqlErrorCloseConnection:
```
cijothomas: As mentioned in another comment, for consistency with existing module, don't track Close() call.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
cijothomas: As mentioned in another comment, for consistency with existing module, don't track Close() call.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,539 @@
...
+
+        private DependencyTelemetry CreateTransactionTelemetry(
+            Guid operationId, string operation, IsolationLevel isolationLevel, SqlConnection connection, long endTimestamp)
+        {
+            var duration = this.DeriveDuration(operationId, endTimestamp);
+
+            var telemetry = new DependencyTelemetry()
+            {
+                Id = operationId.ToString("N"),
+                Name = string.Join(" | ", connection.DataSource, connection.Database, operation, isolationLevel),
```
cijothomas: what would be the name of the operation here?

Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
```diff
@@ -0,0 +1,434 @@
...
+            var dependencyTelemetry = (DependencyTelemetry)this.sendItems.Single();
+
+            Assert.AreEqual(afterExecuteEventData.OperationId.ToString("N"), dependencyTelemetry.Id);
+            Assert.AreEqual(sqlCommand.CommandText, dependencyTelemetry.Data);
+            Assert.AreEqual("(localdb)\\MSSQLLocalDB | master", dependencyTelemetry.Name);
+            Assert.AreEqual("(localdb)\\MSSQLLocalDB | master", dependencyTelemetry.Target);
+            Assert.AreEqual(RemoteDependencyConstants.SQL, dependencyTelemetry.Type);
+            Assert.IsTrue((bool)dependencyTelemetry.Success);
+            Assert.AreEqual(1000000L, dependencyTelemetry.Duration.Ticks);
+            Assert.IsTrue(dependencyTelemetry.Timestamp.Ticks < now.Ticks);
```
cijothomas: The timestamp should be exactly the timestamp set inside beforeExecuteEventData.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,539 @@
...
+        private static void ConfigureExceptionTelemetry(DependencyTelemetry telemetry, Exception exception)
+        {
+            telemetry.Success = false;
+            telemetry.Properties["Exception"] = exception.ToInvariantString();
+
+            var sqlException = exception as SqlException;
+
+            if (sqlException != null)
+            {
+                telemetry.ResultCode = sqlException.Number.ToString(CultureInfo.InvariantCulture);
```
cijothomas: is there unit test covering this block?

Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
```diff
@@ -0,0 +1,434 @@
...
+                SqlClientDiagnosticSourceListener.SqlBeforeExecuteCommand,
+                beforeExecuteEventData);
+
+            var commandErrorEventData = new
+            {
+                OperationId = operationId,
+                Operation = "ExecuteReader",
+                ConnectionId = sqlConnection.ClientConnectionId,
+                Command = sqlCommand,
+                Exception = new Exception("Boom!"),
```
cijothomas: add test with SqlException, and verify that we extract information from it.

Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
```diff
@@ -0,0 +1,434 @@
...
+                commandErrorEventData);
+
+            var dependencyTelemetry = (DependencyTelemetry)this.sendItems.Single();
+
+            Assert.AreEqual(commandErrorEventData.Exception.ToInvariantString(), dependencyTelemetry.Properties["Exception"]);
+            Assert.IsFalse(dependencyTelemetry.Success.Value);
+        }
+        
+        [TestMethod]
+        public void TracksConnectionOpenedError()
```
cijothomas: Since we are NOT supposed to track Successfull connections, add a test to validate we don't track sucessfull Open.

Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
```diff
@@ -0,0 +1,434 @@
...
+            Assert.AreEqual("(localdb)\\MSSQLLocalDB | master", dependencyTelemetry.Target);
+            Assert.AreEqual(RemoteDependencyConstants.SQL, dependencyTelemetry.Type);
+            Assert.AreEqual(1000000L, dependencyTelemetry.Duration.Ticks);
+            Assert.IsTrue(dependencyTelemetry.Timestamp.Ticks < now.Ticks);
+            Assert.AreEqual(errorOpenEventData.Exception.ToInvariantString(), dependencyTelemetry.Properties["Exception"]);
+            Assert.IsFalse(dependencyTelemetry.Success.Value);
+        }
+
+        [TestMethod]
+        public void TracksConnectionCloseError()
```
cijothomas: Close() is not to be tracked, so this can be removed

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
cijothomas: @SergeyKanzhelev / @Dmitry-Matveev - please share your thoughts about the Dictionary usage here. I am not sure if its safe to assume callbacks would always come and dictionary never grows indefinitely. All other DependencyModules have some safety mechanism to clean Dictionary if something misbehaves.

Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
```diff
@@ -0,0 +1,434 @@
...
+        {
+            var operationId = Guid.NewGuid();
+            var sqlConnection = new SqlConnection(TestConnectionString);
+            var sqlCommand = sqlConnection.CreateCommand();
+            sqlCommand.CommandText = "select * from orders";
+            
+            var beforeExecuteEventData = new
+            {
+                OperationId = operationId,
+                Timestamp = 1000000L
```
cijothomas: need separate test cases one with Timestamp set, and one with Timestamp not set.
(corefx#20748)

Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
anpete: Done.

Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
anpete: Done.

Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
anpete: We can't because the timestamp in beforeExecuteEventData is the ticks returned by Stopwatch.GetTimestamp() whereas the DependencyTelemetry.Timestamp is DateTimeOffset. This is why we compute the value by subtracting duration from DateTimeOffset.UtcNow.

Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
anpete: Done.

Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
anpete: Done.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: We are not storing the start timestamp, but the ticks value returned by Stopwatch.GetTimestamp().

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: We are not storing the start timestamp, but the ticks value returned by Stopwatch.GetTimestamp().

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: Added

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: Done.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: Good catch. I restored the the dictionary removal for After open. After close is fine because we removed Before close.

anpete: Done.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: "Commit" or "Rollback"

Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
anpete: We are not storing the start timestamp, but the ticks value returned by Stopwatch.GetTimestamp().

Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
anpete: Done.

Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
anpete: Done.

Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
anpete: Done.

Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
anpete: Done.

Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
cijothomas: ok got it

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
cijothomas: ok

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
cijothomas: ok

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
cijothomas: ok

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
Dmitry-Matveev: I would agree with Cijo on this one.

In the sample of SqlClient provided above, if [Statistics](https://github.com/dotnet/corefx/blob/83b606b5debad20536a581c425f66f4e9c396ed1/src/System.Data.SqlClient/src/System/Data/SqlClient/SqlCommand.cs#L811) call fails by any reason (e.g. NullRef on "." inside that method and further down the callstack), no End or Error is getting logged. Additionally, in the current implementation, `TryRemove` is not called in `finally`, so it's prone to the issues with the module initialization and runtime troubles where the code before it may fail.

I would also say that no cleanup adds a dependency on the internal implementation of the another module (our current design seem to rely on that dependency present state and the fact that it's known to us). This adds a bond between two modules that we are aware about on our end at the moment, but future owners of `SqlClient` or `DependencyCollector` might not be.

With the amount of the SQL calls enterprise apps do, this may grow pretty fast in case of any issues we haven't accounted for. At the same time, I understand that cleanup (if implemented) would need to make sure that telemetry items are still sent out correctly, so there may not be a simple immediate solution.
Can you please give it some thought and see if all of this make sense :) ?

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
Dmitry-Matveev: AFAIK, the clock resolution in `UtcNow` is 10-15 ms; adding time it takes the code to execute between `DeriveDuration` and `UtcNow` (especially if this code may yield execution to the other task) we can get an interval big enough to put the dependency into the wrong spot on the timeline relative to the parent request. We may consider use higher precision timers and/or make sure we calculate duration/start as close as possible (still prone to thread tricks...).

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: @Dmitry-Matveev Thanks for the feedback, I understand where you are coming from. 

My general thoughts on this are that, as engineers, we are constantly evaluating these types of trade-offs - In this case, balancing the extra cost and complexity of a more "defensive" approach against the less expensive, simpler, and more maintainable solution.

I was leaning towards the simpler side in this case after reviewing the SqlClient code and based on my experience using SqlClient - As you point out, that code is massively used, and I have never seen or heard of such a failure - Of course, that is not to say it can't ever happen, it just seemed very unlikely.

I can try to update the impl. Is the best approach to use one of _CacheBasedOperationHolder_ or _ObjectInstanceBasedOperationHolder_. If so, which one?

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
anpete: Does this problem go away if we use CacheBasedOperationHolder? Presumably, we can then just record the timestamp when we get the begin event.

cijothomas: Yes i think we store the start timestamp inside the telemetry object kept inside the cache.
Check this:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared/Implementation/FrameworkSqlProcessing.cs

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
cijothomas: If you are using id's as the key, then CacheBasedOperationHolder. The later, if you store the actual object itself in the dictionary.
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared/Implementation/FrameworkSqlProcessing.cs  This shows usage of cachebasedoperation holde.r

anpete:
@Dmitry-Matveev @cijothomas Pushed a commit that uses ObjectInstanceBasedOperationHolder to do the event correlation. CacheBasedOperationHolder cannot be used on netcore because it relies on System.Runtime.Caching.MemoryCache, which is not available.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,527 @@
...
+
+        void IObserver<KeyValuePair<string, object>>.OnError(Exception error)
+        {
+        }
+
+        void IObserver<KeyValuePair<string, object>>.OnNext(KeyValuePair<string, object> evnt)
+        {
+            var timestamp = DateTimeOffset.UtcNow;
+
+            switch (evnt.Key)
```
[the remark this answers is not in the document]
anpete: Done

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,527 @@
...
+                            Timestamp = timestamp,
+                            Duration = TimeSpan.FromTicks(ticks),
+                            Success = true
+                        };
+
+                        InitializeTelemetry(telemetry, operationId);
+
+                        this.operationHolder.Store(command, Tuple.Create(telemetry, /* isCustomCreated: */ false));
+                    }
+
```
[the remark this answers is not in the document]
anpete: Done

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,527 @@
...
+                    if (tuple != null)
+                    {
+                        this.operationHolder.Remove(command);
+
+                        var telemetry = tuple.Item1;
+
+                        var ticks = (long)CommandAfter.Timestamp.Fetch(evnt.Value);
+
+                        telemetry.Duration = TimeSpan.FromTicks(ticks) - telemetry.Duration;
```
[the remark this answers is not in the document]
anpete: Updated. This is better - I wasn't aware of this existing pattern.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,527 @@
...
+
+                        this.client.Track(telemetry);
+                    }
```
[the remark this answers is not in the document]
anpete: Done

cijothomas:
I think all comments are addressed except Adding Functional Test, and ChangeLog.md.
(For changelog.md, please write a 1 line describing this feature. also mention which aspnet core version/ef version onwards this is supported)

anpete:
Working on the FTs.

anpete:
Added functional tests. I am going to leave the changelog to you as you know the specifics about your releases. Here is some text you can use:

"Adds ADO SQL dependency collection for SqlClient (System.Data.SqlClient) on .NET Core versions 1.0 and 2.0."

Dmitry-Matveev:
@anpete , we are fixing test infra issues to validate the tests. Will let you know as soon as infra is back online.

cijothomas:
@MS-TimothyMothra has fixed the infra and functional tests ran successfully for this.
I'll add changelog.md myself

PR #624, opened 2017-06-21
Added scripts for disable/enable strong name validation for microsoft…

cijothomas:
… assemblies.

### DIFF
--- a/disablestrongnamevalidation.ps1
+++ b/disablestrongnamevalidation.ps1
@@ -1,0 +1,2 @@
+& "C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.6.2 Tools\x64\sn.exe" -Vr *,31bf3856ad364e35
+& "C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.6.2 Tools\sn.exe" -Vr *,31bf3856ad364e35
--- a/disablestrongnamevalidation.ps1
+++ b/disablestrongnamevalidation.ps1
@@ -1,2 +1,3 @@
 & "C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.6.2 Tools\x64\sn.exe" -Vr *,31bf3856ad364e35
 & "C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.6.2 Tools\sn.exe" -Vr *,31bf3856ad364e35
+# running both the above as a hack which is known to work. Its not clear why both are needed.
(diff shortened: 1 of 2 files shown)


disablestrongnamevalidation.ps1
```diff
@@ -0,0 +1,2 @@
+& "C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.6.2 Tools\x64\sn.exe" -Vr *,31bf3856ad364e35
+& "C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.6.2 Tools\sn.exe" -Vr *,31bf3856ad364e35
```
dnduffy: Why run both the x64 and regular versions?
Won't one or the other do the job?
And isn't all of our hardware 64 bit these days?

cijothomas: while i dont know the exact reason why we need both - i am sure that without running both, builds still fail with sn failure. (you can check history to validate the case)

dnduffy: It's the same tool compiled for two different platforms.
Are we running the tests on a 32 bit system?
Is the 64 bit version erroring out but the script continues?

cijothomas: No error while running sn.exe. But tests still dont run with strong name validation errors.

dnduffy: Since we don't know why the system is behaving strangely just comment that running twice is a "hack" that works but we don't know why.

cijothomas: added

pharring: `sn -Vr` writes to the registry. The 64-bit tool writes to
>HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\StrongName\Verification

Whereas the 32-bit tool writes to
>HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\StrongName\Verification

(or rather, it writes to the same location, but on 64-bit Windows, WOW64 redirection kicks in)

These registry keys are read by the CLR loader (fusion) at runtime and, again, the bitness of the calling process determines which one is read.

Therefore, you'll need both if you want this to work for both 32-bit and 64-bit processes.

dnduffy: @pharring Thank-you very much for the explanation, that explains the behavior we were seeing.

pharring:
Does this actually work, BTW? Last time I looked into this (it was in the ApplicationInsights-dotnet-logging repo), "sn -Vr" didn't help. It seemed the assemblies were partially signed with the "31bf..." key, but they wouldn't pass a "sn -vf" check. I wonder if delay-signing is broken for regular builds.

SergeyKanzhelev:
@pharring  it is http://apmtips.com/blog/2015/10/03/open-source-signing/ I'm not sure whether .NET team still using it. The idea is that you can replace assembly without changing all the libraries that depend on it.

If it causes too much problems we can stop using OSS signing or use it in specific configuration only. Not for every debug build.

pharring:
@SergeyKanzhelev thanks. Ah... You want to use OSS signing, but you're not setting the appropriate cor header bit (I think). In the "old days" this was done with the FakeSign tool, but I understand you can now use `<PublicSign>true</PublicSign>` in MSBuild. See https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/public-signing.md

PR #625, opened 2017-06-21
Revert "Disabled tests which started failing in new lab machines due to stron…"

cijothomas:
Reverts Microsoft/ApplicationInsights-dotnet-server#586


cijothomas:
Now that the build machines are configured to skip strong name validation for test dlls, we need to turn back on these tests which were earlier disabled.

cijothomas:
To be merged after https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/624

PR #632, opened 2017-06-28
QuickPulse: adding troubleshooting output to a randomly failing test.

tokaplan:


Test/PerformanceCollector/FunctionalTests/CommonTests.cs
```diff
@@ -183,8 +183,25 @@ internal static void QuickPulseTopCpuProcesses(QuickPulseHttpListenerObservable
...
+            try
+            {
+                Assert.IsFalse(
+                    samples.Any(s => s.TopCpuProcesses.Any(p => string.IsNullOrWhiteSpace(p.ProcessName) || p.CpuPercentage < 0 || p.CpuPercentage > 100)));
+            }
+            catch
+            {
+                var weirdSample = samples.First(s => s.TopCpuProcesses.Any(p => string.IsNullOrWhiteSpace(p.ProcessName) || p.CpuPercentage < 0 || p.CpuPercentage > 100));
+
+                Console.WriteLine("Top CPU test failed, weird sample found. Processes:");
```
cijothomas: Does console write works? Will this show up in build logs? I think Trace.Write. is what other tests use.

cijothomas:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Test/PerformanceCollector/FunctionalTests/IisExpress/IISExpress.cs#L37
Shows an example tracing, which we can see in .trx file.

Test/PerformanceCollector/FunctionalTests/CommonTests.cs
tokaplan: VS shows it... Let's change to Trace.Write for consistency - done.

PR #633, opened 2017-06-28
Update ASPNET Correlation to 1.0.0

cijothomas:


cijothomas:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Test/PerformanceCollector/FunctionalTests/IisExpress/IISExpress.cs#L37 
Shows an example tracing, which we can see in .trx file.

PR #662, opened 2017-08-16
[RDD] Initial RDD DiagnosticSource listener

TomMilos:
Initial implementation of ```DependencyDiagnosticSourceListener``` (#665)


Src/DependencyCollector/Shared/Implementation/DependencyDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,306 @@
...
+        /// <seealso cref="IObserver{T}.OnError(Exception)"/>
+        /// </summary>
+        /// <param name="error">An object that provides additional information about the error.</param>
+        public void OnError(Exception error)
+        {
+        }
+
+        public void OnNext(KeyValuePair<string, object> evnt)
+        {
+            if (evnt.Key.EndsWith(ActivityStopNameSuffix))
```
pharring: Use `.EndsWith(ActivityStopNameSuffix, StringComparison.Ordinal)` for both speed and correctness. Otherwise you get a culture-sensitive comparison.

Src/DependencyCollector/Shared/Implementation/DependencyDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,306 @@
...
+            /// DiagnosticListeners.
+            /// <seealso cref="IObserver{T}.OnNext(T)"/>
+            /// </summary>
+            /// <param name="value">The DiagnosticListener that exists when this listener was added to
+            /// the list, or a DiagnosticListener that got added after this listener was added.</param>
+            public void OnNext(DiagnosticListener value)
+            {
+                if (value != null)
+                {
+                    if (value.Name.EndsWith(ActivityNameSuffix))
```
pharring: Use `StringComparison.Ordinal`

Src/DependencyCollector/Shared/Implementation/DependencyDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,306 @@
...
+                    {
+                        IDisposable eventSubscription = value.Subscribe(this.rddDiagnosticListener, (evnt, r, _) => evnt.EndsWith(ActivityStopNameSuffix));
```
pharring: `StringComparison.Ordinal`

Src/DependencyCollector/Shared/Implementation/DependencyDiagnosticSourceListener.cs
TomMilos: fixed

Src/DependencyCollector/Shared/Implementation/DependencyDiagnosticSourceListener.cs
TomMilos: fixed

Src/DependencyCollector/Shared/Implementation/DependencyDiagnosticSourceListener.cs
TomMilos: fixed

cijothomas:
can you open a GH issue describing this in detail?

TomMilos:
This is an initial implementation of #665 to get early feedback

cijothomas:
Thanks Tom for sharing the context.

Src/DependencyCollector/Shared/Implementation/DependencyDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,308 @@
...
+                if (!telemetry.Context.Properties.ContainsKey(item.Key))
+                {
+                    telemetry.Context.Properties[item.Key] = item.Value;
+                }
+            }
+
+            this.client.Initialize(telemetry);
+
+            // get the Uri from event payload
+            object uriObject = payload != null ? this.dependencyCallRequestUriFetcher.Fetch(payload) : null;
```
lmolkova: I don't like the idea that every DiagnosticsSource should report RequestUri in the payload
1. it's not generic enough: databases have no requests and no Uris
2. it's part of other payloads
3. existing DiagnosticSources does not do it and presumably will not change because of 2

Src/DependencyCollector/Shared/Implementation/DependencyDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,308 @@
...
+    using System.Diagnostics;
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+    using Microsoft.ApplicationInsights.Web.Implementation;
+
+    internal class DependencyDiagnosticSourceListener : IObserver<KeyValuePair<string, object>>, IDisposable
+    {
+        private const string DataSourceNameSuffix = ".Monitoring";
```
lmolkova: this would not work with existing DiagnosticSources. Why not listen to all?

Src/DependencyCollector/Shared/Implementation/DependencyDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,308 @@
...
+        private const string ActivityNameSuffix = ".OutboundCall";
```
lmolkova: why?

Src/DependencyCollector/Shared/Implementation/DependencyDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,308 @@
...
+                }
+
+                // if more than one tag with the same name is specified, the first one wins
+                if (!telemetry.Context.Properties.ContainsKey(tag.Key))
+                {
+                    telemetry.Context.Properties[tag.Key] = tag.Value;
+                }
+            }
+
+            telemetry.Context.Properties["activity"] = currentActivity.OperationName;
```
lmolkova: is it's going to be useful? 
Activity Operation name acts like dependency type - unique for particular library

Src/DependencyCollector/Shared/Implementation/DependencyDiagnosticSourceListener.cs
TomMilos: Noise and performance. This is not a generic DataSource listener - this is a generic __RDD__ DS listener, so we do not want to try to convert any event/activity to a dependency as most of them will be unwanted

Src/DependencyCollector/Shared/Implementation/DependencyDiagnosticSourceListener.cs
TomMilos: same as above - to limit noise

Src/DependencyCollector/Shared/Implementation/DependencyDiagnosticSourceListener.cs
TomMilos: This is "best effort" - in order to create a meaningful dependency item we need to try pulling information from different properties - some libraries will use them, some won't.

Src/DependencyCollector/Shared/Implementation/DependencyDiagnosticSourceListener.cs
TomMilos: in this implementation "type" comes from "peer.service" tag and in some cases it will be a display string (with whitespace). On the other hand we should have activity (and DS) name available so that customers can identify noisy data sources

TomMilos:
@lmolkova - it seems we have different scenarios in mind - this one is supposed to be a generic RDD listener and it's purpose is to allow to collect __high quality__ dependency telemetry - as such it needs to define some requirements. Its goal is not to work with any existing DS but to lay out path for other libraries to onboard, but do it in a generic way that can be consumed by other systems (that's why we adopt open-tracing convention).

let's sync up offline

TomMilos:
Per offline discussions I made some changes to make the DS listener more generic (capable of capturing not only "dependency" items). 

Other changes:
* simplified DS/Activity naming conventions
* stopped relying on event ```payload```
* log DS name in telemetry

pharring:
Doesn't this belong in the base SDK?

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,293 @@
...
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+    using Microsoft.ApplicationInsights.Web.Implementation;
+    using Microsoft.ApplicationInsights.Channel;
+
+    internal class TelemetryDiagnosticSourceListener : IObserver<DiagnosticListener>, IDisposable
+    {
+        private const string ActivityNameSuffix = ".Monitoring";
```
lmolkova: I still prefer to get rid of it :)

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,293 @@
...
+                        }
+                    case "peer.hostname":
+                        {
+                            telemetry.Target = tag.Value;
+                            continue; // skip Properties
+                        }
+                    case "peer.service":
+                        {
+                            telemetry.Type = tag.Value;
+                            continue; // skip Properties
```
lmolkova: I think type should be taken from"component" tag (i.e. protocol)
e.g. we can have http call to the service AND grpc call to another endpoint of the same service.

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,293 @@
...
+                switch (tag.Key)
+                {
+                    case "operation.name": // not defined by OpenTracing
+                        {
+                            telemetry.Name = tag.Value;
+                            continue; // skip Properties
+                        }
+                    case "operation.data": // not defined by OpenTracing
+                        {
+                            telemetry.Data = tag.Value;
```
lmolkova: is there any docs describing what Data property is for?
it seems to have some extended info about dependency (e.g. full url) or db statement. Is it correct?
Why we really need it in general case? What library should put there?
Why they cannot put whatever they want into the properties?

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,293 @@
...
+            telemetry.Context.Properties["Activity"] = currentActivity.OperationName.Substring(0, currentActivity.OperationName.Length - StopSuffix.Length);
+
+            this.client.Initialize(telemetry);
+
+            this.client.Track(telemetry);
+        }
+
+        internal DependencyTelemetry ExtractDependencyTelemetry(Activity currentActivity)
+        {
+            DependencyTelemetry telemetry = new DependencyTelemetry();
```
lmolkova: based on the span.kind we can decide whether it's request or dependency

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,293 @@
...
+            {
+                this.listenerSubscription.Dispose();
+            }
+        }
+
+        /// <summary>
+        /// Handler for Activity stop event (response is received for the outgoing request).
+        /// </summary>
+        /// <param name="diagnosticListener">The diagnostic source of the activity.</param>
+        internal void OnActivityStop(DiagnosticListener diagnosticListener)
```
lmolkova: In case I want to implement http-specific listener, would have to implement another subscriber and listener.
We can, of course, refactor it later, but it's rarely a priority and after a while, it becomes really complicated. It would be great if to design it right now to be extendable.

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,293 @@
...
+        /// <param name="value">The DiagnosticListener that exists when this listener was added to
+        /// the list, or a DiagnosticListener that got added after this listener was added.</param>
+        public void OnNext(DiagnosticListener value)
+        {
+            if (value != null)
+            {
+                IDisposable subscription = value.Subscribe(
+                    new IndividualDiagnosticSourceListener(value, this),
+                    (evnt, r, _) => evnt.EndsWith(ActivityNameSuffix, StringComparison.OrdinalIgnoreCase)
+                        || evnt.EndsWith(ActivityStopNameSuffix, StringComparison.OrdinalIgnoreCase));
```
lmolkova: we also may receive an exception from the same source.  Should be useful to report them.

TomMilos: makes sense, although that would be Exception telemetry collection and for now I'd like to limit this PR to Dependency collection

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
TomMilos: I'll see if I can reasonably combine them together, but I don't think this should be a blocker to get this into 2.5

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
TomMilos: do you mean "span.kind"? ok, I can inspect it

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
TomMilos: they can put anything in tags, however Data is an elevated dependency property which should be specific to given technology - it can be "db.statement" or "http.url" for some services but for others it may be something entirely different (e.g. Redis, Event Hub)

currently there is no doc that specifies it but it will be described in the user guide for this

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
TomMilos: so "component" describes the (client) library, while in dependencies type is about the other (server) side of the call. Also there are multiple Redis libraries available - they would all show as different dependency types in that case

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
TomMilos: the only problem is that in that case we have to subscribe to all events - not only activities. Why? because typically before activity starts (so ".Start" event is raised) the code will check if activity is enabled - and that activity name will not have anything we can filter by so we need to subscribe to all events to get a chance to capture ".Start" events

...unless I'm missing sth

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
SergeyKanzhelev: https://docs.microsoft.com/en-us/azure/application-insights/application-insights-data-model-dependency-telemetry#data

SergeyKanzhelev:
@pharring 
> Doesn't this belong in the base SDK?

Eventually... The idea to move all the `StartOperation` and correlation business from Application Insights SDK down to DiagnosticsSource.

pharring:
I meant the base AI SDK. This issue is in the ApplicationInsights-**dotnet-server** repo. I was suggesting moving it to the ApplicationInsights-**dotnet** repo.
To help answer that:
1. Is there anything in this change that relies on code that is only in the https://github.com/Microsoft/ApplicationInsights-dotnet-server repo?
1. Is there anything in this change that you wouldn't also want in the https://github.com/Microsoft/ApplicationInsights-aspnetcore repo?

SergeyKanzhelev:
@pharring this is the change for DependencyTrackingModule that is shared with aspnetcore. So far all the collection logic is in server repo. Base SDK implementing serialization and channel logic.

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
lmolkova: so, it seems the guidance for the lib owners should be: if you are http library, fill the http.url, if you are db, fill the db.statement. For the rest, use "operation.data".
Is it correct? It seems everyone could just use operation.data to simplify things. 
If it is so, I don't really see the value of supporting all possible opentracing tags. 
We can support them, but the guidance doc should probably tell about "operation.data".
Thoughts?

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
lmolkova: well, it's all about guidance we give and how simple is to follow it. Whatever tag is used, we should tell lib owners to use very high-level name like 'redis' and not 'stack-exchange-redis-version-1.2.3'. 

Still, if there is a http call to docDb and also a tcp call to the same docDb, dependency type should be different, I don't think peer.service would allow this scenario.

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
lmolkova: You can just return false to all events that end with Start and thus enable all possible events.

Now, the 'monitoring' suffix you have, tells to the listener that it's going to be about Activity.
If we are going to filter sources that are compatible with Activity based on something, it should rather be a DiagnosticSource name than Activity name. It would be more efficient (as long names need to be compared only when someone subscribes to a source rather than per each operation).

I assume you want to support scenarios when DiagnosticSource sends both Activity Start/Stop and other events.
Our guidance should be to discourage this scenario at all. As we discussed offline, DiagnosticSource is not a logger.
In the worst case, we let in all other events and filter them out only in the Write-callback

I think once we have a guidance for lib owners, we should review it by few more people (including some lib owners)  and then we can return back to the listener name-based filtering.

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
lmolkova: I'm fine with implementing it later. 
Though the same listener should handle Exception events as well, it cannot be done by other listeners (or would be quite inefficient)

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
lmolkova: Oh, is there a plan to make it into 2.5?

What about lib instrumentation and guidance? Should it also be done in 2.5 timeframe?

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
TomMilos: Agreed :)

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
TomMilos: guidence - yes, I'll start working on it this week and add it to this PR
lib instrumentation is independent (as long as it follows guidance) - we've already started working on 2 but it's not clear when they will be done as they have different dev cycle

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
TomMilos: That makes sense - in the guidance we will only mention **operation.data** but we will silently fall back to **http.url** / **db.statement**.

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
TomMilos: I think I will start writing down the guidance and there it should be easier to see if anything is redundant/missing/inconsistent

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
TomMilos: ok, I'll remove this 
good point about filtering DS vs Activities - with DS white/blacklist that should be flexible enough

billpratt:
I really like this idea. I had a few questions:

1) Will there be a way to filter out activities? I can see not wanting to send some things to app insights.
2) Could this be used for AI events as well? It would be really nice to broadcast an event through DiagnosticSource and have AI pick it up.

TomMilos:
Hi Bill,

Appreciate your feedback! To answer your questions:

Ad.1 Yes – we plan to include whitelist/blacklist capability in this version so you will be able to control it via config

Ad.2 Current focus is on Dependency telemetry, however we do want to enhance this feature to be able to collect other types of telemetry based on DS. We were considering supporting Requests and Exceptions next as they are fairly easy to distinguish. Custom events are very generic so we’d have to think through how to decide which activities/events should be converted. If you have any particular ideas, please share ☺

Thanks
             Tom

PartnerLibraryInstrumentationGuidelines.md
```diff
@@ -0,0 +1,193 @@
...
+[Activity][ActivityGuide] is a class that allows storing and accessing diagnostics context and consuming it with logging system.
+
+Both Diagnostic Source and Activity have been used to instrument [System.Net.Http][SystemNetHttp] and [Microsoft.AspNetCore.Hosting][MicrosoftAspNetCoreHosting], although that instrumentation is not fully complaiant with this guidance.
+
+[This document][DiagnosticSourceActivityHowto] can help to a better understanding on how to efficiently use Diagnostic Source.
+
+### Instrumentation code
+
+The following code sample shows how to instrument the operation logic enclosed in ```ProcessOperationImplAsync()``` method, in the most efficient way which will ensure no performance overhead is added if there are no listeners for that particular Activity.
+
```
vancem: Thanks Tomasz for soliciting my feedback.      What makes logging 'hard' is the distributed nature of the instrumentation points (being authored by many people) which makes consistency hard and pretty much all problem grow from there.  

The ideal situation is if the guidance is 'easy' and hard to mess up.  As you can see, this is not the case for the sample code below (We are suggesting ~100 lines of code JUST for instrumenting ONE start-stop pair.   WE CAN'T GO OUT WITH THIS AS THE GUIDANCE!   It just won't get done correctly/uniformly.  

First, I would like to break the guidance up, and first show the synchronous case (where the operation represented by the activity is simply a call to a function on the same thread and not an async Task)  Then we can show the non-async case (which is a bit more complex, but that is Async complexity not logging complexity).  

But fundamentally we need to cut the lines of code down.   If we take the synchronous case, I really am trying to get as close to the following code as possible (which is still more complex than I want, but not too bad).  Note that I am assuming APIs that don't currently exist but which I would like to exist.   
```
    Activity activity = null;
    if (source.IsEnabled(ActivityName)) // This is here for efficiency (avoiding creating payload object when off.  
        activity = source.StartActivity(ActivityName, new {Input = input });

    var output DoOperation(input); // Do the actual operation (probably more than just one function call) 

    if (activity != null)
         source.StopActivity(activity,  new  { Output = output }); 
```
Of cource I need my new overloads for some methods, but that is trivial and I will ignore it for now...
The main differences between what you have and what I want are the following
1. I don't bother with exception cases or logging the exceptions.
2. I don't add any tags to the activity. (you have component, span.kind, error).   

Note that I would not have a problem with particular important components adding these additional things (like the HTTP incoming and outgoing activities), but I don't really want to mandate that every component should be doing this.   

The question is how bad is it that I have eliminated these things in the common case?     Certainly I am open to negotiation here, but I would argue that it is not bad.   Exceptions are already heavily logged, and they naturally get stamped with the correct activity, so it is not clear we need per-component processing of exceptions.    The issue of the tags are more interesting.  'error' feels redundant with the exception being thrown, so I would not want to mandate it unless we were forced to (especially if we have it on the 'boundary' where HTTP requests come into the system).   span.kind = client is certinaly something that the infrastructure could do for you (it is a constant for all code here correct?.   Component is interesting, but feels like it overlaps with the name of the DiagnosticSource.  I would like to understand the real value here.   It seems like we could live without it (at least for most components). 

This probably needs more discussion, but this is the 'big rock'. 

As an aside, the code we show should definitely a true sample (where we just have  Console.WriteLine for the component body.    When looking at the code below I *think* I see errors that would cause compilation failure.   I don't want to dwell on those until we agree with the actual code that should be in the guidance.

Once again, I want to thank you letting me review.   We are trying to make the logging experience as good as it can be and I think cutting the complexity of what we ask people to write at an instrumentation point is a key aspect of 'getting it right'.   It is worth our time...

vancem:
@brianrob

PartnerLibraryInstrumentationGuidelines.md
TomMilos: Thanks @vancem - as mentioned offline, I agree with some of the things you brought up (particularly complexity of the code sample) and will address those (in separate PR).

TomMilos:
In order to make progress on this PR I've removed the guidance document as it will most likely require a couple more iterations.

I've also implemented "blacklisting" of DS/Activities, added more tests for generic listener and simplified a couple of things. Please take a look

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,369 @@
...
+                        }
+                    case "peer.service":
+                        {
+                            peerService = tag.Value;
+                            break;
+                        }
+                }
+
+                // if more than one tag with the same name is specified, the first one wins
+                if (!telemetry.Properties.ContainsKey(tag.Key))
```
lmolkova: I think this part should live in base SDK OperationCorrelationTelemtryInitializer. This way it would work for Activities created without DiagnosticSource events or non-generic listeners.

Src/DependencyCollector/NuGet/ApplicationInsights.config.install.xdt
```diff
@@ -17,6 +17,18 @@
         <Add>localhost</Add>
         <Add>127.0.0.1</Add>
       </ExcludeComponentCorrelationHttpHeadersOnDomains>
+      <ExcludeDiagnosticSourceActivities>
```
lmolkova: please also take care about ASP.NET Core DiagnosticListener.

lmolkova:
1. Extensibility. I still want to see extension point: how this listener could be extended to substitute existing listeners. If it's not done now, there will never be a chance to refactor it.  I'm fine if you just make sure it's easily possible, even without implementing this extensibility point

2. Tags. 
- As guidance is no longer a plan for 2.5, I suggest that all tags that we are not sure are essential are removed completely. Whatever is released now, will have to stay for backward compatibility. So the overhead and code complications will have to remain forever. It's easy to add things later.
I believe `operation.type`, `component`, `error`, `db,statement`, and all http tags are not really needed (at least for now)

- It would be great to have Tags initialization in base SDK OperationCorrelationTelemtryIntitializer for the Request and Dependency telemetries. It would solve https://github.com/Microsoft/ApplicationInsights-dotnet/issues/562 and enable variety of features customers were asking about (see https://github.com/Microsoft/ApplicationInsights-dotnet/issues/562 linked issues )

Src/DependencyCollector/NuGet/ApplicationInsights.config.install.xdt
TomMilos: Do you mean ```Microsoft.AspNet.TelemetryCorrelation```? It's already excluded

TomMilos:
I've extracted the generic ```DiagnosticSourceListenerBase``` class which encapsulates listener subscription management. It can be reused by other existing listeners (will see if I will refactor them as part of this PR)

I've also removed most of the OT and custom tag handling - although I don't agree that this would be a back-compat liability as this is the consumption side.

Not sure if I follow the ask about ```OperationCorrelationTelemtryIntitializer``` - if I understand correctly that would mean that all telemetry items, regardless of type would always attempt to include the current activity tags?

Src/DependencyCollector/NuGet/ApplicationInsights.config.install.xdt
lmolkova: I mean  https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/develop/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs

DependencyCollector package will be also used in AspNet Core SDK. 
AspNet Core SDK does not have xml configuration file, so `Microsoft.AspNetCore` should be excluded in some other way.

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,186 @@
...
+            return excludedActivities;
+        }
+
+        internal override bool IsEventEnabled(string evnt, object input1, object input2, DiagnosticListener diagnosticListener, HashSet<string> context)
+        {
+            return !this.IsActivityExcluded(evnt, context)
+                && !evnt.EndsWith(ActivityStartNameSuffix, StringComparison.OrdinalIgnoreCase);
+        }
+
+        internal override bool ShouldHandleEvent(KeyValuePair<string, object> evnt, DiagnosticListener diagnosticListener, HashSet<string> context)
```
lmolkova: what is the point in having IsEventEnabled and ShouldHandleEvent? it's pure overhead and I don't think there is a real case for ShouldHandleEvent. In any case, listener may just do this check in HandleEvent method

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,186 @@
...
+            if (currentActivity == null)
+            {
+                DependencyCollectorEventSource.Log.CurrentActivityIsNull();
+                return;
+            }
+
+            DependencyCollectorEventSource.Log.TelemetryDiagnosticSourceListenerActivityStopped(currentActivity.Id, currentActivity.OperationName);
+
+            // extensibility point - can chain more telemetry extraction methods here
+            ITelemetry telemetry = this.ExtractDependencyTelemetry(diagnosticListener, currentActivity);
```
lmolkova: 1. some listeners will want to extract telemetry from the `evnt`. 
2. Passing currentActivity is not necessary - not everyone would need it and those who need may use Activity.Current

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,186 @@
...
+
+            foreach (var item in currentActivity.Baggage)
+            {
+                if (!telemetry.Context.Properties.ContainsKey(item.Key))
+                {
+                    telemetry.Context.Properties[item.Key] = item.Value;
+                }
+            }
+
+            this.Client.Initialize(telemetry);
```
lmolkova: `Track` calls initialize internally, you don't need to call initialize explicitely

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,186 @@
...
+            if (telemetry == null)
+            {
+                return;
+            }
+
+            // properly fill dependency telemetry operation context
+            telemetry.Context.Operation.Id = currentActivity.RootId;
+            telemetry.Context.Operation.ParentId = currentActivity.ParentId;
+            telemetry.Timestamp = currentActivity.StartTimeUtc;
+
```
lmolkova: duration

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
```diff
@@ -0,0 +1,186 @@
...
+            telemetry.Type = diagnosticListener.Name;
+            telemetry.Name = currentActivity.OperationName;
+
+            foreach (KeyValuePair<string, string> tag in currentActivity.Tags)
+            {
+                // interpret Tags as defined by OpenTracing conventions
+                // https://github.com/opentracing/specification/blob/master/semantic_conventions.md
+                switch (tag.Key)
+                {
+                    case "error":
```
lmolkova: I'm still not convinced we really have to have boolean error tag.

lmolkova:
> Not sure if I follow the ask about OperationCorrelationTelemtryIntitializer - if I understand correctly that would mean that all telemetry items, regardless of type would always attempt to include the current activity tags?

There are cases now when Activity may be started in user code (e.g. when there is no instrumentation in the component). Or existing listeners: they do not have this code that assigns Tags from the Activity to telemetry. Tags seem to be useful in many scenarios, but now only listeners that inherit from this base generic listener would use them. This is not consistent approach.
 
So my point is that Activity.Tags are generic part and should be consistently used in all scenarios (i.e. assigned by telemetry initializer in base SDK).
As tags belong only to the current Activity, they should be set on the telemetry that is described by this Activity: i.e. DependencyTelemetry should get Tags from its Activity, but all children (if any) of dependency telemetry should not get tags of this Activity.


So eventually, [here](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/e544ffae4f3188bde01a367364ea3e36f2bf03a9/src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/OperationCorrelationTelemetryInitializer.cs#L23
 ), I would expect to have code like 

```C#
  if (telemetryItem is OperationTelemetry)
  {
       foreach (var tag in currentActivity.Tags)
           if (!telemetryItem.Context.Properties.ContainsKey(tag.Key))
                telemetryItem.Context.Properties.Add(tag);
  }
```

I'm fine if you decide to skip it for now. But eventually it has to be implemented

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
TomMilos: In case of Activities - ```IsEventEnabled``` will return ```true``` for events with activity name and activity stop name, while ```ShouldHandleEvent``` will only return ```true``` for the stop event. 

Yes, ```ShouldHandleEvent``` could be merged into ```HandleEvent``` however I was considering a default implementation which would work for non-activity cases.

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
TomMilos: so ```ExtractDependencyTelemetry()``` is just a method to create dependency telemetry from the activity. To try to convert it to sth else new methods would be added here (in a chain-of-responsibility fashion) and those new methods can have different signature depending on their needs

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
TomMilos: ```Duration``` is specific to RDD telemetry and it is initialized inside ```ExtractDependencyTelemetry()``` method

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
TomMilos: good point, thanks. 
should I also remove from ```HttpCoreDiagnosticSourceListener``` or is there a reason for it there?

lmolkova: there it's done in Start api and I believe serves a different purpose: it initializes all properties before request headers are injected into an outgoing request.

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
lmolkova: what if listener creates Requesttelemetry or OperationTelemetry? 
every virtual method call costs something. Everything that is known about this event is hidden in diagnostics source name, activity name and event payload. Why not have 1 method like: extract telemetry from source, event, payload? What value chain of methods brings?

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
lmolkova: ok, thanks for explaination

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
lmolkova: so it means that every listener would sacrifice some of the performance on the hottest path (IsEnabled being called twice) to enable easier support for non-Activity listeners? I'd prefer to keep non-Activity listeners overriding IsEnabled and Handle however they want, but keeping most interesting, popular and heavy loaded scenarios faster

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
TomMilos: So this particular listener has no fixed extensibility design - it will be figured out when needed. But it is not meant to be subclassed (the ```*Base``` one is), rather more logic should be added in this place - whether in separate methods or the same one that already exists - but that will depend on the actual enhancement that's being made.

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
TomMilos: I've already removed ```ShouldHandleEvent``` as suggested

Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
TomMilos: yeah, it seems to make sense there. thanks for explaining. leaving as is.

TomMilos:
@lmolkova I see, so we already copy the baggage in the ```OperationCorrelationTelemetryInitializer``` - was there any reason to not do the same with Tags?

I can make that change there, but in order to unblock testing I'd rather leave it here as well and then remove after the other PR is merged

```C#
foreach (var baggage in currentActivity.Baggage)
{
    if (!telemetryItem.Context.Properties.ContainsKey(baggage.Key))
    {
        telemetryItem.Context.Properties.Add(baggage);
    }
}
```

lmolkova:
> I see, so we already copy the baggage in the OperationCorrelationTelemetryInitializer - was there any reason to not do the same with Tags?

I can make that change there, but in order to unblock testing I'd rather leave it here as well and then remove after the other PR is merged

@TomMilos 
The reason for not including Tags was that we did not know how to use them (i.e. had no scenarios). It seems now that they are quite useful in many scenarios. But you are right, support for tags in base SDK could be done in addition to this PR and is not really related to it. Could you please leave a comment that once https://github.com/Microsoft/ApplicationInsights-dotnet/issues/562 is fixed, this code may go away?

TomMilos:
@lmolkova Done. I've also removed the code that populates properties based on baggage as it is redundant with the ```OperationCorrelationTelemetryInitializer```

SergeyKanzhelev:
So you decided to have it ON by default in `beta1`?

TomMilos:
@cijothomas EF Diagnostic source does not have to be excluded as it is not instrumented with Activities so this listener will ignore it

PR #664, opened 2017-08-16
Perf collector functional test fix

cijothomas:
Performance Collector Functional Test did not clean up if an exception occurred in test initialize. This is to fix it so as to prevent all future tests from failing due to port conflicts.

### DIFF
--- a/Test/PerformanceCollector/FunctionalTests/Test40.cs
+++ b/Test/PerformanceCollector/FunctionalTests/Test40.cs
@@ -7,6 +7,7 @@
     using Functional.IisExpress;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
     using PerfCounterCollector.FunctionalTests;
+    using System;
 
     [TestClass]
     public class Test40 : SingleWebHostTestBase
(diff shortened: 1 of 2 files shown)


Test/PerformanceCollector/FunctionalTests/Test40.cs
```diff
@@ -38,7 +39,15 @@ public void TestInitialize()
...
 
-            base.LaunchAndVerifyApplication();
+            try
+            {
+                base.LaunchAndVerifyApplication();
+            }
+            catch (Exception ex)
+            {
+                Trace.WriteLine("Exception occured while verifying application, exception:" + ex.Message);
+                this.StopWebAppHost();
```
Dmitry-Matveev: Will PerfCounter tests fail quickly after that? If they rely on some replies from web app host, they might wait for some timeout and it'll prolong the test time.

cijothomas: You are right. I'll throw the ex , so that tests won't be run at all since failure is at test initialize. Good catch!

PR #675, opened 2017-08-22
Dependency Collector for Aspnet core - added 2.0 app

cijothomas:
Duplicated the current test app and created 2.0 app.

There are issues in build machines which cause build failures, but passes locally. Wont merge until build system issue is identified


cijothomas:
Not merging until the issue with build servers are fully understood.. dotnet build - works fine. msbuild is failing in build machines. Probably something to do with msbuild version or visual studio version...

SergeyKanzhelev:
@cijothomas just thinking out loud - can it be arguments we pass to msbuild in build definition (`/toolsversion:15.0 /p:VisualStudioVersion=15.0`). Have you tried to run the same command like as build server locally?

cijothomas:
@SergeyKanzhelev its understood now. VS 2015 update 3 solves the issue.

cijothomas:
This is now passing all builds after vs update. (For some reason, the green build is not reporting status to github)

Test/DependencyCollector/FunctionalTests/FuncTest/AspNetCore20HttpTests.cs
```diff
@@ -0,0 +1,177 @@
...
+    public class AspNetCore20HttpTests
+    {
+        internal const string AspxCoreAppFolder = ".\\AspxCore20";
+        internal const string AspxCoreTestAppFolder = "..\\TestApps\\AspxCore20\\";
+        internal static DotNetCoreTestWebApplication AspxCoreTestWebApplication { get; private set; }
+
+        public TestContext TestContext { get; set; }
+
+        [ClassInitialize]
+        public static void MyClassInitialize(TestContext testContext)
```
Dmitry-Matveev: I remember having [this issue](https://blogs.msdn.microsoft.com/ploeh/2007/01/06/classcleanup-may-run-later-than-you-think/) with `ClassInitialize` of the second test class in the same project running before `ClassCleanup` of the first class.
In this project it would mean that we try to deploy App to the same port and it can be busy with the previous test app because `ClassCleanup` was not run.
Have you noticed similar behavior here or it's always run in the order as it should?

Dmitry-Matveev: I noticed that the port is different below, so should be no issue. However, please check that there are no other side effects if `ClassCleanup` is not executed before the other `ClassInitialize`.

Test/DependencyCollector/FunctionalTests/TestApps/AspxCore20/Startup.cs
```diff
@@ -0,0 +1,50 @@
...
+            loggerFactory.AddConsole(Configuration.GetSection("Logging"));
+            loggerFactory.AddDebug();
+
+            app.UseMvc();
+
+            var teleConfig = TelemetryConfiguration.Active;
+            teleConfig.TelemetryChannel.DeveloperMode = true;            
+
+            // Fake endpoint.
+            teleConfig.TelemetryChannel.EndpointAddress = "http://localhost:8789/v2/track";
```
Dmitry-Matveev: Minor: Looks hardcoded - do we start this endpoint ourselves, so we can pass a variable here as a port?

Test/DependencyCollector/FunctionalTests/TestApps/AspxCore20/Controllers/ExternalCallsController.cs
```diff
@@ -0,0 +1,124 @@
...
+        private static string MakeHttpPostCallSync(int count, string hostname)
+        {
+            string result = "";
+
+            Uri ourUri = new Uri(string.Format("https://www.{0}.com", hostname));
+            HttpClient client = new HttpClient();
+            HttpContent content = new StringContent("thing1=hello&thing2=world", Encoding.ASCII);
+            for (int i = 0; i < count; i++)
+            {
+                result += $"Request {i + 1}:<BR/>{client.PostAsync(ourUri, content).Result}<BR/>";
```
Dmitry-Matveev: Do we instrument any Sync method in ASP.NET core or none such method exists nowadays? ;)

Test/DependencyCollector/FunctionalTests/FuncTest/AspNetCore20HttpTests.cs
cijothomas: I am guessing the issue with class initiazlie/cleanup ordering is whats causing occasional test failures in aspnet core sdk repo.. not seen an issue in web sdk.. 
i'll address this before or after merge anyway.

Test/DependencyCollector/FunctionalTests/TestApps/AspxCore20/Startup.cs
cijothomas: we start it ourself.. will adjust

Test/DependencyCollector/FunctionalTests/TestApps/AspxCore20/Controllers/ExternalCallsController.cs
cijothomas: If we want to test sync vs async we can add it in unit tests. Dont think we need to increase functional test cases for that as its slightly slower than unit test.

Test/DependencyCollector/FunctionalTests/TestApps/AspxCore20/Startup.cs
cijothomas: just realized that no common place to have all constants. its hardcoded to all projects in ai.config or startup.

Issue #678, opened 2017-08-24
PerfCounterCollector module may go into endless loop (ASP.NET Core on Full Framework)


Dmitry-Matveev:
Concurrent access to `ExpandInstanceName()` method of `PerfCounterCollector` module may lead to an infinite loop leading to 100% CPU consumption.

[Here is the first place](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/PerformanceCollector/Perf.Shared/Implementation/PerformanceCounterUtility.cs#L338) within this `ExpandInstanceName()` method access the private static `Dictionary<string, string> PlaceholderCache = new Dictionary<string, string>();`.

[Here is the second place](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/PerformanceCollector/Perf.Shared/Implementation/PerformanceCounterUtility.cs#L363) within the same method that modifies the content of the Dictionary.

It's been reported that such Dictionary access may lead to the infinite loop in ASP.NET Core app running on .NET Full Framework, the relevant stacks from the dump look like this:
![image](https://user-images.githubusercontent.com/12158043/29688401-8466dab2-88d4-11e7-8ab2-a4792a280158.png)

PR #679, opened 2017-08-25
PerformanceCounterUtility to use concurrent dictionary

MS-TimothyMothra:
#678


SergeyKanzhelev:
So what's the root cause of the issue? Why do we have a multiple threads trying to expand the names?

cijothomas:
@SergeyKanzhelev - I believe we can go ahead and merge this, even if we don't know the exact scenario which caused multi-thread situation. Let me know otherwise.

zvolkov:
Experiencing same issue here. When are you planning the next release to Nuget.org?

cijothomas:
@zvolkov this fix will be in 2.5.0-beta1 which will be released very soon.. (1 week or so)

Issue #649, opened 2017-07-18
DiagnosticsListener should have safe OnNext


SergeyKanzhelev:
`HttpCoreDiagnosticSourceListener` should have a safe types extraction in [OnNext](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/16dbd0079e47aac4a6f78bcee4c10698cfe4b3b2/Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs#L100) and should be wrapped in `try{}catch` for any unexpected exceptions.

See https://github.com/dotnet/corefx/issues/22251#issuecomment-315940409

PR #680, opened 2017-08-28
DiagnosticsListener type safe onnext

MS-TimothyMothra:
#649

### DIFF
--- a/Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs
+++ b/Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs
@@ -97,6 +97,11 @@ namespace Microsoft.ApplicationInsights.DependencyCollector.Implementation
         {
         }
 
+        /// <summary>
+        /// Provides the observer with new data.
+        /// <seealso cref="IObserver{T}.OnNext(T)"/>
+        /// </summary>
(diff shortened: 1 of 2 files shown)


MS-TimothyMothra:
@SergeyKanzhelev Can you please review the new logging function and confirm that the EventLevel is correct?

Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs
```diff
@@ -97,49 +97,88 @@ public void OnError(Exception error)
...
-                    this.OnActivityStop(
-                        (HttpResponseMessage)this.stopResponseFetcher.Fetch(evnt.Value),
-                        (HttpRequestMessage)this.stopRequestFetcher.Fetch(evnt.Value),
-                        (TaskStatus)this.stopRequestStatusFetcher.Fetch(evnt.Value));
-                    break;
-                }
+                            if (request != null)
+                            {
+                                this.OnActivityStart(request);
+                            }
```
Dmitry-Matveev: Here and in the other places, do we need to do something in "else"?
It looks like a subtle change in the `OnNext()` behavior without it:
- previously, we tried to execute the code and either succeeded or the failure was reported
- currently, we try to execute the code and either succeed or silently continue

PR #681, opened 2017-08-28
Change Event Ids to match Event attribute

MS-TimothyMothra:


MS-TimothyMothra:
@SergeyKanzhelev the EventRegister package is already referenced in this project.
I think it would be worth fixing this now and creating a separate issue to investigate why events aren't being validated.

Issue #683, opened 2017-09-08
App Id Lookup bug


cijothomas:
Same issue as in aspnet core, need similar fix in web sdk as well.
https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/524

PR #685, opened 2017-09-12
App Id Lookup Bugfix

MS-TimothyMothra:
#683 
Refactored preprocessors. Affirmative if statements are easier to comprehend instead of nested negative statements.
Added OK check from https://github.com/Microsoft/ApplicationInsights-aspnetcore/pull/525/files

### DIFF
--- a/Src/Common/CorrelationIdLookupHelper.cs
+++ b/Src/Common/CorrelationIdLookupHelper.cs
@@ -189,12 +189,18 @@
         /// Format and store an iKey and appId pair into the dictionary of known correlation ids.
         /// </summary>
         /// <param name="ikey">Instrumentation Key is expected to be a Guid string.</param>
-        /// <param name="appId">Application Id is expected to be a Guid string.</param>
+        /// <param name="appId">Application Id is expected to be a Guid string. App Id needs to be Http Header safe, and all non-ASCII characters will be removed.</param>
         private void GenerateCorrelationIdAndAddToDictionary(string ikey, string appId)
         {
-            if (appId != null && Guid.TryParse(appId, out Guid appIdGuid))
+            if (string.IsNullOrWhiteSpace(appId))
             {
-                this.knownCorrelationIds[ikey] = string.Format(CultureInfo.InvariantCulture, CorrelationIdFormat, appIdGuid.ToString());
+                return;
+            }
+
+            appId = HeadersUtilities.SanitizeString(appId);
+            if (!string.IsNullOrEmpty(appId))
+            {
+                this.knownCorrelationIds[ikey] = string.Format(CultureInfo.InvariantCulture, CorrelationIdFormat, appId);
             }
         }
 
--- a/Src/Common/CorrelationIdLookupHelper.cs
+++ b/Src/Common/CorrelationIdLookupHelper.cs
@@ -38,7 +38,7 @@
         private ConcurrentDictionary<string, int> fetchTasks = new ConcurrentDictionary<string, int>();
 
         // Stores failed instrumentation keys along with the time we tried to retrieve them.
-        private ConcurrentDictionary<string, FailedResult> failingInstrumenationKeys = new ConcurrentDictionary<string, FailedResult>();
+        private ConcurrentDictionary<string, FailedResult> failingInstrumentationKeys = new ConcurrentDictionary<string, FailedResult>();
 
         private Func<string, Task<string>> provideAppId;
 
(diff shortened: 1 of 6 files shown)


Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -313,6 +350,20 @@ private void RegisterFailure(string instrumentationKey, Exception ex)
...
+        /// Registers failure for further action in future.
+        /// </summary>
+        /// <param name="instrumentationKey">Instrumentation key for which the failure occurred.</param>
+        /// <param name="httpStatusCode">Response code from AppId Endpoint.</param>
+        private void RegisterFetchFailure(string instrumentationKey, HttpStatusCode httpStatusCode)
+        {
+            this.failingInstrumenationKeys[instrumentationKey] = new FailedResult(DateTime.UtcNow);
+            
+            const string FetchAppIdResponseErrorFormat = "Fetch AppId failed with response code '{0}'";
+            AppMapCorrelationEventSource.Log.FetchAppIdFailed(string.Format(CultureInfo.InvariantCulture, FetchAppIdResponseErrorFormat, httpStatusCode));
```
SergeyKanzhelev: why do we use event log with the `string.Format`? Please make a dedicated event for it with the argument of response code.

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -245,26 +233,75 @@ private Task<string> FetchAppIdFromService(string instrumentationKey)
...
+                SdkInternalOperationsMonitor.Enter();
+
+                string result = null;
+                Uri appIdEndpoint = this.GetAppIdEndPointUri(instrumentationKey);
+                WebRequest request = WebRequest.Create(appIdEndpoint);
+                request.Method = "GET";
+
+                using (HttpWebResponse response = (HttpWebResponse)await request.GetResponseAsync().ConfigureAwait(false))
+                {
+                    if (response.StatusCode == HttpStatusCode.OK)
```
SergeyKanzhelev: wouldn't `GetResponseAsync` throw an exception when the response is not `200`?

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -245,26 +233,75 @@ private Task<string> FetchAppIdFromService(string instrumentationKey)
...
+                    {
+                        using (StreamReader reader = new StreamReader(response.GetResponseStream()))
+                        {
+                            result = await reader.ReadToEndAsync();
```
SergeyKanzhelev: Since the issue is a bad response - should we limit the size of response and maybe filter out special characters and such?

Src/Common/CorrelationIdLookupHelper.cs
```diff
@@ -313,6 +350,19 @@ private void RegisterFailure(string instrumentationKey, Exception ex)
...
 
+        /// <summary>
+        /// FetchAppIdFromService failed.
+        /// Registers failure for further action in future.
+        /// </summary>
+        /// <param name="instrumentationKey">Instrumentation key for which the failure occurred.</param>
+        /// <param name="httpStatusCode">Response code from AppId Endpoint.</param>
+        private void RegisterFetchFailure(string instrumentationKey, HttpStatusCode httpStatusCode)
+        {
+            this.failingInstrumenationKeys[instrumentationKey] = new FailedResult(DateTime.UtcNow);
```
SergeyKanzhelev: will it ever be set multiple times? Will it behave OK when you attempt to override value with the new one?

MS-TimothyMothra: Removing this. I verified that the encapsulating code is performing this step.

Src/Common/CorrelationIdLookupHelper.cs
MS-TimothyMothra: The original bug was an error to put a non-ascii LineReturn character in an http header.
Assuming we trust our services, we shouldn't have to sanitize a successful response.
And these changes will not parse the response body of any failed requests, so I think we're covered.

If you disagree we can scrub it.

Src/Common/CorrelationIdLookupHelper.cs
MS-TimothyMothra: I can't find documentation that states GetResponse will throw an exception for non 200 responses.
But I think we're covered in this scenario as well because the encapsulating code is catching and logging all exceptions.

Do you want us to handle exceptions differently?

Src/Common/CorrelationIdLookupHelper.cs
SergeyKanzhelev: even if we trust our server we still need to make sure that return string will not break customer application. So there should be a sanitization logic either here or when we serialize app-id into the http header.

MS-TimothyMothra: That makes sense. I added a sanitization check (line 195: Guid.TryParse()) after the fact when we add the app-id to the dictionary. I chose to do it here because the FetchAppIdFromService() method can be overridden at runtime.

SergeyKanzhelev: I'd make it less format-specific and more http-safe. Just strip out all characters forbidden in http header value. This way we can encode something into `app-id` in future if we will need to. Like prefix or something

PR #699, opened 2017-09-28
fix for Rest Api filtering bug

MS-TimothyMothra:
#175 

System.Web.Handlers.TransferRequestHandler
	This redirects extensionless requests to a controller (ex: site/home -> site/HomeController.cs)
	The redirection creates a "child" request of the original "parent" request.
	The child request will share the headers of the parent.

Microsoft.ApplicationInsights.Web.RequestTrackingTelemetryModule
	Expected behavior: Tracks telemetry for individual request
	Bug: is not logging telemetry for 200 (successful) requests when - TransferRequestHandler is filtered
	Bug: without filter, logs 2 records for every 200 (successful) request

Sergey's suggested fix: attach id header to parent requests, and only log child request.

### DIFF
--- a/Src/Web/Web.Net45.Tests/AspNetDiagnosticTelemetryModuleTest.cs
+++ b/Src/Web/Web.Net45.Tests/AspNetDiagnosticTelemetryModuleTest.cs
@@ -314,7 +314,7 @@
 
             AspNetDiagnosticTelemetryModule result = new AspNetDiagnosticTelemetryModule();
 
-            var requestModule = new RequestTrackingTelemetryModule();
+            var requestModule = new RequestTrackingTelemetryModule(enableSafeRequestTracking: true);
             var exceptionModule = new ExceptionTrackingTelemetryModule();
             requestModule.Initialize(this.configuration);
             exceptionModule.Initialize(this.configuration);
--- a/Src/Web/Web.Net45.Tests/AspNetDiagnosticTelemetryModuleTest.cs
+++ b/Src/Web/Web.Net45.Tests/AspNetDiagnosticTelemetryModuleTest.cs
@@ -314,7 +314,10 @@
 
             AspNetDiagnosticTelemetryModule result = new AspNetDiagnosticTelemetryModule();
 
-            var requestModule = new RequestTrackingTelemetryModule(enableChildRequestsSuppression: false);
+            var requestModule = new RequestTrackingTelemetryModule()
+            {
+                EnableChildRequestTrackingSuppression = false
+            };
             var exceptionModule = new ExceptionTrackingTelemetryModule();
             requestModule.Initialize(this.configuration);
             exceptionModule.Initialize(this.configuration);
--- a/Src/Web/Web.Net45.Tests/AspNetDiagnosticTelemetryModuleTest.cs
+++ b/Src/Web/Web.Net45.Tests/AspNetDiagnosticTelemetryModuleTest.cs
@@ -314,7 +314,7 @@
 
             AspNetDiagnosticTelemetryModule result = new AspNetDiagnosticTelemetryModule();
 
-            var requestModule = new RequestTrackingTelemetryModule(enableSafeRequestTracking: true);
+            var requestModule = new RequestTrackingTelemetryModule(enableChildRequestsSuppression: false);
             var exceptionModule = new ExceptionTrackingTelemetryModule();
             requestModule.Initialize(this.configuration);
             exceptionModule.Initialize(this.configuration);
--- a/Src/Web/Web.Shared.Net.Tests/Helpers/HttpModuleHelper.cs
+++ b/Src/Web/Web.Shared.Net.Tests/Helpers/HttpModuleHelper.cs
@@ -65,13 +65,6 @@
             Thread.GetDomain().SetData(".appPath", string.Empty);
             Thread.GetDomain().SetData(".appVPath", string.Empty);
 
-            if (headers == null)
-            {
-                headers = new Dictionary<string, string>();
-            }
-
-            headers["AppInsights-RequestTrackingTelemetryModule-Request-Id"] = "test-request-id";
-
             var workerRequest = new SimpleWorkerRequestWithHeaders(UrlPath, UrlQueryString, new StringWriter(CultureInfo.InvariantCulture), headers, remoteAddr);
             
             var context = new HttpContext(workerRequest);
--- a/Src/Web/Web.Shared.Net.Tests/Helpers/HttpModuleHelper.cs
+++ b/Src/Web/Web.Shared.Net.Tests/Helpers/HttpModuleHelper.cs
@@ -65,13 +65,20 @@
             Thread.GetDomain().SetData(".appPath", string.Empty);
             Thread.GetDomain().SetData(".appVPath", string.Empty);
 
+            if (headers == null)
+            {
+                headers = new Dictionary<string, string>();
+            }
+
+            headers["AppInsights-RequestTrackingTelemetryModule-Request-Id"] = "test-request-id";
+
             var workerRequest = new SimpleWorkerRequestWithHeaders(UrlPath, UrlQueryString, new StringWriter(CultureInfo.InvariantCulture), headers, remoteAddr);
             
             var context = new HttpContext(workerRequest);
             HttpContext.Current = context;
             return context;
         }
-
+        
         public static HttpContextBase GetFakeHttpContextBase(IDictionary<string, string> headers = null)
         {
             return new HttpContextWrapper(GetFakeHttpContext(headers));
--- a/Src/Web/Web.Shared.Net.Tests/Helpers/HttpModuleHelper.cs
+++ b/Src/Web/Web.Shared.Net.Tests/Helpers/HttpModuleHelper.cs
@@ -71,7 +71,7 @@
             HttpContext.Current = context;
             return context;
         }
-        
+
         public static HttpContextBase GetFakeHttpContextBase(IDictionary<string, string> headers = null)
         {
             return new HttpContextWrapper(GetFakeHttpContext(headers));
--- a/Src/Web/Web.Shared.Net/Implementation/WebEventSource.cs
+++ b/Src/Web/Web.Shared.Net/Implementation/WebEventSource.cs
@@ -4,7 +4,6 @@
 #if NET45
     using System.Diagnostics.Tracing;
 #endif
-    using System.Globalization;
 
     /// <summary>
     /// ETW EventSource tracing class.
@@ -374,7 +373,7 @@
                 38,
                 this.ApplicationName);
         }
-        
+
         [NonEvent]
         private string GetApplicationName()
         {
--- a/Src/Web/Web.Shared.Net/Implementation/WebEventSource.cs
+++ b/Src/Web/Web.Shared.Net/Implementation/WebEventSource.cs
@@ -4,8 +4,7 @@
 #if NET45
     using System.Diagnostics.Tracing;
 #endif
-    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
-
+    
     /// <summary>
     /// ETW EventSource tracing class.
     /// </summary>
@@ -34,13 +33,13 @@
          }
 
         [Event(
-            1, 
-            Message = "ApplicationInsightsHttpModule failed at initialization with exception: {0}", 
+            1,
+            Message = "ApplicationInsightsHttpModule failed at initialization with exception: {0}",
             Level = EventLevel.Error)]
         public void WebModuleInitializationExceptionEvent(string excMessage, string appDomainName = "Incorrect")
         {
             this.WriteEvent(
-                1, 
+                1,
                 excMessage ?? string.Empty,
                 this.ApplicationName);
         }
@@ -78,7 +77,7 @@
         public void HanderFailure(string exception, string appDomainName = "Incorrect")
         {
             this.WriteEvent(
-                4, 
+                4,
                 exception ?? string.Empty,
                 this.ApplicationName);
         }
@@ -158,8 +157,8 @@
             string cookieValue, string appDomainName = "Incorrect")
         {
             this.WriteEvent(
-                14, 
-                cookieValue ?? string.Empty, 
+                14,
+                cookieValue ?? string.Empty,
                 this.ApplicationName);
         }
 
@@ -168,11 +167,11 @@
             Message = "WebTelemetryInitializerLoaded at {0}",
             Level = EventLevel.Verbose)]
         public void WebTelemetryInitializerLoaded(
-            string typeName, 
+            string typeName,
             string appDomainName = "Incorrect")
         {
             this.WriteEvent(
-                16, 
+                16,
                 typeName ?? string.Empty,
                 this.ApplicationName);
         }
@@ -193,12 +192,12 @@
             Level = EventLevel.Error)]
         public void WebTelemetryInitializerFailure(
             string typeName,
-            string exception, 
+            string exception,
             string appDomainName = "Incorrect")
         {
             this.WriteEvent(
                 18,
-                typeName ?? string.Empty, 
+                typeName ?? string.Empty,
                 exception ?? string.Empty,
                 this.ApplicationName);
         }
@@ -379,12 +378,12 @@
             39,
             Message = "RequestTrackingTelemetryModule ChildRequestTrackingSuppressionModule Method: '{0}' Unknown Exception: {1}",
             Level = EventLevel.Error)]
-        public void ChildRequestUnknownException(string methodName, Exception ex, string appDomainName = "Incorrect")
+        public void ChildRequestUnknownException(string methodName, string exceptionMessage, string appDomainName = "Incorrect")
         {
             this.WriteEvent(
                 39,
                 methodName,
-                ex.ToInvariantString(),
+                exceptionMessage,
                 this.ApplicationName);
         }
 
@@ -392,7 +391,7 @@
             40,
             Message = "RequestTrackingTelemetryModule: Request was not logged. Set EventLevel Verbose for more details.",
             Level = EventLevel.Informational)]
-        public void RequestTrackingTelemetryModule_RequestWasNotLogged_Informational(string appDomainName = "Incorrect")
+        public void RequestTrackingTelemetryModuleRequestWasNotLoggedInformational(string appDomainName = "Incorrect")
         {
             this.WriteEvent(
                 40,
@@ -403,7 +402,7 @@
             41,
             Message = "RequestTrackingTelemetryModule: Request was not logged. Request Id: '{0}' Reason: {1}",
             Level = EventLevel.Verbose)]
-        public void RequestTrackingTelemetryModule_RequestWasNotLogged_Verbose(string requestId, string reason, string appDomainName = "Incorrect")
+        public void RequestTrackingTelemetryModuleRequestWasNotLoggedVerbose(string requestId, string reason, string appDomainName = "Incorrect")
         {
             this.WriteEvent(
                 41,
--- a/Src/Web/Web.Shared.Net/Implementation/WebEventSource.cs
+++ b/Src/Web/Web.Shared.Net/Implementation/WebEventSource.cs
@@ -394,7 +394,9 @@
         {
             this.WriteEvent(
                 40,
-                numMaxActiveRequests.ToString(CultureInfo.InvariantCulture), timeoutSetting.ToString(CultureInfo.InvariantCulture), removedItems.ToString(CultureInfo.InvariantCulture),
+                numMaxActiveRequests.ToString(CultureInfo.InvariantCulture),
+                timeoutSetting.ToString(CultureInfo.InvariantCulture),
+                removedItems.ToString(CultureInfo.InvariantCulture),
                 this.ApplicationName);
         }
 
--- a/Src/Web/Web.Shared.Net/Implementation/WebEventSource.cs
+++ b/Src/Web/Web.Shared.Net/Implementation/WebEventSource.cs
@@ -4,6 +4,7 @@
 #if NET45
     using System.Diagnostics.Tracing;
 #endif
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
 
     /// <summary>
     /// ETW EventSource tracing class.
@@ -371,6 +372,43 @@
         {
             this.WriteEvent(
                 38,
+                this.ApplicationName);
+        }
+
+        [Event(
+            39,
+            Message = "RequestTrackingTelemetryModule ChildRequestTrackingSuppressionModule Method: '{0}' Unknown Exception: {1}",
+            Level = EventLevel.Error)]
+        public void ChildRequestUnknownException(string methodName, Exception ex, string appDomainName = "Incorrect")
+        {
+            this.WriteEvent(
+                39,
+                methodName,
+                ex.ToInvariantString(),
+                this.ApplicationName);
+        }
+
+        [Event(
+            40,
+            Message = "RequestTrackingTelemetryModule: Request was not logged. Set EventLevel Verbose for more details.",
+            Level = EventLevel.Informational)]
+        public void RequestTrackingTelemetryModule_RequestWasNotLogged_Informational(string appDomainName = "Incorrect")
+        {
+            this.WriteEvent(
+                40,
+                this.ApplicationName);
+        }
+
+        [Event(
+            41,
+            Message = "RequestTrackingTelemetryModule: Request was not logged. Request Id: '{0}' Reason: {1}",
+            Level = EventLevel.Verbose)]
+        public void RequestTrackingTelemetryModule_RequestWasNotLogged_Verbose(string requestId, string reason, string appDomainName = "Incorrect")
+        {
+            this.WriteEvent(
+                41,
+                requestId,
+                reason,
                 this.ApplicationName);
         }
 
--- a/Src/Web/Web.Shared.Net/Implementation/WebEventSource.cs
+++ b/Src/Web/Web.Shared.Net/Implementation/WebEventSource.cs
@@ -4,6 +4,7 @@
 #if NET45
     using System.Diagnostics.Tracing;
 #endif
+    using System.Globalization;
 
     /// <summary>
     /// ETW EventSource tracing class.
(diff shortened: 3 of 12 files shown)


Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -309,6 +323,40 @@ internal void OverrideCorrelationIdLookupHelper(CorrelationIdLookupHelper correl
...
+            {
+                requestId = Guid.NewGuid().ToString();
+                context.Request.Headers[RequestIdHeader] = requestId;
+            }
+            else
+            {
+                requestId = context.Request.Headers[RequestIdHeader];
+            }
+            
+            if (!activeRequests.ContainsKey(requestId))
```
SergeyKanzhelev: talked to @iusafaro. If this is for the test purposes only - I'd suggest to expose a protected Boolean flag that allows to disable this functionality. In reality such flag may be useful even as a public one (at least for the beta release) as we do know know if every version will allow to write to the header,

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -16,13 +17,34 @@
     /// </summary>
     public class RequestTrackingTelemetryModule : ITelemetryModule
     {
+        private const string RequestIdHeader = "AppInsights-RequestTrackingTelemetryModule-Request-Id";
```
SergeyKanzhelev: Please refrain from using `AppInsights` in favor to `ApplicationInsights` where possible

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -309,6 +337,50 @@ internal void OverrideCorrelationIdLookupHelper(CorrelationIdLookupHelper correl
...
+        /// When the child request OnEndRequest, the id will be removed from this dictionary and telemetry will not be logged for the parent.
+        /// </see>
+        /// </summary>
+        /// <remarks>
+        /// Unit test projects cannot create an [internal] IIS7WorkerRequest object.
+        /// Without this object, we cannot modify the Request.Headers without throwing a PlatformNotSupportedException.
+        /// Unit tests will have to initialize the RequestIdHeader.
+        /// The second IF will ensure the id is added to the activeRequests.
+        /// </remarks>
+        private void EnsureIsActiveRequestSafe(HttpContext context)
```
SergeyKanzhelev: I don't understand how this method is safe. It still attempts to write to the headers

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -216,7 +240,11 @@ public void OnEndRequest(HttpContext context)
                 requestTelemetry.Source = telemetrySource;
             }
 
-            this.telemetryClient.TrackRequest(requestTelemetry);
+            string requestId = context.Request.Headers[RequestIdHeader];
+            if (requestId != null && activeRequests.TryRemove(requestId, out byte value))
```
SergeyKanzhelev: Why the failure to remove will lead to not tracked request? It's better to report two requests than zero in case you silently attempted to remove the header.

Another thing - add logging in case it failed to remove.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -309,6 +337,50 @@ internal void OverrideCorrelationIdLookupHelper(CorrelationIdLookupHelper correl
...
+            }
+        }
+
+        private void EnsureIsActiveRequest(HttpContext context)
+        {
+            if (context.Request.Headers[RequestIdHeader] == null)
+            {
+                string requestId = Guid.NewGuid().ToString();
+                context.Request.Headers[RequestIdHeader] = requestId;
+                activeRequests.TryAdd(requestId, 0);
```
SergeyKanzhelev: please protect the collection from memory leak. I'd suggest to check collection size from time to time and purge the entire collecting if it overgrown some huge limit. See [here](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/da068a346377663e76a9b382ddf49e73cf96a1e6/Src/Common/CorrelationIdLookupHelper.cs#L21-L24) for example

Src/Web/Web.Shared.Net.Tests/Helpers/HttpModuleHelper.cs
```diff
@@ -65,13 +65,20 @@ public static HttpContext GetFakeHttpContext(IDictionary<string, string> headers
             Thread.GetDomain().SetData(".appPath", string.Empty);
             Thread.GetDomain().SetData(".appVPath", string.Empty);
 
+            if (headers == null)
+            {
+                headers = new Dictionary<string, string>();
+            }
+
+            headers["AppInsights-RequestTrackingTelemetryModule-Request-Id"] = "test-request-id";
```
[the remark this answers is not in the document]
MS-TimothyMothra: TL;DR: Yes.

This flag wasn't meant to enable/disable child request suppression.
The current flag should say areRequestHeadersWriteable, as you suggested.
The module can't write to request headers created by the unit test framework so any headers still have to be created here.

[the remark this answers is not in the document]
MS-TimothyMothra: ok, I've changed the implementation to enable/disable child request suppression.

Src/Web/Web.Net45.Tests/AspNetDiagnosticTelemetryModuleTest.cs
```diff
@@ -314,7 +314,7 @@ private AspNetDiagnosticTelemetryModule CreateModule(string rootIdHeaderName = n
 
             AspNetDiagnosticTelemetryModule result = new AspNetDiagnosticTelemetryModule();
 
-            var requestModule = new RequestTrackingTelemetryModule();
+            var requestModule = new RequestTrackingTelemetryModule(enableSafeRequestTracking: true);
```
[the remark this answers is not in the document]
MS-TimothyMothra: ok, I've changed the implementation to enable/disable child request suppression.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -22,6 +23,31 @@ public class RequestTrackingTelemetryModule : ITelemetryModule
         private bool correlationHeadersEnabled = true;
         private string telemetryChannelEnpoint;
         private CorrelationIdLookupHelper correlationIdLookupHelper;
+        private ChildRequestTrackingSuppressionModule childRequestTrackingSuppressionModule = null;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RequestTrackingTelemetryModule" /> class.
+        /// </summary>
+        public RequestTrackingTelemetryModule() : this(enableChildRequestsSuppression: true)
```
Dmitry-Matveev: This looks like a change in Public API, we would need to mention it in Changelog and in the Public API reference file (we have one in dotnet repo, not sure if there is one in dotnet-server repo, though...).
We might want to consider other implementations if this exposure is not necessary and/or make it configurable.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -347,5 +378,79 @@ private bool TryInitializeCorrelationHelperIfNotInitialized()
...
+        /// IIS Classic Pipeline should disable the ChildRequestTrackingSuppressionModule.
+        /// Classic does not create IIS7WorkerRequest object and Headers will be read-only.
+        /// (Exception System.PlatformNotSupportedException: This operation requires IIS integrated pipeline mode.)
+        /// </remarks>
+        private class ChildRequestTrackingSuppressionModule
+        {
+            /// <summary>
+            /// Max number of request ids to cache.
+            /// </summary>
+            private const int MAXSIZE = 100;
```
Dmitry-Matveev: Couple comments:
- 100 concurrent requests seems like an easy target to reach in heavily loaded applications, so the limit should be bumped;
- When the limit is reached (e.g. some `OnBegin` didn't have `OnEnd`, or some exception happened), it looks like we're clearing up all the active requests and `ShouldTrackRequest()` will decide not to track both requests (parent and child). Let's sync up offline on this and see if we can do something about it.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
MS-TimothyMothra: The public constructor is unchanged, the variable is passed to an internal constructor.
I don't think any documentation would be required.

I like the idea of making it configurable, let's discuss.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -347,5 +378,79 @@ private bool TryInitializeCorrelationHelperIfNotInitialized()
...
+            
+            /// <summary>
+            /// A request must be tracked as Active in order for telemetry to be recorded within OnEndRequest().
+            /// </summary>
+            internal void EnsureActiveRequestIsTracked(HttpContext context)
+            {
+                // Simplistic cleanup to prevent memory leaks.
+                if (activeRequests.Count >= MAXSIZE)
+                {
+                    activeRequests.Clear();
```
cijothomas: ad a log

Src/Web/Web.Shared.Net/Implementation/WebEventSource.cs
```diff
@@ -385,6 +386,18 @@ public void FailedToRemoveRequestFromActiveRequests(string appDomainName = "Inco
                 this.ApplicationName);
         }
 
+        [Event(
+            40,
+            Message = "RequestTrackingTelemetryModule ChildRequestTrackingSuppressionModule exceeded max number of '{0}' active requests. Timeout set to '{1}'. Clearing '{2}' items from dictionary.",
+            Level = EventLevel.Informational)]
```
cijothomas: Do mention the unit of time here. seconds? msecs?

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -402,14 +402,16 @@ private class ChildRequestTrackingSuppressionModule
...
+            private const int TIMEOUTSECONDS = -3;
+            
             private const string RequestIdHeader = "ApplicationInsights-RequestTrackingTelemetryModule-Request-Id";
 
             /// <summary>
-            /// Using this as a hash-set of current active requests. The second value is ignored.
+            /// Using this as a hash-set of current active requests. The second value is used when auditing the dictionary for cleanup.
             /// </summary>
-            private static ConcurrentDictionary<string, byte> activeRequests = new ConcurrentDictionary<string, byte>();
+            private static ConcurrentDictionary<string, long> activeRequests = new ConcurrentDictionary<string, long>(System.Environment.ProcessorCount, MAXSIZE);
```
cijothomas: is there Risk of an exception being thrown when trying to read System. ? If yes this code is not safe. Better to do this inside try catch.

Src/Web/Web.Shared.Net/Implementation/WebEventSource.cs
cijothomas: The unit of time used should also be in the log message. seconds or msec..

SergeyKanzhelev:
@Dmitry-Matveev explained to me the problem with the dictionary clean up on high load. The proposed solution was to keep the "freshness" for the every item in the dictionary. It doesn't work as cleaning up of the dictionary on timeout requires too much CPU.

There are two alternative solutions:

### Change design

This solution relies on two facts:
- `End` is guaranteed to call on request
- There is only one child request possible

Solution is to add id to the dictionary in the `End` of the child request. This dictionary will represent already tracked requests. Then parent request will check that id already present and will erase it from the dictionary.

Alternative to this approach is never delete items from dictionary. Instead rotate two dictionaries. That leads to the second possible solution.

The beauty of this solution is that if we reset request that we didn't intend to - we will track two requests instead of not tracking either.

## Change the way big cache is reset

This approach works as follow - when you noticed that the dictionary is too big, you set it aside for the timeout period of time (say 5 minutes) and create a new dictionary side by side.

For this timeout period of time you adding elements into the new dictionary, but check existence and remove from both.

The goodness of this approach - it triggers extra work only when there is an overflow of the working dictionary (supposed to be rare). And it doesn't require iterating over the full dictionaty and remove expired requests.


**As I mentioned you can combing two approaches**

Dmitry-Matveev:
@MS-TimothyMothra , can you please a second look into the Sergey's comment around sending "two requests instead of not tracking either" at the edge case when cleaning up the dictionaries. It looks like we can combine the approaches and be OK with memory and CPU while occasionally tracking more requests when needed.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -402,69 +402,100 @@ private class ChildRequestTrackingSuppressionModule
...
+                }
+
+                return false;
+            }
+
+            /// <summary>
+            /// Tag new requests.
+            /// Transfer Ids to parent requests.
+            /// </summary>
+            private string TagRequest(HttpContext context)
```
Dmitry-Matveev: Do we plan to use return value in the future? If not, this can be safely removed.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -402,69 +402,100 @@ private class ChildRequestTrackingSuppressionModule
...
+                return activeRequestsA.TryGetValue(requestId, out bool valueA) || activeRequestsB.TryGetValue(requestId, out bool valueB);
+            }
 
-                if (requestId != null && activeRequests.ContainsKey(requestId))
+            /// <summary>
+            /// Track this requestId.
+            /// </summary>
+            private void AddRequestToDictionary(string requestId)
+            {
+                // TODO: LOCK
```
Dmitry-Matveev: Can this be implemented with `Interlocked.CompareExchange()` to avoid lock and still be thread-safe?

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
MS-TimothyMothra: Thanks, an earlier implementation was populating the tracking dictionary in OnBeginRequest. Removed.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
SergeyKanzhelev: You can make it configurable by exposing public property. It is a good idea so if something goes south on specific environment we can shut down this logic completely

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -69,6 +95,8 @@ internal string EffectiveProfileQueryEndpoint
         /// </summary>
         public void OnBeginRequest(HttpContext context)
         {
+            this.childRequestTrackingSuppressionModule?.OnBeginRequest_IdRequest(context);
```
SergeyKanzhelev: I think you can safely move this code to after `this.initializationErrorReported` check.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -347,5 +378,125 @@ private bool TryInitializeCorrelationHelperIfNotInitialized()
...
+        /// </remarks>
+        private class ChildRequestTrackingSuppressionModule
+        {
+            /// <summary>
+            /// Max number of request ids to cache.
+            /// </summary>
+            private const int MAXSIZE = 100000; // TODO: THIS VALUE NEEDS TO BE CONFIGURABLE
+
+            private const string HeaderRootRequestId = "ApplicationInsights-RequestTrackingTelemetryModule-RootRequest-Id";
+            private const string HeaderParentRequestId = "ApplicationInsights-RequestTrackingTelemetryModule-ParentRequest-Id";
```
SergeyKanzhelev: should be defined under `#if DEBUG` to make it clearer why we need those

Src/Web/Web.Shared.Net/Implementation/WebEventSource.cs
```diff
@@ -374,6 +375,31 @@ public void InitializeHasNotBeenCalledOnModuleYetVerbose(string appDomainName =
                 this.ApplicationName);
         }
 
+        [Event(
+            39,
+            Message = "RequestTrackingTelemetryModule ChildRequestTrackingSuppressionModule failed to remove request from ActiveRequests.",
+            Level = EventLevel.Warning)]
+        public void FailedToRemoveRequestFromActiveRequests(string appDomainName = "Incorrect")
```
SergeyKanzhelev: is it ever used?

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -347,5 +378,125 @@ private bool TryInitializeCorrelationHelperIfNotInitialized()
...
+            private void AddRequestToDictionary(string requestId)
+            {
+                // TODO: LOCK
+                if (activeRequestsA.Count >= MAXSIZE)
+                {
+                    activeRequestsB = activeRequestsA;
+                    activeRequestsA = new ConcurrentDictionary<string, bool>(System.Environment.ProcessorCount, MAXSIZE);
+                }
+
+                activeRequestsA.TryAdd(requestId, false);
```
SergeyKanzhelev: Do you ever remove items from the dictionary?

MS-TimothyMothra: The intention was not to because the removal was adding extra overhead.

SergeyKanzhelev: Typically there are a few requests running in parallel. Let's say up to 100. It may be cheaper to add/remove this hundred than insert into very big hashtable. When I meant never delete - I meant never delete expired IDs, drop the full hashtable instead.

As you mentioned - it will be great to run some small synthetic tests to understand perf implications of both approaches.

MS-TimothyMothra: When I started this, I wrote a small synthetic test of multiple threads adding and removing entries into a single concurrentdictionary and it had a SIGNIFICANT cpu overhead.

Trusting that we can determine when a request ends requires a few assumptions, and I've been able to find antidotal examples that disprove each assumption. I would be happy to discuss further.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
MS-TimothyMothra: TL;DR: Nope.

Discussed with Dmitry offline, I'm going to avoid this suggestion.

CompareExchange involves a; target, compareEqualityValue, and newValue.
That would wipe out dictionary A before I could set it to B.
And I can't have the B be a target, based on a comparison of dictionary A.

Also, the strict equality comparison makes me nervous because if two concurrent threads are able to write to dictionary A and push it over the MAX, we would never catch this condition.

MS-TimothyMothra:
@SergeyKanzhelev please re-review

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -415,8 +416,24 @@ private class ChildRequestTrackingSuppressionModule
...
+
+            /// <summary>
+            /// Initializes a new instance of the <see cref="ChildRequestTrackingSuppressionModule" /> class.
+            /// </summary>
+            /// <param name="maxRequestsTracked">The maximum number of active requests to be tracked before resetting the dictionary.</param>
+            internal ChildRequestTrackingSuppressionModule(int maxRequestsTracked = DEFAULTMAXVALUE)
+            {
+                this.MAXSIZE = maxRequestsTracked > 0 ? maxRequestsTracked : DEFAULTMAXVALUE;
+
+                System.Diagnostics.Debug.WriteLine($"{nameof(RequestTrackingTelemetryModule)}.{nameof(ChildRequestTrackingSuppressionModule)} Initialized. {nameof(this.MAXSIZE)}:{this.MAXSIZE}");
```
SergeyKanzhelev: haven't noticed before. We are using EventSource verbose messages for this king of messages.

MS-TimothyMothra:
@cijothomas, @iusafaro Can you please re-review and unblock? :)

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
MS-TimothyMothra: I will remove it because it's not needed anymore.

Src/Web/Web.Shared.Net/Implementation/WebEventSource.cs
```diff
@@ -373,7 +374,7 @@ public void InitializeHasNotBeenCalledOnModuleYetVerbose(string appDomainName =
                 38,
                 this.ApplicationName);
         }
-
```
cijothomas: no change in this file right?

MS-TimothyMothra: I'll revert these changes since they're not needed anymore

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -347,5 +382,151 @@ private bool TryInitializeCorrelationHelperIfNotInitialized()
...
+                return false;
+            }
+
+            /// <summary>
+            /// Tag new requests.
+            /// Transfer Ids to parent requests.
+            /// </summary>
+            private void TagRequest(HttpContext context)
+            {
+                var newId = Guid.NewGuid().ToString();
```
cijothomas: minor. Guid need to be generate only if header is already null. so you can move this to inside if {}

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -347,5 +382,151 @@ private bool TryInitializeCorrelationHelperIfNotInitialized()
...
+
+                var headers = context.Request.Headers;
```
cijothomas: i have seen in other places we do check for Request == NULL. Or do we know for sure that Request will not be NULL here?

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -347,5 +382,151 @@ private bool TryInitializeCorrelationHelperIfNotInitialized()
...
+            /// <summary>
+            /// OnEndRequest - Should this request be logged?
+            /// Will compare a request id against a hash-set of known requests.
+            /// If this request is not known, add it to hash-set and return true (safe to log).
+            /// If this request is known, return false (do not log twice).
+            /// Additional requests with the same id will return false.
+            /// </summary>
+            internal bool OnEndRequest_ShouldLog(HttpContext context)
+            {
+                var headers = context.Request.Headers;
```
cijothomas: should you check for Request!=null ?

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -216,7 +241,10 @@ public void OnEndRequest(HttpContext context)
                 requestTelemetry.Source = telemetrySource;
             }
 
-            this.telemetryClient.TrackRequest(requestTelemetry);
+            if (this.childRequestTrackingSuppressionModule?.OnEndRequest_ShouldLog(context) ?? true)
+            {
+                this.telemetryClient.TrackRequest(requestTelemetry);
+            }
```
cijothomas: Please add else {} block, and log that we (childRequestTrackingSuppressionModule) are dropping this request.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -347,5 +382,151 @@ private bool TryInitializeCorrelationHelperIfNotInitialized()
...
+                {
+                    if (!this.IsRequestKnown(rootRequestId))
+                    {
+                        // doesn't exist add to dictionary and return true
+                        this.AddRequestToDictionary(rootRequestId);
+                        return true;
+                    }
+                }
+
+                return false;
```
cijothomas: I strongly recommend writing something to eventlog whenever we are dropping requests. Without this, it'll be quite difficult if some issues occurs about 'requests missing'.

cijothomas: The above return false occurs if request is not tagged. So please add log about it.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -347,5 +382,151 @@ private bool TryInitializeCorrelationHelperIfNotInitialized()
...
+            /// Gets the Max number of request ids to cache.
+            /// </summary>
+            internal int MAXSIZE { get; private set; }
+
+            /// <summary>
+            /// Request will be tagged with an id to identify if it should be logged later.
+            /// </summary>
+            internal void OnBeginRequest_IdRequest(HttpContext context)
+            {
+                this.TagRequest(context);
```
cijothomas: both begin and end methods in this module should be try..catch..log.. to be safe

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -22,8 +23,30 @@ public class RequestTrackingTelemetryModule : ITelemetryModule
...
+        public bool EnableChildRequestTrackingSuppression { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets a value indicating the size of internal tracking dictionary.
+        /// Must be a positive integer.
+        /// </summary>
+        /// <remarks>
+        /// See also <see cref="ChildRequestTrackingSuppressionModule" />.
+        /// </remarks>
+        public int ChildRequestTrackingInternalDictionarySize { get; set; } = -1;
```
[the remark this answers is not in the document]
MS-TimothyMothra: i gave this property an explicit default value, but it's not needed.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -347,5 +382,151 @@ private bool TryInitializeCorrelationHelperIfNotInitialized()
...
+
+            /// <summary>
+            /// Track this requestId.
+            /// </summary>
+            /// <remarks>
+            /// Dictionary A will be read/write.
+            /// When dictionary A is full, move to B and create new A.
+            /// Dictionary B will be read-only.
+            /// </remarks>
+            private void AddRequestToDictionary(string requestId)
```
Dmitry-Matveev: We were discussing with @SergeyKanzhelev offline on how to limit the lookup time on this dictionary when it grows big enough to have hash collisions but not yet `MAXSIZE` and also on how to limit memory impact of the current solution (let's consider these suggestions in the different enhancement with lower priority):
- We can add `remove()` back but without the complex logic to handle all the special cases by sacrificing some of those;
- We can store Base64 of GUID string in the dictionary or any other non-string value that will be more memory-friendly (e.g. GUID object itself);

PR #703, opened 2017-10-05
testconfig increase max allowed runtime

MS-TimothyMothra:
tests are failing that exceed a max runtime. increasing max.


cijothomas:
Thanks @MS-TimothyMothra . merging now to get all pending builds to green.!

PR #715, opened 2017-10-13
upgrade Base SDK to 2.5 beta 1

MS-TimothyMothra:


MS-TimothyMothra:
The upgrade made several changes to the config in the test apps. 
I'm going to let the tests run here and see if these changes need to be removed.

cijothomas:
Use upgradeVersion.ps1 to upgrade, else its hard to see if all projects are correctly updated.

PR #721, opened 2017-10-24
Dependency Collector Functional Tests new infra

cijothomas:
Replaced the test infra for Dependency Collector Functional Tests. The new infra uses Docker to spin up dependencies, and hence is self-contained.

Pending:
non-hacky way to get application logs from inside Docker.
asp.net core tests not yet added.


Test/E2ETests/E2ETests/Test452Base.cs
```diff
@@ -0,0 +1,1009 @@
...
+                DockerUtils.RestartDockerContainer(app.containerName);
+            }
+        }
+
+        public static void MyClassCleanupBase()
+        {
+            Trace.WriteLine("Started Class Cleanup:" + DateTime.UtcNow.ToLongTimeString());
+            //DockerUtils.ExecuteDockerComposeCommand("down", DockerComposeFileName);
+            //DockerUtils.RemoveDockerImage(Apps[WebAppName].imageName, true);
+            Trace.WriteLine("Completed Class Cleanup:" + DateTime.UtcNow.ToLongTimeString());
```
MS-TimothyMothra: should these lines be commented out?

Test/E2ETests/E2ETests/Helpers/TelemetryItemFactory.cs
```diff
@@ -0,0 +1,93 @@
...
+
+                    case TelemetryItemType.Message:
+                        {
+                            result = JsonConvert.DeserializeObject<TelemetryItem<MessageData>>(content);
+                            break;
+                        }
+
+                    default:
+                        {
+                            throw new InvalidDataException("Unsupported telemetry type");
```
MS-TimothyMothra: Consider adding the type name to both exceptions. 
`throw new InvalidDataException($"Unsupported telemetry type: '{envelope.data.baseType.ToString()}");`

Test/E2ETests/E2ETests/docker-compose452AppOn462.yml
```diff
@@ -0,0 +1,37 @@
...
+
+    azureemulator:
+        build:
+            context: ./AzureEmulator
+            dockerfile: DockerFile        
+
+    sql-server:
+        image: microsoft/mssql-server-windows-express
+        environment:            
+            sa_password: MSDNm4g4z!n4
```
MS-TimothyMothra: Something to consider, but is probably a low priority.

Instead of hardcoding the password, I think you could make this value an argument in the yml. And then when the yml is executed in the ClassInitialize, it could read the value from a constant. 

My concern is if this ever needs to be changed, will have to search for all uses of this string.
That may not ever be an issue, but I wanted to bring it up for consideration. :)

Test/E2ETests/E2ETests/Helpers/TelemetryItemFactory.cs
cijothomas: Leaving as it as from current code.

Test/E2ETests/E2ETests/docker-compose452AppOn462.yml
cijothomas: its a valid point though we dont need to change password.
I am moving EVERY constants off to a new Connstants project. Will address this then.

Test/E2ETests/E2ETests/Test452Base.cs
cijothomas: yes intentionally did that for speedy local runs. in the lab, a script is run to ensure full cleanups after every builds.

Test/E2ETests/E2ETests/Test452Base.cs
```diff
@@ -0,0 +1,1066 @@
...
+            }
+        }
+
+        private static void InitializeDatabase()
+        {
+            var ip = DockerUtils.FindIpDockerContainer("e2etests_sql-server_1");
+            LocalDbHelper localDbHelper = new LocalDbHelper(ip);
+            if (!localDbHelper.CheckDatabaseExists("dependencytest"))
+            {
+                localDbHelper.CreateDatabase("dependencytest", "c:\\dependencytest.mdf");
```
MS-TimothyMothra: This is one of the problems I was trying to fix in my PR.

Please add a follow-up check to determine if the CreateDatabase failed.
Otherwise it will continue to execute all the SQL tests and we would rather this fail immediately.

```
if (!LocalDbHelper.CheckDatabaseExists("dependencytest")) 
{ 
throw new Exception($"Failed to create database: 'dependencytest'"); 
}
```

cijothomas: If any sql error occurs, exception is not handled and would cause test initialize to fail. No tests will be run at all.

cijothomas: if (!LocalDbHelper.CheckDatabaseExists("dependencytest")) 
{ 
throw new Exception($"Failed to create database: 'dependencytest'"); 
}

This check can be added to make it explciity why test failed. Will do

cijothomas: added

SergeyKanzhelev:
LGTM, can you please update `CONTRIBUTING.md` as part of this PR

cijothomas:
@SergeyKanzhelev i've update Contributing.md to reflect usage of Docker.

cijothomas:
Ping...

cijothomas:
thanks!

Issue #723, opened 2017-10-30
When debugging netcoreapp2.0 in VS, http dependencies are tracked twice


lmolkova:
HttpClient in 2.0 fires deprecated and 'new' events.

When debugging, VS subscribes to deprecated events, AppInsights subscribes to 'new' events from 2.0 client (but also listens to old events to support older versions).

AppInsights attempts to filter out old events for 2.0 clients (https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs#L560) with IsEnabled callback.
However when there are 2 listeners for diagnostics source, it's enough if just one of them returns true for IsEnabled, then all listeners get the notification.

So, `isNetCore20HttpClient` check should be done in `OnRequest` and `OnResponse` callbacks too.

SergeyKanzhelev:
So you receive every http call twice?

lmolkova:
@SergeyKanzhelev @Dmitry-Matveev @zakimaksyutov @cijothomas how far are you guys from releasing 2.5? Could we include the fix for it? Should be pretty straightforward and small fix.

lmolkova:
@SergeyKanzhelev exactly, all http calls are tracked twice when debugging in VS

Dmitry-Matveev:
@lmolkova , 2.5-Beta1 is around the corner. 2.5 stable was initially planned for this quarter, but no certain date yet. 2.5-Beta2 is the closest and was planned for November.

lmolkova:
#848

PR #728, opened 2017-11-03
Ignore Deprecated events if running under netcore20

cijothomas:
Ignore Deprecated events if running under netcore20 as it publishes newer events.

Attempt to fix https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/723

### DIFF
--- a/Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs
+++ b/Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs
@@ -33,6 +33,7 @@ namespace Microsoft.ApplicationInsights.DependencyCollector.Implementation
         private readonly bool setComponentCorrelationHttpHeaders;
         private readonly ICorrelationIdLookupHelper correlationIdLookupHelper;
         private readonly TelemetryClient client;
+        private bool isNetCore20HttpClient;
         private readonly TelemetryConfiguration configuration;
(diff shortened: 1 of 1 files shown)


cijothomas:
@lmolkova i was playing with this and ended up with this! Can you see if this is the fix indeed?

Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs
```diff
@@ -66,6 +69,9 @@ internal class HttpCoreDiagnosticSourceListener : IObserver<KeyValuePair<string,
             this.client = new TelemetryClient(configuration);
             this.client.Context.GetInternalContext().SdkVersion = SdkVersionUtils.GetSdkVersion("rdd" + RddSource.DiagnosticSourceCore + ":");
 
+            var httpClientVersion = typeof(HttpClient).GetTypeInfo().Assembly.GetName().Version;
+            this.isNetCore20HttpClient = httpClientVersion.CompareTo(new Version(4, 2)) >= 0;
```
lmolkova: perhaps you can also remove this code from subscriber and pass isNetCore20HttpClient  to subscriber's constructor

cijothomas: right.. less code duplication :)

Issue #724, opened 2017-10-31
"An item with the same key has already been added." when the same request is submitted more than once.


Pvlerick:
Using [Polly](https://github.com/App-vNext/Polly) for retry policy inside a ```DelegateHandler``` (in essence, re-running ```SendAsync``` with the same request a certain number of time), I started to have this exception popping up in Visual Studio.

Turns out that when debugging a netcoreapp2.0 project in VS, ApplicationInsights is automatically injected (see https://github.com/aspnet/Home/issues/2051).

This exception is raised here: https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared/Implementation/HttpCoreDiagnosticSourceListener.cs#L383 since the same request has been already sent in the same channel. There should be a check to update the content instead of a blind Add.

MS-TimothyMothra:
That's definitely a bug. Thanks for reporting it!

In your opinion, how would you expect the retry behavior to be logged:
Would you expect the original (failed), the retry (succeed), or both to be logged?

SergeyKanzhelev:
@lmolkova, is this a dup of https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/723?

Pvlerick:
@lmolkova That's a good question...

I think in that case as the retries are at the application level so AI should log each request distinctly. But that's only my opinion :-)

cijothomas:
#723 is separate and fix is submitted for it separately.

MS-TimothyMothra:
@Pvlerick, can you please share a code snippet of how you are making requests using Polly?
I haven't been able to reproduce your issue.
Depending on my setup, either the duplicate requests succeeds or HttpClient blocks the duplicate request.

Thank you

Pvlerick:
Here is the actual code.

The policy is build like this (you will need to replace a few things that are coming from the configuration field):
```csharp
private IAsyncPolicy<HttpResponseMessage> BuildPolicy()
{
    var timeout = Policy
        .TimeoutAsync<HttpResponseMessage>(
            timeout: TimeSpan.FromSeconds(this.configuration.TimeoutInSeconds),
            timeoutStrategy: TimeoutStrategy.Optimistic);

    var retry = Policy
        .Handle<HttpRequestException>()
        .Or<TimeoutRejectedException>()
        .OrResult<HttpResponseMessage>(response => !response.IsSuccessStatusCode)
[...]

PR #731, opened 2017-11-07
bug fix. check if exists before add

MS-TimothyMothra:
#724 
Bug fix for the add if not exists scenario. SDK should not throw an exception.
Enhancement to track both requests will be discussed at a later date.

### DIFF
--- a/Src/DependencyCollector/Shared.Tests/Implementation/DependencyCollectorDiagnosticListenerTests.Netstandard16.cs
+++ b/Src/DependencyCollector/Shared.Tests/Implementation/DependencyCollectorDiagnosticListenerTests.Netstandard16.cs
@@ -122,6 +122,12 @@ namespace Microsoft.ApplicationInsights.Tests
         /// Very second request does not throw an exception.
         /// Verify that telemetry is collected for both requests.
         /// </summary>
+        /// <remarks>
+        /// IGNORE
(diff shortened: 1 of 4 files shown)


MS-TimothyMothra:
@cijothomas please re-review

Src/DependencyCollector/Shared.Tests/Implementation/DependencyCollectorDiagnosticListenerTests.Netstandard16.cs
```diff
@@ -103,6 +103,70 @@ public void OnRequestWithRequestEventWithNoRequestUri()
...
+            {
+                RequestMessage = request
+            };
+
+            this.listener.OnResponse(response1, loggingRequestId);
+            Assert.IsFalse(this.listener.PendingDependencyTelemetry.TryGetValue(request, out dependency));
+            Assert.AreEqual(1, this.sentTelemetry.Count);
+            Assert.AreEqual(false, ((DependencyTelemetry)this.sentTelemetry.Last()).Success); // first request fails
+
+            Assert.Inconclusive();
```
cijothomas: you can just split this into two separte unit tests, and [Ignore] the second one until fix is made. Easier to track [Ignore] tests as thats what we typically do to temp. disable tests.

PR #732, opened 2017-11-08
Github PR template

cijothomas:
Will update the contributing.md in a follow-up PR to contain more in-depth instructions on running and debugging tests locally.


cijothomas:
Contributing.md and other scripts are also update in this same pr itself.

CONTRIBUTING.md
```diff
@@ -1,43 +1,61 @@
...
+to either IISExpress or IIS after deploying the application.
+The test apps refers to the Web SDK assemblies from your local build. After making the changes to product code, build locally (from Visual Studio or using ```buildDebug.cmd```). Then start the test application
+in either IISExpress or IIS, and attach debugger to it. Open the .cs file you want your breakpoint in and set it. Now triggering a request to the application will hit the breakpoint.
+The exact request to be triggered depends on what you are doing. If investigating functional test failures locally, then the tests logs should contain the url it hit to trigger scenarios.
+
+<TODO>
+	Dependency Collector tests deploy the test apps, along with dependencies (Fake Ingestion, SQL etc) to Docker containers inside same network, so that apps can access the dependencies with their names. However, if 
+	the test apps are deployed to IIS or IISExpress, it won't be able to access dependencies without using their IP Address.
+	(Todo is to write a small PS script, which can find ip addresses and replace them in Web.config or applicationinsights.config)
```
cijothomas: a better solution is for the tests to do this automatically. Working on modifying test logic to use actual IP Address to refer to containers. (will be separate PR)

cijothomas:
Yes, separate PR coming today.

Issue #491, opened 2017-04-24
Rename ``FirstChanceExceptionStatisticsTelemetryModule`` to ``FirstChanceExceptionMetricsTelemetryModule`


AlexBulankou:
Suggestion: rename ``FirstChanceExceptionStatisticsTelemetryModule`` to ``FirstChanceExceptionMetricsTelemetryModule``. ``Statistics`` has implication of some statistical analysis that is happening as part of collection, and this is not the case. I think ``Metrics`` is more appropriate reference to what it is emitting.

SergeyKanzhelev:
We are doing both - metrics and examples based on statistical analysis =)

I was thinking to move the logic of pre-aggregation to the processor that is making all other pre-aggregations and generalize "examples" collection logic to other types. I'll rename the issue.

SergeyKanzhelev:
Do you think we need to do it for 2.4 stable?

AlexBulankou:
@SergeyKanzhelev No, but I think if we release it with this name now, next opportunity to make a breaking change (since this public API) will be 3.0, not 2.5. So either 2.4 or 3.0.

ohadschn:
More importantly, this module needs to be documented *somewhere*. There are literally less than 10 results when Binging/Googling `FirstChanceExceptionStatisticsTelemetryModule`, none of which actually explaining what it does. 

In the meantime, I'll paste what the [source](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/8d88ec1787442c6299c688f743d26380e705209f/Src/WindowsServer/WindowsServer.Shared/FirstChanceExceptionStatisticsTelemetryModule.cs#L19) says for the benefit of web searchers getting here:

> The module subscribed to AppDomain.CurrentDomain.FirstChanceException to send exceptions statistics to ApplicationInsights.

SergeyKanzhelev:
It is a part of this story: https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/212 Commented as we rationalizing the story now. I agree, we need to document it. I was planning to do it. Now that you mentioned, I'll try to squeeze it into the next week.

It was enabled on some of our services so it's quite stable. If you'd enable it in your environment - give feedback! Especially important feedback on the balance of "usefullness" vs. "noisiness"

ohadschn:
[...]

PR #735, opened 2017-11-09
Adds ServiceBus dependency and requests tracking

### LINKED ISSUE #740: Track requests and dependencies from ServiceBus
ServiceBus .NET Client (version 2.1.0) vas instrumented with DiagnosticsSource and Activity.

ServiceBus is a messaging (queue) service, it implements AMQP protocol, and cannot be tracked with HTTP dependency collection.

Typical ServiceBus .Net Client scenario usage:
```C#
    // Sends message on producer side
    queueClient.SendAsync(message); 

   // process message on consumer side
    queueClient.RegisterMessageHandler(
                (message, ct) =>  await Process(message),
                (exceptionArgs) => Task.CompletedTask);
```

This scenario translates into the 3 telemetry items to be collected:
- Send - dependency
- Receive - dependency
- Process - request

There are several other instrumented operations, see https://github.com/Azure/azure-service-bus-dotnet/pull/370

Instrumentation can be hooked with `TelemetryDiagnosticSourceListener` generic listener, however, some customization is needed to make sure all useful data is properly collected:

- We need to collect Requests (message processing).
- Correctly map activity properties and diagnostic source events to AI telemetry

----------
# ServiceBus telemetry properties
 
## Dependencies:

SB operations that are tracked as depednency:
- MessageSender (i.e. QueueClient and TopicClient): SendAsync, ScheduleMessage and CancelScheduledMessage
- MessageReceiver: Receive, ReceiveDeffered, Peek, Abandon, Defer, Complete, DeadLetter, RenewLock
MessageSession: AcceptMessageSession
- SubscriptionClient : AddRule, RemoveRule, GetRules

Corresponding dependencies will have following properties:
- *Target*: 'endpoint | entityPath' (sb://myservicebus.servicebus.windows.net/ | myqueue")
- *Type*: 'Azure Service Bus'
- *Name*: Operation name (i.e. `Send` or `AddRule`)

## Requests
Operations that are tracked as requests:
Processing messages with `QueueClient.RegisterMessageHandler` and `QueueClient.RegisterSessionHandler` 

RequestTelemetry will have following properties:
- *Name*:  operation name - 'Process' or 'ProcessSession'
- *Source*: 'type: Azure Service Bus | name:myqueue | endpoint:sb://myservicebus.servicebus.windows.net/'

Telemetry also must have timestamp, duration, correlation identifiers and success (bool) property.

In case if ServiceBus message/operation had MessageId or SessionId associated with it, it will be added to custom properties.

lmolkova:
Fixes #740

Improves ServiceBus dependency and requests collection for (Microsoft.Azure.ServiceBus 2.1.0)
ServiceBus is a messaging (queue) service. ServiceBus .NET Client was recently instrumented with DiagnosticsSource/Activity approach ensuring correlation between consumer and producer of messages.
ServiceBus .NET client uses Amqp protocol, and cannot be tracked with HTTP dependency collection.

Typical ServiceBus .Net Client scenario usage:
```C#
    // Sends message on producer side
    queueClient.SendAsync(message); 

   // process message on consumer side
    queueClient.RegisterMessageHandler(
                (message, ct) =>  await Process(message),
                (exceptionArgs) => Task.CompletedTask);
```

This scenario translates into the 3 telemetry items to be collected:
- Send - dependency
- Receive - dependency
- Process - request

Without this PR: 
- if users install instrumented ServiceBus client and turn corresponding diagnostics source on (OFF by default), they would get raw dependency telemetry (for all operations) with missing Target and Data fields, and Operation names like "Microsoft.Azure.ServiceBus.Send"
- Process operation will be tracked as a dependency, which would break portal experiences that do not support nested dependency calls and also AppMap that is built based on the Request being a child of the Dependency.

With this PR: 
All 3 operations collected without any changes in user code. 
Since serviceBus client injects tracing context (Activity) into the message and restores it for processing, it ensures telemetry is correlated e2e - from consumer to producer. AppMap and other portal views (without any changes) consider queue call like direct HTTP call (with just different dependency type). This way portal experience works as usual.
Multi ikey is not initially supported either way.

Please take to fix into beta2 considering poor (broken) UI experience users would get in absence of it.


Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
```diff
@@ -0,0 +1,171 @@
...
+            Exception ex = (Exception)exceptionFetcher.Fetch(payload);
+
+            this.telemetryClient.TrackException(ex);
+        }
+
+        private void SetCommonProperties(string eventName, object eventPayload, Activity activity, OperationTelemetry telemetry)
+        {
+            // activity name looks like 'Microsoft.Azure.ServiceBus.<Name>'
+            // as namespace is too verbose, we'll just take the last node from the activity name as telemetry name
+            string[] activityNameSegments = activity.OperationName.Split('.');
```
pharring: `string.Split` is expensive. If all you need is the part after the last period, use `LastIndexOf` to find the last period and `Substring` to extract everything after that.

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
```diff
@@ -0,0 +1,171 @@
...
+        // Not knowing particular service
+        private const string DependencyType = "Queue.ServiceBus";
+
+        private readonly TelemetryClient telemetryClient;
+
+        // Every fetcher is unique for event payload and particular property into this payload
+        // when we first receive an event and require particular property, we cache fetcher for it
+        // There are just a few (~10) events we can receive and each will a have a few payload fetchers
+        // I.e. this dictionary is quite small, does not grow up after service warm-up and does not require clean up
+        private readonly Dictionary<string, Dictionary<string, PropertyFetcher>> propertyFetchers =
```
pharring: Instead of a "dictionary of dictionaries" consider a single Dictionary keyed on a ValueTuple of the two strings.
(If you can't use ValueTuple, consider writing a struct with two strings and a custom GetHashCode - don't just concat the two strings because the idea is to avoid heap allocations for lookup).

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
```diff
@@ -0,0 +1,171 @@
...
+                    telemetry.Context.Properties[item.Key] = item.Value;
+                }
+            }
+
+            PropertyFetcher statusFetcher = this.GetOrCreatePropertyFetcher(eventName, StatusPropertyName);
+            telemetry.Success = (TaskStatus)statusFetcher.Fetch(eventPayload) == TaskStatus.RanToCompletion;
+        }
+
+        private PropertyFetcher GetOrCreatePropertyFetcher(string eventName, string propertyName)
+        {
```
pharring: Do the dictionary look-ups need to be protected against multi-thread access? (Should it be a `ConcurrentDictionary`?)

Src/DependencyCollector/Shared/TelemetryDiagnosticSourceListener.cs
```diff
@@ -19,13 +19,20 @@ internal class TelemetryDiagnosticSourceListener : DiagnosticSourceListenerBase<
         private readonly Dictionary<string, HashSet<string>> includedDiagnosticSourceActivities 
             = new Dictionary<string, HashSet<string>>(StringComparer.OrdinalIgnoreCase);
 
+        private readonly Dictionary<string, HandleDiagnosticsEvent<HashSet<string>>> customEventHanlers = new Dictionary<string, HandleDiagnosticsEvent<HashSet<string>>>(StringComparer.OrdinalIgnoreCase);
```
pharring: typo: customEventHan**d**lers

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
```diff
@@ -0,0 +1,171 @@
...
+            }
+        }
+
+        private void OnDependency(string name, object payload, Activity activity)
+        {
+            DependencyTelemetry telemetry = new DependencyTelemetry { Type = DependencyType };
+
+            this.SetCommonProperties(name, payload, activity, telemetry);
+
+            PropertyFetcher urlFetcher = this.GetOrCreatePropertyFetcher(name, EndpointPropertyName);
```
pharring: Would the following help?
```C#
T FetchPayloadProperty<T>(object payload, string eventName, string propertyName)
{
    PropertyFetcher fetcher = this.GetOrCreatePropertyFetcher(eventName, propertyName);
    return (T)fetcher.Fetch(payload);
}

File: Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs (hunk 1 above)
Comment (lmolkova) [reply Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs]: good point

File: Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs (hunk 2 above)
Comment (lmolkova) [reply Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs]: right

File: Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs (hunk 3 above)
Comment (lmolkova) [reply Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs]: well, nothing should break even without a concurrent dictionary. in the worst case, a few fetchers will be created. but anyway, is better to use a concurrent dictionary and all sugars it provides

File: Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs [hunk 6]
```diff
@@ -147,25 +142,31 @@ private void SetCommonProperties(string eventName, object eventPayload, Activity
...
+            public Property(string eventName, string propertyName)
             {
-                fetchers = new Dictionary<string, PropertyFetcher>();
-                this.propertyFetchers.Add(eventName, fetchers);
+                this.eventName = eventName;
+                this.PropertyName = propertyName;
             }
 
-            if (!fetchers.TryGetValue(propertyName, out PropertyFetcher fetcher))
+            public override int GetHashCode()
```
pharring: You should also override Equals (the default Equals for structs involves reflection).
Perhaps also implement `IEquatable<Property>`
```C#
public override bool Equals(object other) => (other as Property)?.Equals(this) ?? false;
public bool Equals(Property other) => this.eventName == other.eventName && this.PropertyName == other.PropertyName;

File: Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs [hunk 7]
```diff
@@ -119,10 +112,12 @@ private void SetCommonProperties(string eventName, object eventPayload, Activity
...
             // activity name looks like 'Microsoft.Azure.ServiceBus.<Name>'
             // as namespace is too verbose, we'll just take the last node from the activity name as telemetry name
-            string[] activityNameSegments = activity.OperationName.Split('.');
-            if (activityNameSegments.Length > 0)
+            string activityName = activity.OperationName;
+            int lastDotIndex = activityName.LastIndexOf('.');
+            if (lastDotIndex > 0)
             {
-                telemetry.Name = activityNameSegments[activityNameSegments.Length - 1];
+                int length = activityName.Length - lastDotIndex - 1;
```
pharring: You don't need to compute length yourself if you use the single-arg overload of Substring (you just give it the start position and it goes to the end of the string)

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
```diff
@@ -119,10 +112,12 @@ private void SetCommonProperties(string eventName, object eventPayload, Activity
         {
             // activity name looks like 'Microsoft.Azure.ServiceBus.<Name>'
             // as namespace is too verbose, we'll just take the last node from the activity name as telemetry name
-            string[] activityNameSegments = activity.OperationName.Split('.');
-            if (activityNameSegments.Length > 0)
+            string activityName = activity.OperationName;
+            int lastDotIndex = activityName.LastIndexOf('.');
+            if (lastDotIndex > 0)
```
pharring: If lastDotIndex is 0 (activityName has a leading period) or negative (activityName doesn't have a period), does `telemetry.Name` need to be set?

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
```diff
@@ -0,0 +1,172 @@
...
+        {
+            DependencyTelemetry telemetry = new DependencyTelemetry { Type = DependencyType };
+
+            this.SetCommonProperties(name, payload, activity, telemetry);
+
+            // Endpoint is URL of particular ServiceBus, e.g. sb://myservicebus.servicebus.windows.net/
+            telemetry.Data = this.FetchPayloadProperty<Uri>(name, EndpointPropertyName, payload)?.ToString();
+
+            // Queue/Topic name, e.g. myqueue/mytopic
+            telemetry.Target = this.FetchPayloadProperty<string>(name, EntityPropertyName, payload);
```
TomMilos: According to [Application Insights data model](https://docs.microsoft.com/en-us/azure/application-insights/application-insights-data-model-dependency-telemetry#target) should identify the side of dependency call. It is also used to identify nodes in AppMap. Entity path can be the same for different SB accounts so it does not serve that purpose. Endpoint hostname works better here.

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
lmolkova: this is invalid operation name, i.e. it should never ever happen.  I'd say there is no preference for this case. I'll put >= 0 for consistency (as 0 index does not lead to any exception or issue)

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
TomMilos: alternatively we could combine endpoint and entity but that would result in multiple nodes for the same SB account on AppMap - it might not scale properly in some cases (i.e. entity can have high cardinality for some customers)

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
```diff
@@ -0,0 +1,172 @@
...
+            this.SetCommonProperties(name, payload, activity, telemetry);
+
+            // Endpoint is URL of particular ServiceBus, e.g. sb://myservicebus.servicebus.windows.net/
+            telemetry.Url = this.FetchPayloadProperty<Uri>(name, EndpointPropertyName, payload);
+
+            // Entity 
+            // We want to make Source field extendable at the beginning as we may add
+            // multi ikey support and also build some special UI for requests coming from the queues
+            // so the Source looks like roleName:queueName (using the multi ikey schema).
+            telemetry.Source = "roleName:" + this.FetchPayloadProperty<string>(name, EntityPropertyName, payload);
```
TomMilos: like with ```Dependency.Target``` - this is not unique enough (i.e. what if 2 SB accounts have a queue with the same name?)

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        private void OnDependency(string name, object payload, Activity activity)
+        {
+            DependencyTelemetry telemetry = new DependencyTelemetry { Type = DependencyType };
+
+            this.SetCommonProperties(name, payload, activity, telemetry);
+
+            // Endpoint is URL of particular ServiceBus, e.g. sb://myservicebus.servicebus.windows.net/
+            telemetry.Data = this.FetchPayloadProperty<Uri>(name, EndpointPropertyName, payload)?.ToString();
```
TomMilos: ```Endpoint``` has a fixed value fro the whole SB client (is stripped out of connection string) so it does not fir into Data field which should have data specific to particular operation. ```PartitionKey``` would fit better here IMO

Src/DependencyCollector/Shared/TelemetryDiagnosticSourceListener.cs
```diff
@@ -209,6 +209,17 @@ protected override HashSet<string> GetListenerContext(DiagnosticListener diagnos
             return includedActivities;
         }
 
+        protected override HandleDiagnosticsEvent<HashSet<string>> GetEventHandler(DiagnosticListener diagnosticListener)
+        {
+            HandleDiagnosticsEvent<HashSet<string>> eventHandler;
+            if (this.customEventHandlers.TryGetValue(diagnosticListener.Name, out eventHandler))
+            {
+                return eventHandler;
```
TomMilos: So this skips the ```IsActivityIncluded()``` check which allows the user to control which individual activities to track

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        private void OnRequest(string name, object payload, Activity activity)
+        {
+            RequestTelemetry telemetry = new RequestTelemetry();
+
+            this.SetCommonProperties(name, payload, activity, telemetry);
+
+            // Endpoint is URL of particular ServiceBus, e.g. sb://myservicebus.servicebus.windows.net/
+            telemetry.Url = this.FetchPayloadProperty<Uri>(name, EndpointPropertyName, payload);
```
TomMilos: So request URL traditionally identifies something on the processing side (server) - here the Endpoint identifies the triggering side (client). I know Endpoint is the only URL-like data that we have but I think it might be confusing. We might just leave it blank

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
```diff
@@ -0,0 +1,172 @@
...
+        public const string DiagnosticSourceName = "Microsoft.Azure.ServiceBus";
+        private const string StatusPropertyName = "Status";
+        private const string EntityPropertyName = "Entity";
+        private const string EndpointPropertyName = "Endpoint";
+        private const string ExceptionPropertyName = "Exception";
+
+        // We want to reflect in the dependency type that it's one of the queue operations.
+        // This way, UI could have special representation for all kinds of queues
+        // Not knowing particular service
+        private const string DependencyType = "Queue.ServiceBus";
```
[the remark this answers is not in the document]
lmolkova: how would you handle (on the Portal side) Azure Storage Queue, Event Hub, Kafka? Presumably, we want to have a special UI for the queues: e.g. show a node for the queue itself.

Would we have to maintain a list of the all possible dependency types that represent a queue? Could we allow custom queue instrumentations and clients then?

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
```diff
@@ -0,0 +1,172 @@
...
+            Activity currentActivity = Activity.Current;
+
+            switch (evnt.Key)
+            {
+                case "Microsoft.Azure.ServiceBus.ProcessSession.Stop":
+                case "Microsoft.Azure.ServiceBus.Process.Stop":
+                    this.OnRequest(evnt.Key, evnt.Value, currentActivity);
+                    break;
+                case "Microsoft.Azure.ServiceBus.Exception":
+                    this.OnException(evnt.Key, evnt.Value);
```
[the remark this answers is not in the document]
lmolkova: why it will be collected multiple times?

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
```diff
@@ -0,0 +1,172 @@
...
+    using System.Diagnostics;
+    using System.Threading.Tasks;
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+
+    /// <summary>
+    /// Implements ServiceBus DiagnosticSource events handling.
+    /// </summary>
+    internal class ServiceBusDiagnosticsEventHandler
```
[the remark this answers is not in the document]
lmolkova: sure

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
lmolkova: how related it is to partition?
The data model now is too tight to HTTP, that's why another protocol does not really fit.

I'm fine with moving it to `Endpoint` custom dimension though or removing it all. It contains the servicebus name though, so perhaps custom dimension is good enough

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
lmolkova: Right, and Target being queueName fits perfectly fine into the side of dependency call. Combining it with endpoint makes node name not readable (too long to show). So we need more rules here: e.g. showing queue name only, but distinguishing nodes based on Target+Endpoint. That's where having Endpoint in Data field is more efficient than in custom dimensions

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
lmolkova: Whatever we set in the dependency.Data we should set in request.Url.
As I shown above it could be useful to distinguish different SBs based on an endpoint and in this case, it's better to have it in the properties rather than in custom dimensions.

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
lmolkova: See above.

Src/DependencyCollector/Shared/TelemetryDiagnosticSourceListener.cs
lmolkova: not the user, but another handler. I'm fine with adding one more check for IsEventEnabled here  and removing it from handler completely. BTW, it will remove the need to TContext in handler

lmolkova: This check is done in IsEventEnabled, it's also done within a handler.
The point is handler is not requested for the source that is disabled

IsActivityEnabled is also called in this IsEventEnabled, i.e. users may filter out (as tests demonstate)

The only problem is multiple listeners when event may be fired even if your listener returned false in IsEnabled. That's a problem in ServiceBus listener indeed. Will fix

lmolkova: I have separated activity filtering (based on the config) from the particular event handler.
I.e. http handler may have more sophisticated logic to filter events than the EventHub listener.

This was filtering by activity name is part of the TelemetryDiagnosticSourceListener and done before any handler. It allows to remove this logic from the handler, which then could only parse specific event.

Let me know what do you think.

I also expect that in future there will be a need to process somehow start events (e.g. Inject app in when sending data and supporting multi ikey scenarios) - it's not possible now and I don't want to ix it until strictly necessary.

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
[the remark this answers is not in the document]
TomMilos: well we already use "Azure queue" and I think the type names should be user friendly. And we cannot map display name in UX because there would be a disconnect form data in Kusto/Analytics.

Not sure what is the right answer to identifying queue-like services, but we need to figure this out for Azure queues anyway

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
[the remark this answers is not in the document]
TomMilos: We already collect unhandled exceptions automatically.

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
TomMilos: I'm not sure if that's the right interpretation of Target and the documentation leaves some wiggle room. @SergeyKanzhelev thoughts?

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
TomMilos: fair enough - Data and Url should match, still not sure what is the value of putting Endpoint here as it is the same for a single SB
the picture that you're painting is that Entity (queue name) is a higher level thing than the actual SB (Endpoint)

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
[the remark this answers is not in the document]
lmolkova: my point is: all queues should be shown differently than RPS/HTTP calls on the AppMap.
So we need to be able to efficiently tell from the data that it's a queue.
Only after that we may need to know if that's a SB or Storage Queue or EventHub or Kafka, etc...

And if Kafka is instrumented outside of our control, they could ensure that type is queue and it magically works on the portal. Otherwise, we would need to make sure portal knows about this implementation and would show it as a queue and not as standard RPC call.

So the question is: how do we ensure this now. We should not use customDimensions as they are inefficient to query and we should use default properties.

Agreed about user friendliness on the portal. And I understand your concern about consistency between what AppMap shows on the node. though I'm not sure there is an expectation that it maps directly to kusto data.

Anyway do you know if there was a discussion on this problem and perhaps there is some poposal?:

> my point is: all queues should be shown differently than RPS/HTTP calls on the AppMap.
So we need to be able to efficiently tell from the data that it's a queue.

I.e. what should we do to ensure it?

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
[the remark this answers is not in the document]
lmolkova: what if it is handled?

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
lmolkova: it's the most granular and verbose thing we have: it has the most details. You may have 10 queues in one servicebus. Or in case of Azure Storage Queues, queuename is all you have (endpoint is always the same and different for emulators/private/govenrments/etc clouds only).
 We are trying here not only for service bus, we are trying to make sure the same work for other queues as well.

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
lmolkova: We have many constraints here: 
1. we need to collect useful data that and be able to efficiently query it and build UI based on it
2. we have rules around the data model

Data model for dependencies so far has been used for http and sql only. Some protocols may not fit into it as we see it here.
So if we can't satisfy both requirements, I prefer we update the rules, or change schema rather than sacrifice p1 (data and efficiency)

SergeyKanzhelev:
@lmolkova can you give some context here. Are those DS can be hooked in generic DS listener? Or they were implemented before it?

lmolkova:
@SergeyKanzhelev 

ServiceBus was recently instrumented with DiagnosticSource (by me).
It can be hooked with generic listener, however, some customization is needed to make sure all useful data is properly collected:
1. We also collect Requests here (message processing). Generic listener is capable of collecting dependencies only
2. generic listener does not yet have a guidance on what tags to set on telemetry and how exactly map it to AI properties. So with @TomMilos, we agreed that at this point we can only instrument a lib with a minimum required to collect data without depending on this lib.

This change adds this customization for SB only and also adds extensibility point to the listener to allow such customizations while reusing other infra

SergeyKanzhelev:
LGTM overall. I'll do in depth review later, do not block on me please

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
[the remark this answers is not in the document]
TomMilos: I don't know about such discussion, I will start one

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
TomMilos: Data is a generic purpose field - it can contain URL, SQL query, ID of the data being accessed. So it has data that is vague and not well defined - because of that it will be hard to use it to increase perf of checking endpoint as that will be only limited to a subset of all data (and checking that costs). If we don't see anything useful to put there though let's just leave it empty

I think Endpoint is the most important information about the dependency as it allows to identify where to look next. That's why I believe it should be used as "Target"

lmolkova: so if i have 10 queues, what do I do?
Where would you put queue name?

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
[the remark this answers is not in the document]
zakimaksyutov: Re: visualizing queues differently than HTTP calls. I think on top of it we have at least two types of queues - standard queue (Azure Queue) and multiplexer queue (EventHub). I would expect different visualization for these queues. So, if we're solving the problem how to differentiate dependencies then maybe we should think about different queue types as well.

Src/DependencyCollector/Shared/ServiceBusDiagnosticsListener.cs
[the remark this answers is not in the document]
lmolkova: after reading https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/491#issuecomment-325829913 and the whole discussion I agree, we should not track exceptions here

Src/DependencyCollector/Shared/ServiceBusDiagnosticsEventHandler.cs
```diff
@@ -0,0 +1,171 @@
...
+        private readonly ConcurrentDictionary<Property, PropertyFetcher> propertyFetchers = new ConcurrentDictionary<Property, PropertyFetcher>();
+
+        internal ServiceBusDiagnosticsEventHandler(TelemetryConfiguration configuration)
+        {
+            this.telemetryClient = new TelemetryClient(configuration);
+        }
+
+        public bool IsEventEnabled(string evnt, object arg1, object arg2)
+        {
+            return !evnt.EndsWith(TelemetryDiagnosticSourceListener.ActivityStartNameSuffix, StringComparison.OrdinalIgnoreCase);
```
pharring: Is `IgnoreCase` intentional? Elsewhere (e.g. `OnEvent`) the case is significant.

Src/DependencyCollector/Shared/TelemetryDiagnosticSourceListener.cs
TomMilos: Looks good - thanks!

Src/DependencyCollector/Shared/ServiceBusDiagnosticsEventHandler.cs
```diff
@@ -0,0 +1,171 @@
...
+
+        private void SetCommonProperties(string eventName, object eventPayload, Activity activity, OperationTelemetry telemetry)
+        {
+            // activity name looks like 'Microsoft.Azure.ServiceBus.<Name>'
+            // as namespace is too verbose, we'll just take the last node from the activity name as telemetry name
+            string activityName = activity.OperationName;
+            int lastDotIndex = activityName.LastIndexOf('.');
+            if (lastDotIndex >= 0)
+            {
+                telemetry.Name = activityName.Substring(lastDotIndex + 1);
```
TomMilos: So I'm not sure if this *Name* will work equally well for __Dependencies__ and __Requests__ - Dependencies have the *Type* property to combine it with but Requests don't and the *Source* field does not seem to be used in traditional UX. So this way "Process" Requests from SB and EH will be jammed together under same UX entry. Maybe we should use sth like ```{OperationName} from {DependencyTypeName}``` just with the proper processing - so in this case it would be ```Process from Azure ServiceBus```. Thoughts?

TomMilos:
@lmolkova Currently the ServiceBus Diagnostic Source is disabled by default - should we enable it?

lmolkova:
@TomMilos 
> Currently the ServiceBus Diagnostic Source is disabled by default - should we enable it?

Yes. I wonder, do we have any special infra to enable it with AspNetCore SDK?

Src/DependencyCollector/Shared/ServiceBusDiagnosticsEventHandler.cs
lmolkova: If we continue HTTP analogy, Name describes king of operation that happens on this node (i.e. GET controller).
Here we have a verb only (Process) and no route (controller).

While Source describes where to operation came from. I believe type belongs to the Source (until it has a special field).

Also, this part never ever changes. What is the benefit of including it into the telemetry?

Src/DependencyCollector/Shared/ServiceBusDiagnosticsEventHandler.cs
```diff
@@ -0,0 +1,171 @@
...
+            
+            // Endpoint is URL of particular ServiceBus, e.g. sb://myservicebus.servicebus.windows.net/
+            string endpoint = this.FetchPayloadProperty<Uri>(name, EndpointPropertyName, payload)?.ToString();
+
+            // Queue/Topic name, e.g. myqueue/mytopic
+            string queueName = this.FetchPayloadProperty<string>(name, EntityPropertyName, payload);
+
+            // Target uniquely identifies the resource, we use both: queueName and endpoint 
+            // with schema used for SQL-dependencies
+            telemetry.Target = string.Join(" | ", queueName, endpoint);
```
cijothomas: should it be endpoint followed by queuename to be consistent with sql. (sqlserverinstancename | dbname)

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
```diff
@@ -151,6 +151,7 @@ public void Initialize(TelemetryConfiguration configuration)
                             if (this.IncludeDiagnosticSourceActivities != null && this.IncludeDiagnosticSourceActivities.Count > 0)
                             {
                                 this.telemetryDiagnosticSourceListener = new TelemetryDiagnosticSourceListener(configuration, this.IncludeDiagnosticSourceActivities);
+                                this.telemetryDiagnosticSourceListener.RegisterHandler(ServiceBusDiagnosticsEventHandler.DiagnosticSourceName, new ServiceBusDiagnosticsEventHandler(configuration));
```
cijothomas: registering handler for ServiceBus is required _only if_  IncludeDiagnosticSourceActivities contains ServiceBus right?

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Changelog
 
 ## Version 2.5.0-beta2
-
+- Improved ServiceBus dependency and requests collection for (Microsoft.Azure.ServiceBus 2.1.0)
```
cijothomas: Can you open a GH issue describing the enhancement (use content which is already given in the PR itself), and link to it here? 
The description provided is part of PR, and its not easy to find the PR which implement a given feature.

cijothomas: Also, mention that 'Request' telemetry is not a typical request., and how to filter them out in UI. (using Source field..)

SergeyKanzhelev: Why are you saying that this `Request` is not real?

cijothomas:
@lmolkova 
Once everyone agree TelemetryDiagnosticSourceListener should be enabled by default for certain sources, then it should be done for asp.net core sdk as well. (as of now this is NOT enabled in default for Web SDK and .CORE SDK)
https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/c7e34ea2484f90a542e000c7a98656bf45ccf436/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs#L143

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
lmolkova: nope, handler is registered anyway, however if there ServiceBus is disabled, handler will never ever be called - filtering of activities is done by `TelemetryDiagnosticSourceListener`

CHANGELOG.md
lmolkova: @cijothomas I expect typical case would be that node with roleName 'worker-a' only reports requests from the service-bus, i.e. filtering them could be done by role name.
We don't really support filtering different requests types (as we don't have type property on the request).

I can describe how they can be filtered/grouped via Sourcefield. Where would you like me to write that?

Src/DependencyCollector/Shared/ServiceBusDiagnosticsEventHandler.cs
lmolkova: agree

CHANGELOG.md
cijothomas: Ok, if processing of SB is done in worker roles. I was visualizing a scenario where this was done in WebRole itself, which has its own regular requests. So overview blades would show regular incoming 'Requests' and 'SB Requests'.
image

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
cijothomas: ok

CHANGELOG.md
cijothomas: @SergeyKanzhelev The Request telemetry collected/created by this PR is not 'real' in the sense its not arising out of an Incoming User Request, and it does not have typical HTTP Response Codes. 
@lmolkova please correct me if this understanding is wrong.

cijothomas: I can describe how they can be filtered/grouped via Sourcefield. Where would you like me to write that? -- in the GH issue to be opened describing this issue.
(The Changelog.md should link to the GH Issue , so its easier to find)

cijothomas:
@lmolkova final ask - If this PR is changing the 'default AI.Config' then please update the following test apps AI.Config file to also include the change to make sure the test app is as close to default :)
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Test/E2ETests/TestApps/Net452/E2ETestApp/ApplicationInsights.config
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Test/E2ETests/TestApps/Net452/E2ETestWebApi/ApplicationInsights.config

CHANGELOG.md
lmolkova: @cijothomas for now we are not building UI to show SB requests, UI will remain the same as for HTTP requests. We might do this in future.
I think it's perfectly fine to consider them as a normal RequestsTelemetry (that represent server-side telemetry that describes operation processing). As you mentioned the only difference is that 'status code' makes no sense, but there are other instrumented protocols like that, e.g. ServiceFabric remoting

I'll update GH issue (#740) with details about source and reference it in the changelog

TomMilos: @cijothomas - we've been manually collecting requests like that for workers (CDS, Iris) for at least 2 years now and the UX plays with it well even if some things are missing. This ServiceBus telemetry includes enough information to make the current UX light up nicely (i.e. name, timestamp, duration, success)

cijothomas: @TomMilos yes i understand that. I was referring to users who might be doing SB read/process in a WebRoles where there are other web requests as well. They will suddenly start seeing additional requests (SB process) once upgraded to beta2, and this may be unexpected for many. Thats why i asked to have it explicitly called out in GH issue/changelogs.
(Esp. for asp.net core users, where this feature is just automatically enabled and somewhat invisible as well due to lack of AI.Config )

cijothomas:
@SergeyKanzhelev , @Dmitry-Matveev  explicitly to approve as we ship generic Diag listener by default with this.

CHANGELOG.md
lmolkova: @cijothomas I have described data structure in the #740 (and updated changelog to reference it), so customers have information on how to distinguish SB requests from the other. The assumption though is that in the majority of cases you have a web role and receive HTTP requests or you have a worker role and receive work items from the queue.

cijothomas: Thanks!

lmolkova:
@cijothomas I think I addressed everything. Let me know if I missed something. I also enabled ServiceBus source in asp.net core SDK (https://github.com/Microsoft/ApplicationInsights-aspnetcore/pull/550).

cijothomas:
Merging.

Issue #753, opened 2017-11-23
Track dependencies from EventHubs


TomMilos:
The upcoming version of EventHubs .NET SDK was instrumented with DiagnosticSource and Activities. This instrumentation allows to collect dependency telemetry for ```Send``` and ```Receive``` operations which otherwise is not possible as EH uses AMQP protocol which is not HTTP-based. 

The EH SDK instrumentation will be picked up by the generic ```TelemetryDiagnosticSourceListener``` however some customizations are needed in order to make the telemetry user and UX friendly

MS-TimothyMothra:
@TomMilos is there any more work for this item?
Would it be appropriate to close this issue and track the remaining work in #758 ?

MS-TimothyMothra:
@TomMilos any update?

PR #754, opened 2017-11-23
[DS] Add EventHubs diagnostic event handler

TomMilos:
Fix issue #753.

Short description of the fix:
The upcoming version of EventHubs .NET SDK was instrumented with DiagnosticSource and Activities. This instrumentation allows to collect dependency telemetry for Send and Receive operations which otherwise is not possible as EH uses AMQP protocol which is not HTTP-based.

The EH SDK instrumentation will be picked up by the generic TelemetryDiagnosticSourceListener however some customizations are needed in order to make the telemetry user and UX friendly

- [X] I ran all unit tests locally
- [X] CHANGELOG.md __updated__ 

- [ ] I ran [functional tests](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/CONTRIBUTING.md#functional-tests) locally.
- [X] The PR will trigger build, unit test, and functional tests automatically. If your PR submitted from fork - mention one of committers to initiate the build for you.


Src/DependencyCollector/Shared/Implementation/TelemetryDiagnosticSourceListener.cs
```diff
@@ -1,256 +0,0 @@
...
-    using System;
-    using System.Collections.Generic;
-    using System.Diagnostics;
-    using Microsoft.ApplicationInsights.Channel;
-    using Microsoft.ApplicationInsights.DataContracts;
-    using Microsoft.ApplicationInsights.Extensibility;
-    using Microsoft.ApplicationInsights.Extensibility.Implementation;
-    using Microsoft.ApplicationInsights.Web.Implementation;
-
-    internal class TelemetryDiagnosticSourceListener : DiagnosticSourceListenerBase<HashSet<string>>
```
TomMilos: This was a duplicate file of ```Src/DependencyCollector/Shared/TelemetryDiagnosticSourceListener.cs``` - removed to avoid confusion

Src/DependencyCollector/Shared/ServiceBusDiagnosticsEventHandler.cs
```diff
@@ -1,171 +0,0 @@
...
-    using System.Globalization;
-    using System.Threading.Tasks;
-    using Microsoft.ApplicationInsights.DataContracts;
-    using Microsoft.ApplicationInsights.Extensibility;
-    using Microsoft.ApplicationInsights.Extensibility.Implementation;
-
-    /// <summary>
-    /// Implements ServiceBus DiagnosticSource events handling.
-    /// </summary>
-    internal class ServiceBusDiagnosticsEventHandler : IDiagnosticEventHandler
```
TomMilos: This was moved to ```Src/DependencyCollector/Shared/Implementation/EventHandlers/``` and reusable logic was pulled up to ```DiagnosticsEventHandlerBase``` class

Src/DependencyCollector/Shared.Tests/Implementation/EventHubsDiagnosticListenerTests.cs
```diff
@@ -0,0 +1,168 @@
...
+                Assert.AreEqual("v1", telemetry.Properties["k1"]);
+                Assert.AreEqual("eventhubname.servicebus.windows.net", telemetry.Properties["peer.hostname"]);
+                Assert.AreEqual("ehname", telemetry.Properties["eh.event_hub_name"]);
+                Assert.AreEqual("SomePartitionKeyHere", telemetry.Properties["eh.partition_key"]);
+                Assert.AreEqual("EventHubClient1(ehname)", telemetry.Properties["eh.client_id"]);
+            }
+        }
+
+        [TestMethod]
+        public void EventHubsSendExceptionsAreIgnored()
```
xiaomi7732: Could you please kindly explain why is exception ignored? Thanks!

TomMilos: Exceptions are usually well covered by user code or would be tracked in other context. The concern is that collecting a lot of transient exceptions would be noisy and incur telemetry cost

xiaomi7732: Fair enough. I see that the implementation for S.B. shares the same logic. I totally have no problem with it. Thanks for letting me know the reason behind.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - [Track requests and dependencies from ServiceBus .NET Client 2.1.0](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/740)
 - [Fix for REST API Request filter bug](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/175)
 - [Fix for SyntheticUserAgentTelemetryInitializer null check](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/750)
+- [Track dependencies from EventHubs .NET Client 2.0.0](https://github.com/Azure/azure-event-hubs-dotnet/pull/203)
```
cijothomas: Please link to the GH issue (https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/753).
Inside the issue, you can link to the eventHub PR .

cijothomas: make it explicit that 2.0.0 version of EventHub/Sb client to avoid potential confusion with .NET itself

Src/DependencyCollector/Shared/Implementation/EventHandlers/EventHubsDiagnosticsEventHandler.cs
```diff
@@ -0,0 +1,52 @@
...
+        private const string EntityPropertyName = "Entity";
+        private const string EndpointPropertyName = "Endpoint";
+
+        internal EventHubsDiagnosticsEventHandler(TelemetryConfiguration configuration) : base(configuration)
+        {
+        }
+
+        public override void OnEvent(KeyValuePair<string, object> evnt, DiagnosticListener ignored)
+        {
+            Activity currentActivity = Activity.Current;
```
cijothomas: minor - move this inside if stmt.

Src/DependencyCollector/Shared.Tests/Implementation/EventHubsDiagnosticListenerTests.cs
```diff
@@ -0,0 +1,168 @@
...
+
+                DiagnosticListener listener = new DiagnosticListener("Microsoft.Azure.EventHubs");
+
+                Activity parentActivity = new Activity("parent").AddBaggage("k1", "v1").Start();
+                var telemetry  = this.TrackOperation<DependencyTelemetry>(listener, "Microsoft.Azure.EventHubs.Send", TaskStatus.RanToCompletion);
+
+                Assert.IsNotNull(telemetry);
+                Assert.AreEqual("Send", telemetry.Name);
+                Assert.AreEqual(RemoteDependencyConstants.AzureEventHubs, telemetry.Type);
+                Assert.AreEqual("sb://eventhubname.servicebus.windows.net/ | ehname", telemetry.Target);
```
cijothomas: Can you describe this in the original GH issue opened? 
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/753

Specifically - mention which properties goes to which field of DependencyTelemetry. Like target will be ehurl | ehname etc.
(Since our documentations are generally weak, GH issue is only we have to explain to customers about features:) )

TomMilos: Done

CHANGELOG.md
TomMilos: Done

TomMilos: Done

Src/DependencyCollector/Shared/Implementation/EventHandlers/EventHubsDiagnosticsEventHandler.cs
TomMilos: Done

Src/DependencyCollector/Shared.Tests/Implementation/EventHubsDiagnosticListenerTests.cs
```diff
@@ -0,0 +1,168 @@
...
+                listener.StopActivity(
+                    activity,
+                    new
+                    {
+                        Entity = "ehname",
+                        Endpoint = new Uri("sb://eventhubname.servicebus.windows.net/"),
+                        PartitionKey = "SomePartitionKeyHere",
+                        Status = status
+                    });
+                return this.sentItems.Last() as T;
```
Dmitry-Matveev: Will this miss a case where duplicate telemetry was produced because it only takes the last?
Would it be better to clear `sentItems` between tests and look for `Single()`?
I'm asking because I remember this biting us in the other .NET SDK tests where we haven't noticed the dupes for a while due to the similar implementation.

Src/DependencyCollector/Shared/Implementation/EventHandlers/ServiceBusDiagnosticsEventHandler.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            switch (evnt.Key)
+            {
+                case "Microsoft.Azure.ServiceBus.ProcessSession.Stop":
+                case "Microsoft.Azure.ServiceBus.Process.Stop":
+                    this.OnRequest(evnt.Key, evnt.Value, currentActivity);
+                    break;
+                case "Microsoft.Azure.ServiceBus.Exception":
+                    break;
+                default:
```
Dmitry-Matveev: This seems to be catch-all to generate dependency from every event other than 3 above. Are we sure that any new events that may appear in `ServiceBus` namespace are "dependency-worthy"?

Test/E2ETests/TestApps/Net452/E2ETestApp/ApplicationInsights.config
```diff
@@ -36,6 +36,7 @@
         <Add>e2etests_azureemulator_1</Add>
       </ExcludeComponentCorrelationHttpHeadersOnDomains>
       <IncludeDiagnosticSourceActivities>
+        <Add>Microsoft.Azure.EventHubs</Add>
```
Dmitry-Matveev: Do you plan to add some functional tests for ServiceBus / EventHub? I see that configuration of the tests will now force them to listen to those dependencies but we do not produce such calls in the test web apps currently to validate their collection.

Src/DependencyCollector/Shared.Tests/Implementation/EventHubsDiagnosticListenerTests.cs
TomMilos: This is a test helper method and it can be used multiple times during a single test (isn't yet, but there will be a next iteration of changes for supporting Request tracking).
I get your point though so I've added an assertion to check that only a single item is added to the ```sentItems``` list

Src/DependencyCollector/Shared/Implementation/EventHandlers/ServiceBusDiagnosticsEventHandler.cs
TomMilos: This is a philosophical question :) If new instrumentation is added to EH/SB SDK should we ignore it or collect as something? We think the latter is a better solution as users will start seeing additional telemetry without a need to upgrade AI SDK. If that telemetry is not desired by them they still have an option to ignore it through config (```IncludeDiagnosticSourceActivities``` whitelist)

Test/E2ETests/TestApps/Net452/E2ETestApp/ApplicationInsights.config
TomMilos: If these are expected than we will add them, although I don't think we'll be able to do that for Beta. We can make sure they are in place before Stable though

Src/DependencyCollector/Shared/Implementation/EventHandlers/ServiceBusDiagnosticsEventHandler.cs
TomMilos: That being said I did change the logic for EH listener to explicitly list the dependency activities (*Send* and *Receive*) as we are currently adding new instrumentation there which will produce *Process* activity which will be collected as request. Once that change is approved (https://github.com/Azure/azure-event-hubs-dotnet/pull/214) we will add support for handling it in AI SDK and update this logic to catch-all

Test/E2ETests/TestApps/Net452/E2ETestApp/ApplicationInsights.config
Dmitry-Matveev: We typically add functional tests coverage for the auto-collected dependencies and requests to make sure they are flowing in the actual web application before we ship the feature as well as for the regression detection purposes. I would recommend adding them in the stable timeframe if you already had any other type of the validation for beta 2 other than unit tests (e.g. manual test). Let's discuss more offline if needed.

Src/DependencyCollector/Shared/Implementation/EventHandlers/ServiceBusDiagnosticsEventHandler.cs
Dmitry-Matveev: RE: "We think the latter is a better solution as users will start seeing additional telemetry without a need to upgrade AI SDK."
Agree, it's a bit philosophical; however, I would expect us to limit this philosophy to auto-accept only a certain patterns of the events to make them Dependency Telemetry Items, e.g. Start/Stop pattern or similar. Converting all events from the code we do not own exposes us to the risk of sending telemetry for every light operation someone decided to mark on the other side. That leads to billing implications and will be most likely corrected by the customer only after the first bill. "Converting pattern" obviously has similar risks but it's a bit more contained.

TomMilos: So yes - we are only collecting Activities based on Start/Stop/Exception event naming convention. That being said the inheritance hierarchy does not make it clear what kinds of event names would get here and changes in base classes could affect this behavior so I've added an explicit check for ```.Stop``` suffix in catch-all block for SB (EH has no catch-all at this point)

Test/E2ETests/TestApps/Net452/E2ETestApp/ApplicationInsights.config
TomMilos: Let's sync up on the details next week - currently the corresponding changes in EH & SB SDKs have not been published yet so we'd have to emulate them anyway. I've created an issue for tracking this: #758

PR #768, opened 2017-12-15
Library instrumentation guidance

TomMilos:
Guidance for instrumenting libraries with Diagnostic Source and Activities (WIP)


TomMilos:
@AlexKlimovMS - please take a look

LibraryInstrumentationGuidance.md
```diff
@@ -0,0 +1,107 @@
...
+
+Both Diagnostic Source and Activity have been used to instrument [System.Net.Http][SystemNetHttp] and [Microsoft.AspNetCore.Hosting][MicrosoftAspNetCoreHosting], although that instrumentation is not fully complaiant with this guidance.
+
+More recently two new libraries were instrumented and that work was the basis for this guidance. These libraries are client SDKs for [Azure Event Hubs][MicrosoftEventHubs] and [Azure Service Bus][MicrosoftServiceBus], both of which support very high throuput scenarios.
+
+[This document][DiagnosticSourceActivityHowto] goes into more details on how to efficiently use Diagnostic Source.
+
+## What should be instrumented
+
+The goal of instrumentation is to give users the visibility to how particular operations are being performed inside the library. This information can be later used to diagnose performance or issues. It is up to the library authors to identify operations that the library performs and are worthy of monitoring. Some good examples are external dependency calls, cache utilization (hit/miss/refresh) or awaiting for system events.
```
lmolkova: > Some good examples are external dependency calls, cache utilization (hit/miss/refresh) or awaiting for system events.

This sentence (being correct) gives the wrong impression as this doc targets lib owners. For the libs, we expect some public API operations that cause one or more external dependency calls to be instrumented. I think it should be more clear that it's not a single isolated external dependency call, but the full operation that may consist of several retries, input and output processing.

LibraryInstrumentationGuidance.md
```diff
@@ -0,0 +1,107 @@
...
+| Property name | Description |
+|:--------------|:-------------------|
+| `Endpoint` | The ```Uri``` of an endpoint the activity is for (e.g. target database, service) |
+| `PartitionKey` | The key/ID of the partition the activity is for |
+| `Status` | The ```TaskStatus``` of a completed asynchronous task |
+| `Exception` | The captured exception object |
+
+### Performance
+
+In order to avoid unnecessary overhead it is highly recommended to check if there is any listener for given activity. This can be done by calling
```
lmolkova: Please also mention that any instrumentation should start with DiagnosticListener.IsEnabled() - no args. It checks for ANY listener for this source and is very efficient. IsEnabled(string) is virtual method and does not give zero instrumentation cost.

LibraryInstrumentationGuidance.md
```diff
@@ -0,0 +1,107 @@
...
+
+```C#
+    source.IsEnabled("Example.MyLibrary.MyOperation", input1, input2)
+```
+
+For more detailed performance considerations please refer to [Diagnostic Source][DiagnosticSourceGuide] and [Activity][ActivityGuide] guides.
+
+### Tags
+
+Activities can have additional tracing information in tags. Tags are meant to be easily consumable and are expected to be logged with the activity without any processing.
```
lmolkova: Please mention that Tags should be set (if any) as early as possible - e.g. it does not make sense to set them right after calling StartActivity, they should be set before that (or before StopActivity).

I  think it's also worth mentioning what should be added and performance implications of that:  Tags are intended to be added along the telemetry blindly, so they should be used wisely (just set a few) and are completely optional. In some sense, they substitute event payloads and give tracing system essential information about the operation (while payloads give full context).

LibraryInstrumentationGuidance.md
```diff
@@ -0,0 +1,107 @@
+# Guidence for instrumenting libraries with Diagnostic Source
```
pharring: Spelling: Guidance

LibraryInstrumentationGuidance.md
```diff
@@ -0,0 +1,107 @@
...
+
+This document provides guidence for adding Diagnostic Source instrumentation to external libraries, which will allow Application Insights and other monitoring solutions to collect meaningful and rich telemetry.
+
+## Diagnostic Source and Activities
+
+[Diagnostic Source][DiagnosticSourceGuide] is a simple module that allows code to be instrumented for production-time logging of rich data payloads for consumption within the process that was instrumented. At runtime consumers can dynamically discover data sources and subscribe to the ones of interest.
+
+[Activity][ActivityGuide] is a class that allows storing and accessing diagnostics context and consuming it with logging system.
+
+Both Diagnostic Source and Activity have been used to instrument [System.Net.Http][SystemNetHttp] and [Microsoft.AspNetCore.Hosting][MicrosoftAspNetCoreHosting], although that instrumentation is not fully complaiant with this guidance.
```
pharring: typo: compliant

LibraryInstrumentationGuidance.md
```diff
@@ -0,0 +1,107 @@
...
+
+More recently two new libraries were instrumented and that work was the basis for this guidance. These libraries are client SDKs for [Azure Event Hubs][MicrosoftEventHubs] and [Azure Service Bus][MicrosoftServiceBus], both of which support very high throuput scenarios.
```
pharring: typo: throughput

LibraryInstrumentationGuidance.md
AlexKlimovMS: Referring SystemNetHttp and AspNetCore.Hosting is good as it builds trust with this guidance. And then the next sentence confuses people. Are they using it incorrectly? Are they ignoring the part of guidance which might hurt performance?

The point I'm making - do we need to mention this at all? How this information should be used by readers of this guidance?

LibraryInstrumentationGuidance.md
TomMilos: I've changed this description although I am not sure if it conveys exactly what you mean. Please take a look

LibraryInstrumentationGuidance.md
TomMilos: Good point - added that remark and updated code samples

LibraryInstrumentationGuidance.md
TomMilos: Thanks, updated this paragraph, please review

LibraryInstrumentationGuidance.md
TomMilos: thanks, fixed!

LibraryInstrumentationGuidance.md
TomMilos: thanks, fixed!

LibraryInstrumentationGuidance.md
TomMilos: thanks, fixed!

LibraryInstrumentationGuidance.md
TomMilos: Good point, I've removed that remark

TomMilos:
@lmolkova 

>I think this guide should become part of DiagnosticSource guide or just published somewhere outside of ApplicationInsights repos. There could be other consumers (Glimpse, Visual Studio, Geneva, 3rd party systems). Also it's beneficial for everyone including AppInsights to keep instrumentation and AppInsights decoupled

I agree, we are explicitly trying to not mention AppInsights in the guide for that reason, so keeping it here unnecessarily ties them together. We can just have a shorter doc here pointing to that generic one.

Do you think DiagnosticSource repo would be better or somewhere in MicrosoftDocs?

I think we should continue reviewing here in the meantime

LibraryInstrumentationGuidance.md
```diff
@@ -0,0 +1,115 @@
...
+
+    Activity activity = null;
+    // create activity only when requested
+    if (source.IsEnabled() && source.IsEnabled("Example.MyLibrary.MyOperation"))
+    {
+        activity = new Activity("Example.MyLibrary.MyOperation");
+        source.StartActivity(activity, new { Input = input });
+    }
+
+    object outout = null;
```
lmolkova: typo: outPut

LibraryInstrumentationGuidance.md
```diff
@@ -0,0 +1,115 @@
...
+Tags are not propagated to child activities.
+
+```csharp
+    activity.AddTag("size", "small");
+    activity.AddTag("color", "blue");
+```
+
+#### Tag naming
+
+A single application can include numerous libraries instrumented with Diagnostic Source. In order to maintain a certain level of consistency of the whole diagnostic data it is recommended to use a common tag naming convention. [OpenTracing][OpenTracingNamingConvention] published naming convention for such tags and can be used as a reference. If no suitable tag was defined there then a new name can be used, ideally following the same convenction.
```
lmolkova: convenction -> convention

lmolkova:
> Do you think DiagnosticSource repo would be better or somewhere in MicrosoftDocs?

MsDocs is a great option. It would be easier to discover the guide than md files.

LibraryInstrumentationGuidance.md
```diff
@@ -1,39 +1,43 @@
...
     // create activity only when requested
-    if (source.IsEnabled("Example.MyLibrary.MyOperation"))
-        activity = source.StartActivity("Example.MyLibrary.MyOperation", new { Input = input });
+    if (source.IsEnabled() && source.IsEnabled("Example.MyLibrary.MyOperation"))
+    {
+        activity = new Activity("Example.MyLibrary.MyOperation");
+        source.StartActivity(activity, new { Input = input });
+    }
 
+    object outout = null;
```
AlexKlimovMS: Typo: output

LibraryInstrumentationGuidance.md
```diff
@@ -0,0 +1,115 @@
...
+In the simplest case the operation that is being monitored needs to be wrapped by an activity. However in order to minimize performance impact the activity should be only created if there is any listener waiting for it
+
+```csharp
+    static DiagnosticListener source = new DiagnosticListener("Example.MyLibrary");
+
+    Activity activity = null;
+    // create activity only when requested
+    if (source.IsEnabled() && source.IsEnabled("Example.MyLibrary.MyOperation"))
+    {
+        activity = new Activity("Example.MyLibrary.MyOperation");
```
SergeyKanzhelev: not to overcomplicate this section. Perhaps create a separate one. We have a promise to distributed trace to propagate all incoming headers/context to all outgoing calls. So even if `IsEnabled` returned `false` you may want to create lightweight `Activity` and still send headers. 

We  discussed it with @lmolkova. @lmolkova have you thought about it more?

lmolkova: @SergeyKanzhelev we are far from reading any headers except Request-Id. We don't know yet how Activity would look like and what headers to propagate blindly or modify. I suggest we update guidance as soon as we know the details and we'd refer to the propagation standard from here. 

@TomMilos BTW, I just noticed that this guide does not cover injection/extraction at all. I think it worth mentioning.

SergeyKanzhelev: Even with `Request-ID` we may need a feature to propagate ID and context even if telemetry was sampled out. Alternatively we need to say that `IsEnabled` means not participating in distributed tracing at all, not that this Activity was sampled out

lmolkova: now IsEnabled serves 2 purposes:
1. preventing unwanted requests to be instrumented (e.g. http calls to AI that report telemetry)
we probably do not want anything to be propagated, but let's not overcomplicate and let's say we don't care if context will be injected into such requests as well.

2. control and sampling. I think we have a lot of open questions here
- what if there is no listener at all? I.e. tracing system. Do we expect .NET and libs to inject headers still? Does 'Trace-Context' give any guidance here?
- assuming there is a listener and it samples at least some activities. Perhaps we want to propagate context even for sampled-out activities. And assuming we still want to save some performance for sampled-out activity, the context should be propagated 'blindly' without modification. Modification (child activity) creation is not possible: imagine if Id was not hierarchical and a new child was created for sampled-out Activity. Such Activity would not be reported (as sampled out), so if some downstream service eventually would report something, if may not be possible to recover the flow - all the intermediate parent/child relationships would be lost because telemetry was not reported. 

What do you think?

LibraryInstrumentationGuidance.md
TomMilos: I remember fixing it :/ Fixed now

LibraryInstrumentationGuidance.md
TomMilos: fixed

LibraryInstrumentationGuidance.md
TomMilos: fixed

LibraryInstrumentationGuidance.md
TomMilos: I agree with @lmolkova's suggestion to leave this out of this guide until the details are figured out. This is intended to be an entry-level guide and distributed tracing is a more advanced subject so we can briefly describe it here and point to a separate guide focused on that (once it is ready). 

Regarding:
> @TomMilos BTW, I just noticed that this guide does not cover injection/extraction at all. I think it worth mentioning.

This also is an advanced use case which might not fit well into this document. I'll try to describe it in a simple and generic way and if we are happy with it then we can leave it here. Otherwise we could direct to your Service Bus guide as a reference point for such scenario

SergeyKanzhelev: It is important detail. Yes, we saying that identifiers/headers should be propagated even when telemetry item was sampled out. It is crucial for multi-tenant environments with separate sampling control.

So there should be a note on the meaning of `isEnabled` and whether we will have lightweight activity population for the case when Activity is sampled, but headers still have to be propagated.

LibraryInstrumentationGuidance.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Performance
+
+In order to avoid unnecessary overhead, it is highly recommended to check if there is any listener for given activity: 
+
+```csharp
+    source.IsEnabled() && source.IsEnabled("Example.MyLibrary.MyOperation")
+```
+
+The parameterless ```source.IsEnabled()``` check should be put before any other as it is very efficient and can virtually prevent any overhead in absence of any listener for given diagnostic source.
```
pharring: Is it too late to modify DiagnosticSource to incorporate the functionality of the parameterless IsEnabled() within the overload that takes a string? i.e. the first line of `bool IsEnabled(string activityName)` does a short-circuit test: `if (!IsEnabled()) return false;`

pharring: Oh, I see. You are trying to avoid the virtual call.

MS-TimothyMothra:
@SergeyKanzhelev is it safe to merge this?
Liudmila already signed off, but it sounds like you were requesting additional changes.

SergeyKanzhelev:
@MS-TimothyMothra let's wait a little bit on this one. I'll merge when time will come. Thanks for asking

MS-TimothyMothra:
Thanks @SergeyKanzhelev, consider blocking the merge until you're ready.

PR #790, opened 2018-01-12
Fix a corner case test failure.

cijothomas:
Fix issue # .
The very 1st test in StatusMonitor suite failed couple times.  Telemetry was added to find root cause. It looks like after SM is installed, the very 1st request could take a very long time. (15+ secs as per newly added test telemetry).
This adds a sleep of 20 secs before the very 1st test only for SM cases to avoid this rare failure.

Short description of the fix:

- [ ] I ran all unit tests locally
- [ ] CHANGELOG.md updated or do not need to be updated
  - If not tell why: 

- [ ] I ran [functional tests](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/CONTRIBUTING.md#functional-tests) locally.
- [ ] The PR will trigger build, unit test, and functional tests automatically. If your PR submitted from fork - mention one of committers to initiate the build for you.

### DIFF
--- a/SDL/FxCopSetup.ps1
+++ b/SDL/FxCopSetup.ps1
@@ -7,12 +7,39 @@ Param(
     [string]$fxCopDirectory = (Join-Path -Path (Split-Path -parent (Split-Path -parent (Split-Path -parent $PSCommandPath))) -ChildPath "fxCop")
 )
 
+function IsFileDependency {
+    [CmdletBinding()]
+    param
+    (    
+    $file
+    )
+        $dependencyFiles | ForEach-Object {
+            if($file.Name -eq $_) {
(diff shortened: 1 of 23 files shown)


d3r3kk:
Is there no way to wait for a concrete event or for a status check to discover when it is safe to begin testing? 

Perhaps a sacrificial request that you wait for (and timeout on, should the request take too long, in a highly visible/reportable way).

cijothomas:
@d3r3kk Yes - i made the change as per your suggestion! We already had a 'warm-up' request. Now waiting till telemetry from that request arrives beforestarting any tests.
(Merging this now to get builds back to green, please let me know if any concerns)

Issue #782, opened 2018-01-10
SQL Dependency collection with DiagnosticSourceListener may fail with System.InvalidCastException


Dmitry-Matveev:
Newly added SQL DiagnosticSourceListener (2.5-beta1) may fail with an exception like the following:
![image](https://user-images.githubusercontent.com/12158043/34752465-5fde9c84-f567-11e7-877c-e7716d6d86ca.png)

and

![image](https://user-images.githubusercontent.com/12158043/34752475-6926c4d8-f567-11e7-90fc-9751a7a9d32d.png)

@anpete took a look at this issue and commented with the following:

"Looks like a bug in the listener. For some reason SqlClient is [including](https://github.com/dotnet/corefx/blob/master/src/System.Data.SqlClient/src/System/Data/SqlClient/SqlClientDiagnosticListenerExtensions.cs#L289) a TransactionName property in transaction rollback messages (but not in transaction commit messages).

The fix is to add separate “[fetchers](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared/SqlClientDiagnosticSourceListener.cs#L482)” for the rollback message data."

PR #792, opened 2018-01-17
add separate fetchers for rollback and commit message data

MS-TimothyMothra:
Fix issue #782  .

Short description of the fix:
SqlClient is including a TransactionName property in transaction rollback messages (but not in transaction commit messages). Added a separate “fetchers” for the rollback message data.

- [x] I ran all unit tests locally
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+
+## Version 2.5.0
+- [Fix: System.InvalidCastException for SOL Dependency](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/782)
+
+
 ## Version 2.5.0-beta2
 - [Fix: When debugging netcoreapp2.0 in VS, http dependencies are tracked twice](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/723)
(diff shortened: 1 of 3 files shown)


anpete:
I think you need to handle the transaction After and Error cases, too. Add TransactionName to the message in this unit test: https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs#L508

anpete:
Consider adding a unit test for the Rollback error case.

Dmitry-Matveev:
@anpete, I looked and `After` and `Error` fetchers, they seem to only care about the subset of the available fields anyway - do you expect them to fail as well if there no 1:1 match? @MS-TimothyMothra can expand this PR to modify fetchers for those as well and add them into the separate blocks within `switch` if required - I would just like to confirm with you if it is.

anpete:
@Dmitry-Matveev IIRC the key thing is that a fetcher may never be used with more than one incoming message type (anonymous object shape). We need to have unit test coverage for all of the events we handle, and in those tests we should be careful to construct exact replicas of the actual SQL Client message payloads.

MS-TimothyMothra:
I'm marking this PR as resolved and will confirm that we have unit tests for every possible type in a follow up issue: #796

PR #795, opened 2018-01-19
Add Azure Instance Metada properties to heartbeat

d3r3kk:
(Associated) Completion of issue Microsoft/ApplicationInsights-dotnet#666, originally started in PR Microsoft/ApplicationInsights-dotnet#636

This is for adding the remainder of the proposed default payload fields for the heartbeat feature. These fields are obtained from the Azure Instance Metadata service present on Azure VMs.

- [x] Design discussion issue Microsoft/ApplicationInsights-dotnet#666 & Microsoft/ApplicationInsights-dotnet#628
- [x] CHANGELOG.md updated with one line description of the fix.


d3r3kk:
@SergeyKanzhelev and @cijothomas I have completed this change, the Heartbeat properties for Azure IMS are now included via the WindowsServer package. We can discuss ways to get into the .NET packages for Linux/Mac after this.

Note that I've proven out my App2Infra end to end scenario with this change + my change in Microsoft/ApplicationInsights-dotnet#681.

Note that there is a small unrelated fix to TargetFrameworks in one of the testing project files. I noticed that the .NET runtime for 2.0/3.5 is required without this fix.

Src/WindowsServer/WindowsServer.Shared.Tests/AzureInstanceMetadataTests.cs
```diff
@@ -0,0 +1,53 @@
...
+        [TestMethod]
+        public void GetAzureInstanceMetadataFieldsAsExpected()
+        {
+            HeartbeatProviderMock hbeatMock = new HeartbeatProviderMock();
+            AzureInstanceMetadataRequestMock azureInstanceRequestorMock = new AzureInstanceMetadataRequestMock();
+            AzureHeartbeatProperties azureIMSFields = new AzureHeartbeatProperties(azureInstanceRequestorMock, true);
+            int counter = 1;
+            foreach (string field in azureIMSFields.DefaultFields)
+            {
+                azureInstanceRequestorMock.ComputeFields.Add(field, $"testValue{counter++}");
```
Dmitry-Matveev: Did you plan to validate the value of the counter as part of the test later on?

Src/WindowsServer/WindowsServer.Shared.Tests/AzureInstanceMetadataTests.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            foreach (string fieldName in azureIMSFields.DefaultFields)
+            {
+                Assert.True(hbeatMock.HbeatProps.ContainsKey(fieldName));
+                Assert.False(string.IsNullOrEmpty(hbeatMock.HbeatProps[fieldName]));
+            }
+        }
+
+        [TestMethod]
+        public void FailToObtainAzureInstanceMetadataFieldsAltogether()
```
Dmitry-Matveev: Will it be closer to the actual behavior if you make IMS request to fail via mock and then somehow prompt it to execute and validate that only default properties are present?

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureHeartbeatProperties.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        public async Task<bool> SetDefaultPayload(IEnumerable<string> disabledFields, IHeartbeatPropertyManager provider)
+        {
+            bool hasSetFields = false;
+
+            if (!AzureHeartbeatProperties.isAzureMetadataCheckCompleted)
+            {
+                // only ever attempt this once
+                AzureHeartbeatProperties.isAzureMetadataCheckCompleted = true;
```
Dmitry-Matveev: If SetDefaultPayload is invoked twice at the same time, the communication with IMS may still be attempted twice due to positive concurrency (both calls may pass a check above before they set `isAzureMetadataCheckCompleted` to true.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureHeartbeatProperties.cs
```diff
@@ -0,0 +1,95 @@
...
+
+                var allFields = await this.azureInstanceMetadataRequestor.GetAzureInstanceMetadataComputeFields()
```
Dmitry-Matveev: Do we consider metadata check completed even if this fails due to legitimate reason on IMS metadata side (e.g. some throttling at the moment we tried to access it)? If not - then we might need to move `isAzureMetadataCheckCompleted` assignment after this call.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
```diff
@@ -0,0 +1,108 @@
...
+#endif
+    using Extensibility;
+
+    internal class AzureMetadataRequestor : IAzureMetadataRequestor
+    {
+        /// <summary>
+        /// Azure Instance Metadata Service exists on a single non-routable IP on machines configured
+        /// by the Azure Resource Manager. See <a href="https://go.microsoft.com/fwlink/?linkid=864683">to learn more.</a>
+        /// </summary>
+        private static string baseImdsUrl = $"http://169.254.169.254/metadata/instance/compute";
```
Dmitry-Matveev: If this code is invoked on non-azure machine, it looks like a malicious agent can spoof this http destination for us. What are the consequences for us in this case? E.g. an attacker makes our calls to this service slow or return gigabytes of data to process, or returns data which is an injection of some type (e.g. JS/XML/SQL), so later in the portal it exploits some issue. And along those lines..

Src/WindowsServer/WindowsServer.Shared.Tests/AzureInstanceMetadataTests.cs
d3r3kk: No, this was a completely unnecessary addition. I've removed it.

Src/WindowsServer/WindowsServer.Shared.Tests/AzureInstanceMetadataTests.cs
d3r3kk: Agreed. Throwing an exception in the Mock requestor approximates the behavior of a proper failure now.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureHeartbeatProperties.cs
d3r3kk: Understood, and there is the possibility that this could happen regardless (what if someone instantiated two objects that call this?). 

It is unimportant however as the underlying heartbeat manager will just happily set the fields twice without issue, there is no problematic contention here.

Admittedly, this was a leftover concept from when I had it in the hearbeat manager, and I should have removed it. 

The `isAzureMetadataCheckCompleted` will now be used to reduce the likelihood it gets called twice in any single instance of `AzureHeartbeatProperties`.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureHeartbeatProperties.cs
d3r3kk: Yes, let's consider it failed at this point in the code. Any retries we decide to implement can be handled within the `IAzureMetadataRequestor` implementation, should we deem that important enough to do.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
d3r3kk: Interesting, an attacker could most certainly cause trouble for us here, even outside of the current process! To mitigate this we could test to see that we are running on Azure hardware first (an issue raised by Zaki earlier actually, one that I've not resolved thus far) but there is no clear way for us to determine that yet.

Firstly, we call this in a separate thread and we ignore return values/exceptions. If the call fails the heartbeat simply doesn't get any IMS data in the properties and the SDK allows the caller to run regardless of the outcome, so I think delays/timeouts are not an issue.

I've altered the code to ensure we only pull a maximum number of bytes when we receive a request, anything over this buffer amount will cause the code to simply fail.

I'm not sure how to ensure there is no injection/exploit though - is there something I can refer to in order to guard against his sort of thing? (I'll start poking around and see what I can find on my own of course!).

d3r3kk: One last point here, we are also only limiting the response to exactly a known set of fields, for specifics please take a look at `AzureHeartbeatProperties.DefaultFields` and then `AzureHeartbeatProperties.SetDefaultPayload` where I set `enabledImdsFields` to the intersection of the known fields with the fields extracted from IMS.

I believe this will further mitigate any risk (along with my limit of the content received being less than 256 bytes now).

d3r3kk:
@Dmitry-Matveev all through resolving your feedback, and ensuring builds/tests pass. Please let me know if you are satisfied with the result.

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryInitializer.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    using System.Threading.Tasks;
+    using Microsoft.ApplicationInsights.Channel;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+    using Microsoft.ApplicationInsights.WindowsServer.Implementation;
+
+    /// <summary>
+    /// A telemetry initializer that will gather Azure Role Environment context information.
```
cijothomas: minor - incorrect comment.

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryInitializer.cs
```diff
@@ -0,0 +1,53 @@
...
+            foreach (var module in telemetryModules.Modules)
+
+            {
+                if (module is IHeartbeatPropertyManager)
+                {
+                    var hbeatManager = (IHeartbeatPropertyManager)module;
+
+                    // start off the heartbeat property collection process, but don't wait for it nor report
+                    // any status from here. The thread running the collection will report to the core event log.
+                    var hbeatPropertyHandler = new AzureHeartbeatProperties();
```
cijothomas: variable name does not reflect class/purpose

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
```diff
@@ -0,0 +1,120 @@
...
+                {
+                    getFieldValueClient.DefaultRequestHeaders.Add("Metadata", "True");
+                    requestResult = await getFieldValueClient.GetStringAsync(metadataRequestUrl).ConfigureAwait(false);
+                }
+
+#elif NET45 || NET46
+
+                WebRequest request = WebRequest.Create(metadataRequestUrl);
+                request.Method = "GET";
+                request.Headers.Add("Metadata", "true");
```
cijothomas: netstandard does true with uppercase 'T'. (likely irrelevant)

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
```diff
@@ -0,0 +1,120 @@
...
+        private async Task<string> MakeAzureMetadataRequest(string metadataRequestUrl)
+        {
+            string requestResult = string.Empty;
+
+            SdkInternalOperationsMonitor.Enter();
+            try
+            {
+#if NETSTANDARD1_3
+
+                using (var getFieldValueClient = new HttpClient(new HttpClientHandler() { MaxRequestContentBufferSize = MAX_IMS_RESPONSE_BUFFER_SIZE }))
```
cijothomas: MaxRequestContentBufferSize - did you meant to use ResponseContentBufferSize?

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
```diff
@@ -0,0 +1,120 @@
...
+                                int bufferIndex = await content.ReadAsync(buffer, 0, buffer.Length);
+
+                                // this will probably never exceed the buffer's size in bytes returned, scrap anything else that is left
+                                if (bufferIndex < buffer.Length - 1)
+                                {
+                                    requestResult = buffer.ToString();
+                                }
+                                else
+                                {
+                                    WindowsServerEventSource.Log.AzureInstanceMetadataRequestFailure(metadataRequestUrl, "Content received from Azure Metadata Instance service exceeds expected size, failing the call to avoid potential attack", string.Empty);
```
cijothomas: these error logging is not done for netstandard

Src/WindowsServer/WindowsServer.Shared.Tests/AzureInstanceMetadataTests.cs
```diff
@@ -0,0 +1,64 @@
+namespace Microsoft.ApplicationInsights.WindowsServer
```
cijothomas: AzureMetadataRequestor is the class which does the actual work of querying Azure endpoint, and it looks like it is not tested fully.
Can you unit test it by mocking the endpoint it queries? It should validate scenarios like the endpoint does not respond, responds with non sucess, responds with HUGE payload etc.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
cijothomas: MAX_IMS_RESPONSE_BUFFER_SIZE - where is this constant defined?

Src/WindowsServer/WindowsServer.Shared.Tests/AzureInstanceMetadataTests.cs
d3r3kk: I stuck to unit tests this far, but I agree we do need to test through this path as well. I'll add some dependency injection feature for this class so I can get around the actual call to the URL (and mock it out). This might take until EOD.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
d3r3kk: No idea how this got through the compiler! Fixed.

d3r3kk: I don't think so, I think I'm looking for `MaxRequestContentBufferSize` as the intent is to say "I expect all requests to be smaller than this maximum, anything larger I want to fail and bail".

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryInitializer.cs
d3r3kk: Fixed

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryInitializer.cs
d3r3kk: Agreed. Fixed.

Src/WindowsServer/WindowsServer.Shared.Tests/AzureInstanceMetadataTests.cs
cijothomas: https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/NetCore.Tests/DependencyTrackingTelemetryModuleTestNetCore.cs#L191
We have lot of examples to mock any http endpoint. Please take a look at them.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
cijothomas: What Dmitry raised earlier was - what if some other process(may be malicious) is listening on this Azure Metadata IP Address, and that process returns a HUGE payload in its response. We need to make sure we don't read all that HUGE response. So its not the request we should be limiting size, it should be the response.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
cijothomas: No - what i meant is in the above code, you have an If {netstandard} {} block and if{net45} {} block. The error logging is not done in the netstandard block.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
Dmitry-Matveev: Even with these precautions we will allow someone to submit normal size but completely wrong values into someone else's telemetry, including, but not limited to the injections of some sort. That may diminish trust in our SDK as a trustworthy telemetry source. Is there a way to make sure we only execute this code on Azure where the service is actually available?

As per my understanding, that endpoint does not have any protected connection type like SSL that we can use, let's try to protect SDK by other means then.

d3r3kk:
@Dmitry-Matveev I have covered all of the topics discussed and am ready to merge unless you have any concerns. Please review this morning?

MS-TimothyMothra:
@d3r3kk can you please update the Changelog.md for this? That'll help me out a bunch when I start the next release.

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryInitializer.cs
```diff
@@ -0,0 +1,53 @@
...
+    using Microsoft.ApplicationInsights.Channel;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+    using Microsoft.ApplicationInsights.WindowsServer.Implementation;
+
+    /// <summary>
+    /// A telemetry initializer that adds Azure specific context information to the heartbeat, if it is available.
+    /// </summary>
+    public class AzureInstanceMetadataTelemetryInitializer : ITelemetryInitializer
```
cijothomas: Is this initializer meant to be automatically added to ai.config? If yes, ApplicationInsights.config.install.xdt need updated.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureHeartbeatProperties.cs
```diff
@@ -0,0 +1,116 @@
...
+            "subscriptionId",
+            "resourceGroupName"
+        };
+
+        /// <summary>
+        /// Flags that will tell us whether or not Azure VM metadata has been attempted to be gathered or not, and
+        /// if we should even attempt to look for it in the first place. Static to ensure we only ever attempt this
+        /// once for any given consuming application.
+        /// </summary>
+        private bool isAzureMetadataCheckCompleted = false;
```
cijothomas: Comment mentions this is static but actually not? Please update comment/code.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureHeartbeatProperties.cs
```diff
@@ -0,0 +1,116 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Add all enabled, present Azure IMS fields to the heartbeat properties.
+        /// </summary>
+        /// <param name="disabledFields">Fields that are to be disabled.</param>
+        /// <param name="provider">Heartbeat provider to set the properties on.</param>
+        /// <returns>True if any property values were successfully set, false if none were set.</returns>
+        public async Task<bool> SetDefaultPayload(IEnumerable<string> disabledFields, IHeartbeatPropertyManager provider)
```
cijothomas: IHeartbeatPropertyManager  already knows the list of disable fields. No need for passing this additional argument here.
https://github.com/Microsoft/ApplicationInsights-dotnet/blob/217297b03b8190947cac57e8895523d5e2c10235/src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/IHeartbeatPropertyManager.cs#L26

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureHeartbeatProperties.cs
```diff
@@ -0,0 +1,116 @@
...
+                        hasSetFields = hasSetFields || addedProperty;
+                    }
+                }
+                else
+                {
+                    WindowsServerEventSource.Log.AzureInstanceMetadataNotAdded();
+                }
+            }
+
+            return hasSetFields;
```
cijothomas: where is this return value used?

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureHeartbeatProperties.cs
```diff
@@ -0,0 +1,116 @@
...
+
+                var azureComputeMetadata = await this.azureInstanceMetadataRequestor.GetAzureComputeMetadata()
+                                .ConfigureAwait(false);
+
+                if (azureComputeMetadata != null)
+                {
+                    var enabledImdsFields = this.ExpectedAzureImsFields.Except(disabledFields);
+                    foreach (string field in enabledImdsFields)
+                    {
+                        string value = azureComputeMetadata.GetValueForField(field);
```
cijothomas: this call is not reuqired

d3r3kk:
@MS-TimothyMothra done, changelog updated.

Src/WindowsServer/WindowsServer.Shared/Implementation/DataContracts/AzureInstanceComputeMetadata.cs
```diff
@@ -0,0 +1,177 @@
...
+        [DataMember(Name = "vmSize", IsRequired = true)]
+        internal string VmSize { get; set; }
+
+        [DataMember(Name = "subscriptionId", IsRequired = true)]
+        internal string SubscriptionId { get; set; }
+
+        [DataMember(Name = "resourceGroupName", IsRequired = true)]
+        internal string ResourceGroupName { get; set; }
+
+        internal string GetValueForField(string fieldName)
```
cijothomas: why 2 separate methods? can we just combine the fetching logic and sanitize logic into one?

Src/WindowsServer/WindowsServer.Shared/Implementation/DataContracts/AzureInstanceComputeMetadata.cs
```diff
@@ -0,0 +1,177 @@
...
+            }
+            else if (fieldName.Equals("subscriptionId", StringComparison.OrdinalIgnoreCase))
+            {
+                Guid g = new Guid();
+                valueOk = Guid.TryParse(valueToVerify, out g);
+
+                if (valueOk)
+                {
+                    value = valueToVerify;
+                }
```
cijothomas: what happens when valueOk is false?

Src/WindowsServer/WindowsServer.Shared/Implementation/DataContracts/AzureInstanceComputeMetadata.cs
```diff
@@ -0,0 +1,177 @@
...
+
+        /// <summary>
+        /// Because the Azure IMS is on a non-routable IP we need to do some due diligence in our accepting
+        /// values returned from it. This method takes the fieldname and value received for that field, and
+        /// if we can test that value against known limitations of that field we do so here. If the test fails
+        /// we return the empty string, otherwise we return the string given.
+        /// </summary>
+        /// <param name="fieldName">Name of the field acquired from the call to Azure IMS.</param>
+        /// <returns>The value of the field, verified, or the empty string.</returns>
+        internal string VerifyExpectedValue(string fieldName)
```
cijothomas: please merge this into the above method itself.

Src/WindowsServer/WindowsServer.Shared/Implementation/DataContracts/AzureInstanceComputeMetadata.cs
```diff
@@ -0,0 +1,177 @@
...
+    using System.Linq;
+    using System.Runtime.Serialization;
+    using System.Text.RegularExpressions;
+
+    /// <summary>
+    /// Class representing the returned structure from an Azure Instance Metadata request
+    /// for Compute information.
+    /// </summary>
+    [DataContract]
+    internal class AzureInstanceComputeMetadata
```
cijothomas: https://github.com/Microsoft/ApplicationInsights-dotnet-server/tree/develop/Src/Common - @MS-TimothyMothra has made lot of helper classes already available here for sanitize etc. Re-use if applicable.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureHeartbeatProperties.cs
```diff
@@ -0,0 +1,116 @@
+﻿namespace Microsoft.ApplicationInsights.WindowsServer.Implementation
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Threading.Tasks;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+
+    internal class AzureHeartbeatProperties
```
cijothomas: Please rename to indicate the true purpose of this class. Its not just a list of properties, so current name does not do full justice

Src/WindowsServer/WindowsServer.Shared.Tests/AzureInstanceMetadataTests.cs
cijothomas: It does not look like this comment is addressed?

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
```diff
@@ -0,0 +1,113 @@
...
+        private async Task<AzureInstanceComputeMetadata> MakeAzureMetadataRequest(string metadataRequestUrl)
+        {
+            AzureInstanceComputeMetadata requestResult = null;
+
+            SdkInternalOperationsMonitor.Enter();
+            try
+            {
+                if (this.azureIMSRequestor != null)
+                {
+                    requestResult = await this.azureIMSRequestor(metadataRequestUrl);
```
cijothomas: can we simplify the code here by just calling the this.azureIMSRequestor() - and let constructor worry about which method this.azureIMSRequestor points to?

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
```diff
@@ -0,0 +1,113 @@
...
+        internal TimeSpan AzureImsRequestTimeout = TimeSpan.FromSeconds(10);
+
+        /// <summary>
+        /// Private function for mocking out the actual call to IMS in unit tests. Available to internal only.
+        /// </summary>
+        private Func<string, Task<AzureInstanceComputeMetadata>> azureIMSRequestor = null;
+
+        private string baseImdsUri = $"http://169.254.169.254/metadata/instance/compute";
+
+        internal AzureMetadataRequestor(Func<string, Task<AzureInstanceComputeMetadata>> makeAzureIMSRequestor = null)
```
cijothomas: What is the string parameter here? Is it used anywhere?

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryInitializer.cs
cijothomas: @SergeyKanzhelev to confirm if this is needed.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureHeartbeatProperties.cs
```diff
@@ -0,0 +1,116 @@
...
+    {
+        internal static string HeartbeatPropertyPrefix = "azInst_"; // to ensure no collisions with base heartbeat properties
+
+        /// <summary>
+        /// Expected fields extracted from Azure IMS to add to the heartbeat properties. 
+        /// Set as internal for testing.
+        /// </summary>
+        internal readonly List<string> ExpectedAzureImsFields = new List<string>()
+        {
+            "osType",
```
cijothomas: osType already added in base sdk ?

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureHeartbeatProperties.cs
```diff
@@ -0,0 +1,116 @@
...
+            "offer",
+            "platformFaultDomain",
+            "platformUpdateDomain",
+            "publisher",
+            "sku",
+            "version",
+            "vmId",
+            "vmSize",
+            "subscriptionId",
+            "resourceGroupName"
```
cijothomas: there should be some user warning/info about collecting these fields in the comment in this Initializer, if we make this pat of default ai.config.

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryInitializer.cs
pharring: Considering the Initialize method is blank, does it really need to be an ITelemetryInitializer? Could it be an ITelemetryModule instead?

Src/WindowsServer/WindowsServer.Shared/Implementation/DataContracts/AzureInstanceComputeMetadata.cs
```diff
@@ -0,0 +1,177 @@
...
+    {
+        private static int resourceGroupNameLengthMax = 90;
```
pharring: Can these (int and string fields) be `const` instead of `static`?

Src/WindowsServer/WindowsServer.Shared.Tests/AzureInstanceMetadataTests.cs
d3r3kk: It is, but in the interest of getting this initial (beta) functionality in place for beta1 I created an issue to track it and pushed it to beta2 (plus, there will likely be further functional/integration tests further to the ones you brought up as well). See issue #832.

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryInitializer.cs
d3r3kk: @cijothomas I think I might not understand this comment… If we don’t create this initializer the heartbeat will never be updated?

@pharring I’m not aware of the difference, or the cost/benefit of choosing Module over Initializer. I need a place to do an initial ‘fire and forget’ to populate some properties into a cache. Is there performance or other considerations that makes a `TelemetryModule` more adequate? I chose `TelemetryInitializer` as that is the true nature of what I’m trying to accomplish, but I am happy to be corrected if there is a consideration I’ve missed.

Src/WindowsServer/WindowsServer.Shared/Implementation/DataContracts/AzureInstanceComputeMetadata.cs
d3r3kk: I would argue that couples the two different concerns a bit too much.

Src/WindowsServer/WindowsServer.Shared/Implementation/DataContracts/AzureInstanceComputeMetadata.cs
d3r3kk: If you look further down at line ~170 you will see that the valueOk flag is tested for false, and if it is false a log message is written to the WindowsServerEventSource object. Also, if the valueOk flag is false, the valueToVerify is never returned, effectively ‘sanitizing’ the field.

Src/WindowsServer/WindowsServer.Shared/Implementation/DataContracts/AzureInstanceComputeMetadata.cs
d3r3kk: I respectfully disagree, I believe this to be too much coupling.

Src/WindowsServer/WindowsServer.Shared/Implementation/DataContracts/AzureInstanceComputeMetadata.cs
d3r3kk: Very interesting! I didn’t know about that group of utilities. I would like to push this out as a new issue though, to be handled in beta2…

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureHeartbeatProperties.cs
d3r3kk: Agreed. Although a tad verbose, I have changed this to `AzureComputeMetadataHeartbeatPropertyProvider`.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
d3r3kk: This is an interesting piece of code – I went with what I saw as readability over simplicity in code-calls. I wanted to be very clear to the reader of this code what was going on. I can be convinced to simplify it here  as you suggest but I’d like to discuss further first. Please raise an issue to be addressed in beta2 if you wish.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
d3r3kk: That’s up to the caller to decide. For instance, if we point this functor at ‘MakeWebRequest’ the string value is the Uri to call to obtain the JSON response. This method will be used only for Test scenarios. I’ve updated the docs to reflect this.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureHeartbeatProperties.cs
d3r3kk: Yes it is, and we can opt to remove this redundant data later on. However, I wanted to capture it going forward in case there was any use case for this osType value vs. the osType value returned by the .NET framework. (For instance, an Azure-specific osType moniker).

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureHeartbeatProperties.cs
d3r3kk: I’ve added a comment to the install.xdt file similar to the comment for perf counters.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureHeartbeatProperties.cs
d3r3kk: Done

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureHeartbeatProperties.cs
d3r3kk: Agreed, done.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureHeartbeatProperties.cs
d3r3kk: It is used in testing only, to assert that at least one value was added to the dictionary. It was also put there to follow best practices in returning from an async method.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureHeartbeatProperties.cs
d3r3kk: Good catch, thank you.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureHeartbeatProperties.cs
```diff
@@ -6,7 +6,7 @@
     using System.Threading.Tasks;
     using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
 
-    internal class AzureHeartbeatProperties
+    internal class AzureComputeMetadataHeartbeatPropertyProvider
     {
         internal static string HeartbeatPropertyPrefix = "azInst_"; // to ensure no collisions with base heartbeat properties
```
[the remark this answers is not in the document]
d3r3kk: Yes, this should be const. Changed.

Src/WindowsServer/WindowsServer.Shared/Implementation/DataContracts/AzureInstanceComputeMetadata.cs
d3r3kk: Yes, and they've been changed.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
```diff
@@ -0,0 +1,115 @@
...
+        internal TimeSpan AzureImsRequestTimeout = TimeSpan.FromSeconds(10);
+
+        /// <summary>
+        /// Private function for mocking out the actual call to IMS in unit tests. Available to internal only.
+        /// </summary>
+        /// parameter sent to the func is a string representing the Uri to request Azure IMS data from.
+        /// <returns>An instance of AzureInstanceComputeMetadata or null.</returns>
+        private Func<string, Task<AzureInstanceComputeMetadata>> azureIMSRequestor = null;
+
+        private string baseImdsUri = $"http://169.254.169.254/metadata/instance/compute";
```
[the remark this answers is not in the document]
d3r3kk: done.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
```diff
@@ -0,0 +1,115 @@
...
+    using Extensibility;
+    using Microsoft.ApplicationInsights.WindowsServer.Implementation.DataContracts;
+
+    internal class AzureMetadataRequestor : IAzureMetadataRequestor
+    {
+        /// <summary>
+        /// Azure Instance Metadata Service exists on a single non-routable IP on machines configured
+        /// by the Azure Resource Manager. See <a href="https://go.microsoft.com/fwlink/?linkid=864683">to learn more.</a>
+        /// </summary>
+        internal static string AzureImsApiVersion = $"api-version=2017-08-01"; // this version has the format=text capability
```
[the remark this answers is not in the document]
d3r3kk: Agreed.

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryInitializer.cs
pharring: Yes, performance. The `Initialize` method of an `ITelemetryInitializer` will be called for every tracked telemetry item. Even an empty method has an overhead (it's a virtual call through the interface). On the other hand an `ITelemetryModule` has an `Initialize` method that is called just once when the module is added to the telemetry pipeline. It's more appropriate for the fire-and-forget task (move it out of the constructor into the `ITelemetryModule.Initialize` method)

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryInitializer.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            WindowsServerEventSource.Log.TelemetryInitializerLoaded(this.GetType().FullName);
+
+            var telemetryModules = TelemetryModules.Instance;
+            foreach (var module in telemetryModules.Modules)
+
+            {
+                if (module is IHeartbeatPropertyManager)
+                {
+                    var hbeatManager = (IHeartbeatPropertyManager)module;
```
pharring: There's a C#6 'pattern matching' syntax for doing the "is" and "cast" in one line:
```C#
if (module is IHeartbeatPropertyManager hbeatManager)
```

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryInitializer.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// Metadata service if it is present and available to the currently running process. If it is not
+        /// present no added IMS data is added to the heartbeat.
+        /// </summary>
+        public AzureInstanceMetadataTelemetryInitializer()
+        {
+            WindowsServerEventSource.Log.TelemetryInitializerLoaded(this.GetType().FullName);
+
+            var telemetryModules = TelemetryModules.Instance;
+            foreach (var module in telemetryModules.Modules)
+
```
pharring: remove blank line

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryInitializer.cs
pharring: Now I look closer at what you're doing within the constructor (enumerating all previous telemetry modules), perhaps that wouldn't be appropriate because you can't ensure you're the last module to be loaded (or at least after all the other `IHeartbeatPropertyManager` instances). Indeed, even as an `ITelemetryInitializer` you're relying on an implementation detail that they are constructed after all modules are created.

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryInitializer.cs
d3r3kk: Yes, does our minimum .NET SDK (4.5) support this? When I wrote this originally I got compiler warnings for v4.0 not supporting it...

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryInitializer.cs
d3r3kk: Yes, this could be problematic in the future if we change the implementation of our startup/initialization. It would be better if I could simply get 'plugged into' a module (the base SDK's `HeartbeatProvider` module, in this case) and the provider module could then initiate the heartbeat property acquisition when it was entirely ready.

Let's take this off of this PR into an issue perhaps, where I can get some further input and make necessary changes in the next beta release?

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryInitializer.cs
d3r3kk: Done

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryInitializer.cs
d3r3kk: Changed it, I'll let the build tell me if it's supported now or not.

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryInitializer.cs
d3r3kk: I've discussed this with @Dmitry-Matveev and we agree it should be an `ITelemetryModule` for now. There are still some things to consider here regarding order of initialization that you brought up which we will need to handle. I am indeed leveraging knowledge of the implementation of baseSDK here, the `DiagnosticsTelemetryModule` always is added to the pipeline before all other modules, and it owns the `IHeartbeatPropertyManager` implementation.

Raised issue #834 for us to track, will be slated for beta2 timeframe.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureComputeMetadataHeartbeatPropertyProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+
+    internal class AzureComputeMetadataHeartbeatPropertyProvider
+    {
+        internal const string HeartbeatPropertyPrefix = "azInst_"; // to ensure no collisions with base heartbeat properties
+
+        /// <summary>
+        /// Expected fields extracted from Azure IMS to add to the heartbeat properties. 
+        /// Set as internal for testing.
+        /// </summary>
+        internal readonly List<string> ExpectedAzureImsFields = new List<string>()
```
pharring: Consider typing as `IReadOnlyCollection` and using a string array instead of `List<string>`.
i.e.
```C#
internal readonly IReadOnlyCollection<string> ExpectedAzureImsFields = new string[]
{
    "location",
    //etc...
};
```

That makes it clear that it is supposed to be immutable. (Technically, you could still cast it back to a `string[]` and mutate the contents, but you have to go out of your way.)
```

File: Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs [hunk 33]
```diff
@@ -0,0 +1,115 @@
...
+        /// <summary>
+        /// Gets or sets the base URI for the Azure Instance Metadata service. Internal to allow overriding in test.
+        /// </summary>
+        internal string BaseAimsUri
+        {
+            get => this.baseImdsUri;
+            set => this.baseImdsUri = value;
+        }
+
+        public async Task<AzureInstanceComputeMetadata> GetAzureComputeMetadata()
```
pharring: Couple of comments.
1. The usual naming convention is for async (task-returning) methods to have the suffix `Async`. i.e. `GetAzureComputeMetadataAsync`.
2. When you have an async method that is simply a straight-line call to another async method, you can (and should) remove the "async" and "await" because that adds unnecessary (compiler-generated) overhead. i.e.
```C#
public Task<AzureInstanceComputeMetadata> GetAzureComputeMetadataAsync()
{
    string metadataRequestUri = $"etc.";
    return this.MakeAzureMetadataRequest(metadataRequestUrl);
}
```

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            string metadataRequestUrl = $"{this.BaseAimsUri}?{AzureMetadataRequestor.AzureImsJsonFormat}&{AzureMetadataRequestor.AzureImsApiVersion}";
+
+            AzureInstanceComputeMetadata jsonResponse = await this.MakeAzureMetadataRequest(metadataRequestUrl);
+
+            return jsonResponse;
+        }
+
+        private async Task<AzureInstanceComputeMetadata> MakeAzureMetadataRequest(string metadataRequestUrl)
```
pharring: Rename with `-Async` suffix

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureComputeMetadataHeartbeatPropertyProvider.cs
d3r3kk: Yes, that is the intention here, thanks for calling that out.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        /// <summary>
+        /// Gets or sets the base URI for the Azure Instance Metadata service. Internal to allow overriding in test.
+        /// </summary>
+        internal string BaseAimsUri
+        {
+            get => this.baseImdsUri;
+            set => this.baseImdsUri = value;
+        }
```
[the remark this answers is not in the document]
d3r3kk: Awesome, thanks.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
d3r3kk: Convention is always good to follow, makes it trivial to follow along. Thanks for calling this out.

To the compiler overhead issue, I will try and keep chained async calls out of my future PRs. However, in this case I'm going from a public API call `GetAzureComputeMetadataAsync()` all the way down to the call to `HttpClient`'s `.GetStreamAsync`. I realize I've placed a few interim methods in between the two (to try and keep the meaning of each method as concise as possible - potentially _too_ concise), but it's all about that `.GetStreamAsync` here.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
d3r3kk: Done.

d3r3kk:
@Dmitry-Matveev & @pharring all ready for a last check. I was able to convert the task to a `ITelemetryModule` without a problem.

@cijothomas I believe I have all your comments addressed in this PR, or I've raised issues to account for them in beta2.

@pharring I believe I've addressed all your comments as well, let me know if you have any further concerns.

Thanks for the review everyone, truly appreciated.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureComputeMetadataHeartbeatPropertyProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+
+    internal class AzureComputeMetadataHeartbeatPropertyProvider
+    {
+        internal const string HeartbeatPropertyPrefix = "azInst_"; // to ensure no collisions with base heartbeat properties
+
+        /// <summary>
+        /// Expected fields extracted from Azure IMS to add to the heartbeat properties. 
+        /// Set as internal for testing.
+        /// </summary>
+        internal readonly IReadOnlyCollection<string> ExpectedAzureImsFields = new List<string>()
```
pharring: I know we've been here before, but I recommend using a string array instead of a List. The reason being the List initializer syntax expands into a series of calls to `Add`. It will result in resizing the List as it goes. The string array initializer will set the size up-front. Just change `new List<string>()` to `new string[]`

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryModule.cs
```diff
@@ -0,0 +1,64 @@
...
+                        this.isInitialized = true;
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Dispose of this class, is expected to be disposed of by the infrastructure.
+        /// </summary>
+        public void Dispose()
```
pharring: Why make this class `IDisposable` if `Dispose()` does nothing?

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureComputeMetadataHeartbeatPropertyProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+                this.isAzureMetadataCheckCompleted = false;
+            }
+        }
+
+        /// <summary>
+        /// Add all enabled, present Azure IMS fields to the heartbeat properties.
+        /// </summary>
+        /// <param name="provider">Heartbeat provider to set the properties on.</param>
+        /// <returns>True if any property values were successfully set, false if none were set.</returns>
+        public async Task<bool> SetDefaultPayload(IHeartbeatPropertyManager provider)
```
pharring: Suggest: SetDefaultPayloadAsync

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+            this.azureIMSRequestor = makeAzureIMSRequestor;
+        }
+
+        /// <summary>
+        /// Gets or sets the base URI for the Azure Instance Metadata service. Internal to allow overriding in test.
+        /// </summary>
+        internal string BaseAimsUri { get; set; } = "http://169.254.169.254/metadata/instance/compute";
+
+        public async Task<AzureInstanceComputeMetadata> GetAzureComputeMetadataAsync()
```
pharring: I commented on this previously. It's not necessary to use `async` `await` if all you're doing in the body is calling another Task-returning method and returning its result directly. Replace with:
```C#
public Task<AzureInstanceComputeMetadata> GetAzureComputeMetadataAsync()
{
    string metadataRequestUrl = $"{etc...";
    return this.MakeAzureMetadataRequestAsync(metadataRequestUrl);
}
```

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
```diff
@@ -0,0 +1,109 @@
...
+        private async Task<AzureInstanceComputeMetadata> MakeAzureMetadataRequestAsync(string metadataRequestUrl)
+        {
+            AzureInstanceComputeMetadata requestResult = null;
+
+            SdkInternalOperationsMonitor.Enter();
+            try
+            {
+                if (this.azureIMSRequestor != null)
+                {
+                    requestResult = await this.azureIMSRequestor(metadataRequestUrl);
```
pharring: .ConfigureAwait(false);

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
```diff
@@ -0,0 +1,109 @@
...
+            AzureInstanceComputeMetadata azureIms = null;
+            DataContractJsonSerializer deserializer = new DataContractJsonSerializer(typeof(AzureInstanceComputeMetadata));
+
+            using (var azureImsClient = new HttpClient())
+            {
+                azureImsClient.MaxResponseContentBufferSize = AzureMetadataRequestor.AzureImsMaxResponseBufferSize;
+                azureImsClient.DefaultRequestHeaders.Add("Metadata", "True");
+                azureImsClient.Timeout = this.AzureImsRequestTimeout;
+
+                Stream content = await azureImsClient.GetStreamAsync(requestUrl);
```
pharring: Two things:
1. `.ConfigureAwait(false);`
2. Dispose of `content` (assuming the deserializer doesn't already close the stream)

Src/WindowsServer/WindowsServer.Shared/Implementation/DataContracts/AzureInstanceComputeMetadata.cs
```diff
@@ -0,0 +1,196 @@
...
+            string valueToVerify = this.GetValueForField(fieldName);
+
+            string value = string.Empty;
+            bool valueOk = true;
+            TimeSpan regexTimeout = TimeSpan.FromMilliseconds(AzureInstanceComputeMetadata.RegexTimeoutMs);
+
+            if (fieldName.Equals("resourceGroupName", StringComparison.OrdinalIgnoreCase))
+            {
+                valueOk = valueToVerify.Length <= AzureInstanceComputeMetadata.ResourceGroupNameLengthMax;
+                valueOk &= valueToVerify.Length >= AzureInstanceComputeMetadata.ResourceGroupNameLengthMin;
```
pharring: The technique of keeping a `bool` and masking it (`&=`) with each clause doesn't let the compiler do any short-circuiting. For example, if the first clause fails (the max length check) we still evaluate all the other clauses (including the regexp). Either combine the clauses into a single expression with `&&` or extract it out into a (static?) helper method with early return.

Src/WindowsServer/WindowsServer.Shared/Implementation/DataContracts/AzureInstanceComputeMetadata.cs
```diff
@@ -0,0 +1,196 @@
...
+    /// <summary>
+    /// Class representing the returned structure from an Azure Instance Metadata request
+    /// for Compute information.
+    /// </summary>
+    [DataContract]
+    internal class AzureInstanceComputeMetadata
+    {
+        private const int ResourceGroupNameLengthMax = 90;
+        private const int ResourceGroupNameLengthMin = 1;
+        private const string ResourceGroupNameValidChars = @"^[a-zA-Z0-9\.\-_]+$";
```
pharring: Do you really need to use RegExp for this? If this is called a lot (I think it is), then consider a zero-allocation static helper method.
e.g.
```C#
private static bool IsValidResourceGroupName(string s)
{
    if (string.IsNullOrEmpty(s)) return false;
    foreach(var ch in s)
    {
        if (!IsValidResourceGroupNameChar(ch)) return false;
    }

    return true;
}

private static bool IsValidResourceGroupNameChar(char ch)
{
    // todo
}
```

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureComputeMetadataHeartbeatPropertyProvider.cs
d3r3kk: Done. Best not to ever-expand in this case for sure.

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryModule.cs
d3r3kk: I was following the pattern of all the other `TelemetryModules`, thinking perhaps it was a requirement. Taking it out here.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureComputeMetadataHeartbeatPropertyProvider.cs
d3r3kk: Agreed. Done.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
d3r3kk: This seems odd to me, but I think I get it, the `async` part is not necessary to `await` - just that it returns a `Task<>`, thanks for the heads up.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
d3r3kk: Yes, missed that, thank you.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
d3r3kk: Again, thanks. I am not clear on the necessity of calling `Dispose` explicitly though - would have thought that the Stream going out of scope would take care of it? Perhaps there is a benefit to performance/memory usage here?

Src/WindowsServer/WindowsServer.Shared/Implementation/DataContracts/AzureInstanceComputeMetadata.cs
d3r3kk: Agreed. Changed.

Src/WindowsServer/WindowsServer.Shared/Implementation/DataContracts/AzureInstanceComputeMetadata.cs
d3r3kk: This is not called a lot. It is called exactly once each when Azure IMS service is present upon app startup (in a 'fire and forget' manner). Which means each of these regex calls happen exactly once each for a successful Azure IMS request.

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryModule.cs
```diff
@@ -0,0 +1,56 @@
...
+
+                        foreach (var module in telemetryModules.Modules)
+                        {
+                            if (module is IHeartbeatPropertyManager hbeatManager)
+                            {
+                                // start off the heartbeat property collection process, but don't wait for it nor report
+                                // any status from here, fire and forget. The thread running the collection will report 
+                                // to the core event log.
+                                var heartbeatProperties = new AzureComputeMetadataHeartbeatPropertyProvider();
+                                Task.Factory.StartNew(
```
Dmitry-Matveev: Do we need to make sure it does not throw, so it won't break the app? I see that we catch web request exception down the line - should we have a global handling as well right within this task to keep the app running in case of any other issues?

d3r3kk: Agreed. I've added a try...catch and an event log entry for this.

d3r3kk: And yes, I've opened issues for all outstanding requests from this PR and marked them as due for beta2 milestone.

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryModule.cs
```diff
@@ -40,10 +40,17 @@ public void Initialize(TelemetryConfiguration unused)
...
-                                    async () => await heartbeatProperties.SetDefaultPayloadAsync(hbeatManager)
-                                    .ConfigureAwait(false));
+                                try
+                                {
+                                    var heartbeatProperties = new AzureComputeMetadataHeartbeatPropertyProvider();
+                                    Task.Factory.StartNew(
+                                        async () => await heartbeatProperties.SetDefaultPayloadAsync(hbeatManager)
+                                        .ConfigureAwait(false));
+                                }
+                                catch (Exception heartbeatAquisitionException)
```
pharring: This will not work. Because the task returned from StartNew is not awaited (it's fire-and-forget), exceptions won't be surfaced here. If the code in SetDefaultPayloadAsync throws an exception, the task will be Faulted and, since there's no-one to observe it, when the task is eventually finalized, you'll get an UnobservedTaskException event raised in the scheduler.

Instead, you need to add the try/catch inside SetDefaultPayloadAsync.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureComputeMetadataHeartbeatPropertyProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+            "vmSize"
+        };
+
+        /// <summary>
+        /// Flags that will tell us whether or not Azure VM metadata has been attempted to be gathered or not, and
+        /// if we should even attempt to look for it in the first place. 
+        /// </summary>
+        private bool isAzureMetadataCheckCompleted = false;
+
+        private IAzureMetadataRequestor azureInstanceMetadataRequestor = null;
```
pharring: readonly ?

pharring: BTW, you don't need to explicitly initialize to null (and `false` on line 40). All fields are guaranteed to be default-initialized. I thought there was a compiler warning about that?

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureComputeMetadataHeartbeatPropertyProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// </summary>
+        /// <param name="azureInstanceMetadataHandler">For testing: Azure metadata request handler to use when requesting data from azure specifically. If left as null, an instance of AzureMetadataRequestor is used.</param>
+        /// <param name="resetCheckCompleteFlag">For testing: set to true to reset the check that we've already acquired this data.</param>
+        internal AzureComputeMetadataHeartbeatPropertyProvider(IAzureMetadataRequestor azureInstanceMetadataHandler = null, bool resetCheckCompleteFlag = false)
+        {
+            this.azureInstanceMetadataRequestor = azureInstanceMetadataHandler;
+            if (this.azureInstanceMetadataRequestor == null)
+            {
+                this.azureInstanceMetadataRequestor = new AzureMetadataRequestor();
+            }
```
pharring: You could rewrite these 4 lines as 1:
```C#
this.azureInstanceMetadataRequestor = azureInstanceMetadataHandler ?? new AzureMetadataRequestor();
```
Not only is the syntax compact, it's more efficient generated code because there is just one field access (instead of 2 or 3)

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureComputeMetadataHeartbeatPropertyProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            if (resetCheckCompleteFlag)
+            {
+                this.isAzureMetadataCheckCompleted = false;
```
pharring: I don't see the point of this. The `isAzureMetadataCheckCompleted` field is already false. Is it supposed to be a static field?

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
```diff
@@ -0,0 +1,109 @@
...
+            AzureInstanceComputeMetadata azureIms = null;
+            DataContractJsonSerializer deserializer = new DataContractJsonSerializer(typeof(AzureInstanceComputeMetadata));
+
+            using (var azureImsClient = new HttpClient())
+            {
+                azureImsClient.MaxResponseContentBufferSize = AzureMetadataRequestor.AzureImsMaxResponseBufferSize;
+                azureImsClient.DefaultRequestHeaders.Add("Metadata", "True");
+                azureImsClient.Timeout = this.AzureImsRequestTimeout;
+
+                Stream content = await azureImsClient.GetStreamAsync(requestUrl).ConfigureAwait(false);
```
pharring: Consider:
```C#
using( Stream content = await azureImsClientEtc... )
{
   azureIms = etc...
}
```
(On the assumption that ReadObject or the cast could throw)

Src/WindowsServer/WindowsServer.Shared/Implementation/DataContracts/AzureInstanceComputeMetadata.cs
pharring: OK, then.

Src/WindowsServer/WindowsServer.Shared/AzureInstanceMetadataTelemetryModule.cs
d3r3kk: I think perhaps catching them in both places may be prudent?

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureComputeMetadataHeartbeatPropertyProvider.cs
d3r3kk: Not sure I understand what you mean here. The `azureInstanceMetadataRequestor` field could be set to readonly, and set in the constructor and not given this default value of null, but what would that achieve?

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureComputeMetadataHeartbeatPropertyProvider.cs
d3r3kk: Sure, readability vs. performance I suppose was my reasoning here. But it does seem that this form is quite commonly used in our codebase.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureComputeMetadataHeartbeatPropertyProvider.cs
d3r3kk: Ah yes, leftover cruft code that I'd missed. That was indeed a static field at one point that I needed to reset. However, due to the usage pattern of this module I don't believe the use of the static field is altogether prudent.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
d3r3kk: Yes, I was thinking that way when you originally brought this up. However, I've seen in the past that nesting using blocks can cause problems, [a quick search results in finding articles such as this](https://superdevelopment.com/2014/03/13/common-pitfalls-with-idisposable-and-the-using-statement/).

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureComputeMetadataHeartbeatPropertyProvider.cs
d3r3kk: Interesting, the 2nd comment you gave me here didn't show up as I was going through them. No compiler warning that I've seen, but we definitely have a few warnings to clean up in our code base. Still curious about the benefit of making the requestor field readonly.

pharring: > Still curious about the benefit of making the requestor field readonly.

The general rule is: If the field (or property) is written to only in the constructor, then make it readonly. For private fields, it prevents unintended modification. It's an extra safety check. (In theory, it also opens up further optimizations but, in practice those are very rare).

Regarding the default-initialized fields: I know R# warns about them. I thought either the compiler, StyleCop or FxCop warned about them too. In any case, I think it's not generally accepted in the codebases I've worked on to do explicit default initialization and, indeed, it's frowned upon. This is the opposite of C++, but it is one of the design principals of .NET: eliminate memory errors due to garbage values.

Src/WindowsServer/WindowsServer.Shared/Implementation/AzureMetadataRequestor.cs
pharring: I skimmed through the article; nothing wrong with it, but I don't think it applies here because the two using statements aren't connected. Even so, ReadObject or the cast could throw, so use try/finally.

PR #801, opened 2018-01-29
fix .Test project references

MS-TimothyMothra:
Removing duplicate files in our test projects. Resolved as one file in Shared project.
Resolving hidden files in test projects by adding to project.
Removed one empty project not referenced (TestFramework.Net45).
Resolved several compiler warnings;

- CS1591 - missing XML comments in test projects.
- CS0436 - conflicting Assembly.cs in the shared test project.


Src/DependencyCollector/NetCore.Tests/DependencyCollector.NetCore.Tests.csproj
```diff
@@ -15,28 +15,18 @@
...
-    <Compile Include="..\Shared.Tests\Implementation\DependencyCollectorDiagnosticListenerTests.Netstandard20.cs" Link="DependencyCollectorDiagnosticListenerTests.Netstandard20.cs" />
-    <Compile Include="..\Shared.Tests\Implementation\TelemetryDiagnosticSourceListenerTests.cs" Link="TelemetryDiagnosticSourceListenerTests.cs" />
-    <Compile Include="..\Shared.Tests\Implementation\SqlClientDiagnosticSourceListenerTests.cs" Link="SqlClientDiagnosticSourceListenerTests.cs" />
-    <Compile Include="..\Shared.Tests\Implementation\ServiceBusDiagnosticListenerTests.cs" Link="ServiceBusDiagnosticListenerTests.cs" />
-    <Compile Include="..\Shared.Tests\Implementation\EnumerableAssert.cs" Link="EnumerableAssert.cs" />
-    <Compile Include="..\Shared.Tests\Implementation\MockCorrelationIdLookupHelper.cs" Link="MockCorrelationIdLookupHelper.cs" />
-    <Compile Include="..\..\TestFramework\Shared\SdkVersionHelper.cs" Link="SdkVersionHelper.cs" />
-  </ItemGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <NoWarn>1701;1702;1705;1591</NoWarn>
```
d3r3kk: What do all of these warnings represent? Are they just style/trivial warnings that only affect the building of the test projects?

Src/DependencyCollector/Shared.Tests/DependencyCollector.Shared.Tests.projitems
```diff
@@ -10,33 +10,40 @@
...
-    <Compile Include="$(MSBuildThisFileDirectory)Implementation\EventHubsDiagnosticListenerTests.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)Implementation\ServiceBusDiagnosticListenerTests.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)Implementation\SqlClientDiagnosticSourceListenerTests.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Implementation\TelemetryDiagnosticSourceListenerTests.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)Implementation\WebRequestDependencyTrackingHelpersTests.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)Implementation\DependencyCollectorTestHelpers.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)Implementation\MockTelemetryInitializer.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)Implementation\TelemetryExtensionsForDependencyCollectorTests.cs" />
+  </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' != 'netcoreapp1.0' ">
```
d3r3kk: Interesting way to do this. What happens when we want to extend this grouping? Do we have to know to edit the .csproj file manually, or is this supported in the IDE?

If we are to manually edit the csproj I'd love to see something in the way of a README perhaps...

Src/TestFramework/Shared/AssemblyInfo.cs
```diff
@@ -2,16 +2,16 @@
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
-[assembly: AssemblyCompany("Microsoft")]
-[assembly: AssemblyCopyright("Copyright © Microsoft. All Rights Reserved.")]
-
-[assembly: ComVisible(false)]
+#pragma warning disable CS0436
```
d3r3kk: is there a matching 
`#pragma warning enable`
...at the bottom of this file?

MS-TimothyMothra: added, thanks

Src/DependencyCollector/Shared.Tests/DependencyCollector.Shared.Tests.projitems
MS-TimothyMothra: as far as i'm aware, this is the standard on all of our projects.
and it's MUCH simpler to do this kind of filtering in one place, vs individually across several other projects.

you're right, a note in the README would bebeneficial. :)

Src/DependencyCollector/NetCore.Tests/DependencyCollector.NetCore.Tests.csproj
MS-TimothyMothra: the first three were imported, i'm assume they came from a common project file, but i haven't searched that out yet. i added the forth (detailed above)

Src/DependencyCollector/Shared.Tests/DependencyCollector.Shared.Tests.projitems
d3r3kk: Ok, that's good for me to know. I'll be on the lookout for that going forward, and if I can think of some better solution I will "put up or shut up"! No need to block you on this item.

Issue #802, opened 2018-01-29
Populate Access-Control-Expose-Headers for cross component correlation 


iusafaro:
Populate Access-Control-Expose-Headers with Request-Context value to allow read value of Request-Context header from JavaScript SDK and use it for target in AJAX dependencies allowing cross-component correlation between AJAX dependencies and server-side requests.

https://www.w3.org/TR/cors/#http-access-control-expose-headers

MS-TimothyMothra:
One thing to note, this change is appending a new Value to the Header Key:


![image](https://user-images.githubusercontent.com/28785781/39387021-2d6e1b8e-4a2c-11e8-8859-2fbbafa632aa.png)


I don't know if this is intended. The HTTP standard might prefer comma delimited values:


> Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma. The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

PR #803, opened 2018-01-29
Add Access-Control-Expose-Headers and remove roleName from correlation

iusafaro:
Fix issue #802.

Populate Access-Control-Expose-Headers with Request-Context value to allow read value of Request-Context header from JavaScript SDK and use it for target in AJAX dependencies allowing cross-component correlation between AJAX dependencies and server-side requests.

https://www.w3.org/TR/cors/#http-access-control-expose-headers


cijothomas:
@iusafaro Do we need to do the same for aspnet core sdk also?
Also mention in Changelog that we are removing rolename fields.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -283,6 +257,10 @@ public void AddTargetHashForResponseHeader(HttpContext context)
                     if (this.correlationIdLookupHelper.TryGetXComponentCorrelationId(requestTelemetry.Context.InstrumentationKey, out correlationId))
                     {
                         context.Response.Headers.SetNameValueHeaderValue(RequestResponseHeaders.RequestContextHeader, RequestResponseHeaders.RequestContextCorrelationTargetKey, correlationId);
+
+                        // set additional header that allows to read this Request-Context from Javascript SDK
+                        // append this header with additional value to the potential ones defined by customer and they will be concatenated on cliend-side
+                        context.Response.AppendHeader(RequestResponseHeaders.AccessControlExposeHeadersHeader, RequestResponseHeaders.RequestContextHeader);
```
SergeyKanzhelev: make a config setting for this so it can be turned off if needed, please. I can see customers want to hide this information from exposure to third party web sites

iusafaro: Makes sense. Added EnableAccessControlExposeHeader - are you ok with the name?

PR #825, opened 2018-02-20
QuickPulse func test stabilization work.

tokaplan:
QuickPulse functional tests stabilization fixes.

- [x ] I ran Unit Tests locally.


cijothomas:
https://mseng.visualstudio.com/AppInsights/_build/index?buildId=5281377&_a=summary&tab=ms.vss-test-web.test-result-details test failure for QP

tokaplan:
Don't let test failures block this PR, failures will happen every once in a while, we must identify their frequency and gradually zero in on the causes (it appears that there are many).

Issue #829, opened 2018-02-28
Current version of StyleCop in the project (4.7.54) depends on .NET Microsoft.Build.Utilities.v3.5


d3r3kk:
If you are reporting bug/issue, please provide detailed Repro instructions.

## Repro Steps
1. Create a new VM image with Windows 10
2. Install all prerequisite software (but not .NET3.5)
3. Clone our repo and checkout the develop branch
4. Build the project on the command line with the ./buildRelease.cmd script

##  Actual Behavior
Build fails an error message like so:
`..\packages\StyleCop.MSBuild.4.7.54.0\build\StyleCop.MSBuild.Targets(101,5): error MSB4062:`
`The "StyleCopTask" task could not be loaded from the assembly ..\packages\StyleCop.MSBuild.4.7.54.0\build\.`
`.\tools\StyleCop.dll. Could not load file or assembly 'Microsoft.Build.Utilities.v3.5, Version=3.5.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies.`
`The system cannot find the file specified.`

## Expected Behavior
Project should build without error with all dependencies accounted for

## Version Info 
SDK Version : 2.6.0-beta
.NET Version : 4.5
How Application was onboarded with SDK(VisualStudio/StatusMonitor/Azure Extension) : na
OS : Windows
Hosting Info (IIS/Azure WebApps/ etc) : na

## Proposed solution
Upgrade current StyleCop package to v5.0.0 to remove the dependency.

d3r3kk:
Note that I have a local change that solves this issue and will submit it once beta1 is locked down. I can submit it before that if the team wishes.

MS-TimothyMothra:
I have no problem with merging this in now! :)
If we decide to wait for Beta2, I would vote that we skip this and migrate straight to the analyzer.

d3r3kk:
Let's put this in now to solve an immediate problem, and add an issue to move to the analyzers then.

MS-TimothyMothra:
I'm closing this in favor of tracking the new issue.
Feel free to reopen if this is still needed.

PR #830, opened 2018-02-28
Remove dependency on .NET v3.5, solves issue #829

d3r3kk:
Fix Issue #829 .
Update stylecop to v5. Update one project (TestConstants.csproj) to .NET 4.5.2 from 2.0.

- not a significant update


SergeyKanzhelev:
I'm OK with this as temporary solution. Longer term - please migrate to StyleCop analyzers for better VS experience. See https://github.com/Microsoft/ApplicationInsights-dotnet/blob/7321d2bf7b9c4733fb43b6c186e381863931eb1a/src/Microsoft.ApplicationInsights/Microsoft.ApplicationInsights.csproj#L49

MS-TimothyMothra:
I have no problem with merging this in now! :)
If we decide to wait for Beta2, I would vote that we skip this and migrate straight to the analyzer.

d3r3kk:
I'll merge this shortly, thanks folks! See issue #831, slated for 2.6-beta2

PR #840, opened 2018-03-05
Fix for build problem with base SDK 2.6.0-beta1

d3r3kk:
This PR solves issue with build, in preparation for the beta2 replacement release.


MS-TimothyMothra:
the build failed, looks like some formatting rules.
you should be able to resolve these by running the Release configuration locally.

Issue #854, opened 2018-03-14
Enable .NET Core in WindowsServer SDK


d3r3kk:
Enable collection of server information across platforms that is already present in .NET Full Framework

Specifically, we want to enable the collection of the Azure Instance Metadata Heartbeat properties in .NET Core scenarios (Linux/Mac).

## Intended Version Info 
SDK Version : 2.6.0-beta3
.NET Version : .NET Core

PR #855, opened 2018-03-14
Enable .NET Core in WindowsServer SDK

d3r3kk:
Complete heartbeat feature by enabling multi-platform scenarios in the WindowsServer SDK. *Note: There will be an upcoming rename to a more suitable one later.*

- Create new .NetCore and .NetCore.Tests projects in the WindowsServer package.
- Move windows specific classes into Net45 projects (where it is non-trivial to make the code/class .NET Core compatible)
  - *We should raise issues on non-Core classes and move them to .NET Core separately*
- Update code where possible to make it .NET Core compatible

       - No changes to Public API
- [x] Design discussion issue #854 


Src/WindowsServer/WindowsServer.NetCore/WindowsServer.NetCore.csproj
```diff
@@ -0,0 +1,50 @@
...
+  </Target>
+
+  <Import Project="..\WindowsServer.Shared\WindowsServer.Shared.projitems" Label="Shared" />
+
+  <Import Project="..\..\Common\Common.projitems" Label="Shared" />
+
+  <Import Project="..\..\..\..\packages\MicroBuild.Core.0.2.0\build\MicroBuild.Core.targets" Condition="Exists('..\..\..\..\packages\MicroBuild.Core.0.2.0\build\MicroBuild.Core.targets')" />
+
+  <PropertyGroup>
+    <Copyright>Copyright ? Microsoft. All Rights Reserved.</Copyright>
```
SergeyKanzhelev: is it not defined somewhere globally? In any case - deal with the question mark sign

Src/WindowsServer/WindowsServer.NetCore/WindowsServer.NetCore.csproj
```diff
@@ -0,0 +1,50 @@
...
+    <VersionPrefix>2.6.0-beta2</VersionPrefix>
```
SergeyKanzhelev: why is it not taken from the global props?

Src/WindowsServer/WindowsServer.NetCore/WindowsServer.NetCore.csproj
```diff
@@ -0,0 +1,50 @@
...
+    <DefineConstants>$(DefineConstants);NETCORE</DefineConstants>
+    <DelaySign>true</DelaySign>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <AssemblyName>Microsoft.AI.WindowsServer</AssemblyName>
+    <AssemblyOriginatorKeyFile>../../../Keys/35MSSharedLib1024.snk</AssemblyOriginatorKeyFile>
+    <SignAssembly>true</SignAssembly>
+    <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
+    <PackageId>Microsoft.AI.WindowsServer</PackageId>
+    <NetStandardImplicitPackageVersion>1.6.1</NetStandardImplicitPackageVersion>
+    <PackageTargetFallback>$(PackageTargetFallback);dnxcore50</PackageTargetFallback>
```
SergeyKanzhelev: why?

Src/WindowsServer/WindowsServer.NetCore/WindowsServer.NetCore.csproj
```diff
@@ -0,0 +1,50 @@
...
+  <PropertyGroup>
+    <Copyright>Copyright ? Microsoft. All Rights Reserved.</Copyright>
+    <VersionPrefix>2.6.0-beta2</VersionPrefix>
+    <Authors>Microsoft</Authors>
+    <TargetFramework>netstandard1.6</TargetFramework>
+    <DefineConstants>$(DefineConstants);NETCORE</DefineConstants>
+    <DelaySign>true</DelaySign>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <AssemblyName>Microsoft.AI.WindowsServer</AssemblyName>
+    <AssemblyOriginatorKeyFile>../../../Keys/35MSSharedLib1024.snk</AssemblyOriginatorKeyFile>
```
SergeyKanzhelev: Use global properties?

Src/WindowsServer/WindowsServer.NetCore/WindowsServer.NetCore.csproj
```diff
@@ -0,0 +1,50 @@
...
+    <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
+    <PackageId>Microsoft.AI.WindowsServer</PackageId>
+    <NetStandardImplicitPackageVersion>1.6.1</NetStandardImplicitPackageVersion>
+    <PackageTargetFallback>$(PackageTargetFallback);dnxcore50</PackageTargetFallback>
+    <GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>
+    <GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="..\..\Web\Web.Net45\Implementation\SdkVersionUtils.cs" Link="SdkVersionUtils.cs" />
```
SergeyKanzhelev: is it a standard practice in this repo?

Src/WindowsServer/WindowsServer.NetCore/WindowsServer.NetCore.csproj
```diff
@@ -0,0 +1,50 @@
...
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="MicroBuild.Core" Version="0.2.0" />
```
SergeyKanzhelev: privateasset

Src/WindowsServer/WindowsServer.NetCore/WindowsServer.NetCore.csproj
```diff
@@ -0,0 +1,50 @@
...
+    <PackageReference Include="Microsoft.ApplicationInsights" Version="2.6.0-beta2" />
+    <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="1.1.0" />
```
SergeyKanzhelev: what is this package for?

Src/WindowsServer/WindowsServer.NetCore.Tests/WindowsServer.NetCore.Tests.csproj
```diff
@@ -0,0 +1,58 @@
...
+
+  <ItemGroup>
+    <ProjectReference Include="..\WindowsServer.NetCore\WindowsServer.NetCore.csproj" />
+  </ItemGroup>
+
+  <Import Project="..\..\TestFramework\Shared\TestFramework.Shared.projitems" Label="Shared" />
+
+  <Import Project="..\WindowsServer.Shared.Tests\WindowsServer.Shared.Tests.projitems" Label="Shared" />
+
+  <!--<Import Project="..\WindowsServer.Shared.Tests\WindowsServer.Shared.Tests.projitems" Label="Shared" />
```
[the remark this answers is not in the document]
d3r3kk: Agreed.

Src/WindowsServer/WindowsServer.NetCore.Tests/WindowsServer.NetCore.Tests.csproj
```diff
@@ -0,0 +1,58 @@
...
+    <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
+    <RuntimeFrameworkVersion>1.0.4</RuntimeFrameworkVersion>
+    <PackageTargetFallback>$(PackageTargetFallback);dnxcore50</PackageTargetFallback>
+    <GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>
+    <GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
+    <DoNotSign>true</DoNotSign>
+    <RootNamespace>Microsoft.ApplicationInsights.Tests</RootNamespace>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
```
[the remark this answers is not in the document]
d3r3kk: This is a quirk of the Visual Studio 2017 system! Remove any `<NoWarn>` tags in your csproj, and then look at the UX properties of the project file - 1701, 1702, 1705, are automatically added to the 'Suppress Warnings' item. Add in an empty `<NoWarn>` tag, they are removed (and you get a bunch of warnings about System.Runtime 4.0.20.0 vs. 4.1.0.0). 

My intention here was *only* removing warnings in a test-only project (1591). However, 1701,1702, 1705, are all there  as well not added by me. I see that is the same for all NetCore projects in our repo. So, I am ultimately going to leave it this way.

> I spent *way too much time* investigating this. Eventually, [I stumbled across this conversation](https://github.com/aspnet/Mvc/issues/5050) and near the bottom the magical 1701 suppression is discussed. Apparently netcoreapp1.0 and netstandard1.3 are not using the same System.Runtime in our case (base SDK targets netstandard1.3,  web SDK targets netstandard 1.6/netcoreapp1.0).

Src/WindowsServer/WindowsServer.NetCore/WindowsServer.NetCore.csproj
d3r3kk: Not defined globally from what I can find. Perhaps adding these to the Common.props might be best. Let's push this to an issue for later (I fixed the question mark both here and in Perf.NetCore.csproj)

Src/WindowsServer/WindowsServer.NetCore/WindowsServer.NetCore.csproj
d3r3kk: There is no global props that contains this information. Instead, all  the nuget packages pull this from the respective assembly information. I've consolidated how this works across all the nuget projects in another PR #857.

Src/WindowsServer/WindowsServer.NetCore/WindowsServer.NetCore.csproj
d3r3kk: Copied this from one of the other NetCore projects, thanks for catching it.

I am unfamiliar with this setting, so I found [a discussion here](https://stackoverflow.com/questions/45569378/upgrading-to-net-core-2-0-packagetargetfallback-and-assettargetfallback-cannot) to figure out why this exists. From the discussion there, I believe we don't need it here at all.

Removed. Also, removed it from the Perf and DependencyCollection projects.

d3r3kk: [Also here dnxcore50 is deprecated now] (https://docs.microsoft.com/en-us/dotnet/standard/frameworks#deprecated-target-frameworks).

Src/WindowsServer/WindowsServer.NetCore/WindowsServer.NetCore.csproj
d3r3kk: Yes. This is a problem! 

The SdkVersionUtils method is used across all projects, and the method is defined in multiple places as well - the line above is a symptom of it and not the only one.

I've created another issue & PR specifically for this problem. Please check that out here #859.

Src/WindowsServer/WindowsServer.NetCore/WindowsServer.NetCore.csproj
d3r3kk: Done. I've added a new 'NetCore.props' file to consolidate some of the NetCore-specific settings, and as well I've added import of the already-existing Signing.props files to each *NetCore*.csproj file. I believe this improves the issue.

Src/WindowsServer/WindowsServer.NetCore/WindowsServer.NetCore.csproj
d3r3kk: Done.

Src/WindowsServer/WindowsServer.NetCore/WindowsServer.NetCore.csproj
```diff
@@ -0,0 +1,50 @@
...
+  <Import Project="..\..\Common\Common.projitems" Label="Shared" />
+
+  <Import Project="..\..\..\..\packages\MicroBuild.Core.0.2.0\build\MicroBuild.Core.targets" Condition="Exists('..\..\..\..\packages\MicroBuild.Core.0.2.0\build\MicroBuild.Core.targets')" />
+
+  <PropertyGroup>
+    <Copyright>Copyright ? Microsoft. All Rights Reserved.</Copyright>
+    <VersionPrefix>2.6.0-beta2</VersionPrefix>
+    <Authors>Microsoft</Authors>
+    <TargetFramework>netstandard1.6</TargetFramework>
+    <DefineConstants>$(DefineConstants);NETCORE</DefineConstants>
```
pharring: Careful. You're conflating a runtime (.NET Core) concept with a standard library (NETSTANDARD1_6) target framework. These things are not equivalent. A netstandard library can be consumed by many different runtimes (.NET Framework, Mono, CoreCLR). This may lead to confusion later on. Note that the project system automatically defines NETSTANDARD1_6 for you if the TargetFramework is netstandard1.6 and there are similar defines for other target frameworks. If you need a preprocessor constant that covers both NETSTANDARD1_3 and NETSTANDARD1_6, I would recommend NETSTANDARD.

Src/WindowsServer/WindowsServer.NetCore/WindowsServer.NetCore.csproj
MS-TimothyMothra: could this be a NET Core quirk?
It looks like <VersionPrefix> is only in NetCore projects.

Src/WindowsServer/WindowsServer.NetCore/WindowsServer.NetCore.csproj
pharring: Following on from this, I would rename the folder and project "WindowsServer.NetStandard16" instead of "WindowsServer.NetCore"

Probably not in this PR, but we should then try to consolidate all the targets (net40, net45, netstandard16) into a single .csproj with multiple target frameworks. Like we did in the base SDK.

Src/WindowsServer/WindowsServer.NetCore/WindowsServer.NetCore.csproj
d3r3kk: Not sure, this was also from copying data from the other NetCore projects. I've removed it and it seems to not have any effect.

Src/WindowsServer/WindowsServer.NetCore/WindowsServer.NetCore.csproj
d3r3kk: I see, thanks for calling this out. I imagine this is a result of my blind copying some of the other NetCore projects - test projects specifically that are actually netcore1.0 apps.

d3r3kk: Done. I will leave the multi-targeted project to later, but I agree we need to get that done as well. I opened issue #863.

d3r3kk: On the subject of renaming the projects to NetStandard instead of NetCore - I'll also leave that to the consolidation issue. NetCore was the naming convention for this type of work before I got here (correct or not) and I'm going to call fixing that out of scope for this PR.

d3r3kk:
@SergeyKanzhelev, @pharring, @MS-TimothyMothra 

I've completed addressing all the concerns raised thus far in this PR. One comment that is now hidden from @MS-TimothyMothra about the '?' in place of (c) in the Nuget Copyright field is addressed in a separate PR for nuget packaging that also needs to be reviewed.

Please let me know if you have further concerns, thanks!

SergeyKanzhelev:
I signed off to unblock the PR. Not reviewed carefully second time. @d3r3kk

Src/WindowsServer/WindowsServer.Shared/WindowsServer.Shared.projitems
```diff
@@ -12,28 +12,15 @@
...
-    <Compile Include="$(MSBuildThisFileDirectory)Implementation\RoleEnvironment.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)Implementation\RoleInstance.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)Implementation\RuntimeBindingObject.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)Implementation\ServiceRuntime.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Implementation\WindowsServerEventSource.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)Implementation\AzureServiceRuntimeAssemblyLoader.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Implementation\MetricManager.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)UnhandledExceptionTelemetryModule.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)FirstChanceExceptionStatisticsTelemetryModule.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)UnobservedExceptionTelemetryModule.cs" />
```
cijothomas: Can you tell why UnobservedExceptionTelemetryModule is included for netstandard whereas UnhandledExceptionTelemetryModule,FirstChanceExceptionStatisticsTelemetryModule are only for net45?

Src/WindowsServer/WindowsServer.Shared/WindowsServer.Shared.projitems
```diff
@@ -12,28 +12,15 @@
...
     <Compile Include="$(MSBuildThisFileDirectory)Implementation\IAzureMetadataRequestor.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)Implementation\IAzureRoleEnvironmentContextReader.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)Implementation\Role.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)Implementation\RoleEnvironment.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)Implementation\RoleInstance.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)Implementation\RuntimeBindingObject.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)Implementation\ServiceRuntime.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Implementation\WindowsServerEventSource.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)Implementation\AzureServiceRuntimeAssemblyLoader.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Implementation\MetricManager.cs" />
```
cijothomas: MetricManager is only used for FirstChanceExceptionStatisticsTelemetryModule - and since that module is not in netstandard, there is no need for MetricManager either in netstandard.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Version 2.6.0-beta3
 - [Ignore Deprecated events if running under netcore20](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/848)
 - [Implement unhandled exception auto-tracking (500 requests) for MVC 5 and WebAPI 2 applications.](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/847)
+- [Enable .NET Core platform in WindowsServer SDK, completing heartbeat feature across platforms](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/854)
```
cijothomas: Can you add more details to the issue? Specifically what Initializers/Modules from WindowsServer package are made available for .net core?

Src/WindowsServer/WindowsServer.Nuget/Package.nuspec
```diff
@@ -22,6 +22,12 @@
         <dependency id="Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel" version="[$coresdkversion$]" />
         <dependency id="Microsoft.ApplicationInsights.PerfCounterCollector" version="$version$" />
       </group>
+      <group targetFramework="netstandard1.6">
```
cijothomas: Please create a Github task to validate that the new assembly for netstandard is also signed. (We previously had an issue when DepCollector was added for netcore, the assembly was not signed)

Src/WindowsServer/WindowsServer.Shared.Tests/AzureInstanceMetadataEndToEndTests.cs
```diff
@@ -62,32 +65,30 @@ public void SpoofedResponseFromAzureIMSDoesntCrash()
...
-                jsonStream.WriteTo(context.Response.OutputStream);
-            }))
+                AzureInstanceMetadataEndToEndTests.MockTestUri,
+                testPath,
+                (response) =>
+                {
+                    response.StatusCode = (int)HttpStatusCode.OK;
+
+                    var jsonStream = this.GetTestMetadataStream();
+                    response.SetContentLength(3 * jsonStream.Length);
```
cijothomas: What is the limit above which we ignore the response? Why no make this (1+allowedlimit) ?

Src/WindowsServer/WindowsServer.Shared/WindowsServer.Shared.projitems
d3r3kk: This is purely because `UnobservedExceptionTelemetryModule` compiles & runs (tests) in both netstandard and net45, whereas `UnhandledExceptionTelemetryModule` and `FirstChanceExceptionStatisticsTelemetryModule` only work in net45 without adding effort.

If possible, I would like to keep this PR to *adding netstandard1.6* to the WindowsServer SDK, without removing any currently existing functionality in net45. 

I hope to address any issues about getting *all* functionality to work on both platforms in a future PR.

cijothomas: Ok. Please do mention either in changelog or in the linked GH issue, about which features from WindowsServer package is enabled for .netcore apps. We don't want users to think entire WindowsServer package features works for .net core.

Src/WindowsServer/WindowsServer.Shared/WindowsServer.Shared.projitems
d3r3kk: Agreed, done.

CHANGELOG.md
d3r3kk: Done, I enumerated all the modules that are now enabled in .NET Standard applications.

Src/WindowsServer/WindowsServer.Nuget/Package.nuspec
d3r3kk: Can I simply verify this myself, now? Or do we need  a task to do this?

Src/WindowsServer/WindowsServer.Shared.Tests/AzureInstanceMetadataEndToEndTests.cs
d3r3kk: I've added a test that puts us over the defined max length limit purposefully, instead of the 3x approach.

Src/WindowsServer/WindowsServer.Shared/WindowsServer.Shared.projitems
d3r3kk: Agreed, done.

PR #859, opened 2018-03-16
Remove duplicated methods across projects and share in Common shared project instead

d3r3kk:
Fix Issue #858 .
Consolidate duplicated methods across projects. Specifically, SdkVersionUtils.

      - no public API changes


runUnitTests.cmd
```diff
@@ -1,7 +1,7 @@
 @echo off
 @setlocal enabledelayedexpansion enableextensions
 
-CALL buildDebug.cmd
+@REM CALL buildDebug.cmd
```
MS-TimothyMothra: what was this for?

MS-TimothyMothra:
Thankyou!

runUnitTests.cmd
d3r3kk: Unintended, I will remove this.

PR #862, opened 2018-03-19
Add parameters to disable/enable strong name validation scripts

d3r3kk:
Script-only fix: Allow scripts to work on machines without a specific .NET windows SDK version.

The scripts previously assume .NET 4.6.2 is installed on a machine. Mine has 4.6.1, and 4.7. Instead of assuming this, use the environment to drive any assumptions and allow for them to be overridden as parameters.


d3r3kk:
I will improve this over the next day or two and re-open it.

PR #865, opened 2018-03-20
Addendum to #854

d3r3kk:
Missed a checkin from #854!
- Covers PR suggestions from @cijothomas from that PR


d3r3kk:
Note that these changes were already approved in #854!

Issue #868, opened 2018-03-26
Enabling Heartbeat in App Services for .NET and .NET Core


d3r3kk:
## Heartbeat feature for App Services
We would like the heartbeat feature to be extended to serve App Services instrumented with Application Insights.

## Heartbeat Content
Heartbeat data already includes some default properties, and for Azure App Services we'd like to add a few more default properties:
- asSiteName: The site name reported by Azure App Services.
- asStamp: The website "Home Stampname" reported by Azure App Services.
- asHost: The site host name as reported by Azure App Services.

These properties will help us to aid our customers discover the specific infrastructure their application is running under.

## Implementation Options
Thus far I've identified two options for implementing this change. In either case, the process will be to create a heartbeat `ITelemetryModule` that will get added to the AppInsights pipeline, acquire the default field values, and add them to the heartbeat upon initialization. 

*Option 1:*
- Create an `ITelemetryModule` in the Microsoft.ApplicationInsights.Web SDK (this repo)
- Create a similar `ITelemetryModule` in the Microsoft.ApplicationInsights.AspNetCore SDK

*Option 2:*
- Create an `ITelemetryModule` in the Microsoft.ApplicationInsights.Web SDK (this repo)
- Alter the Microsoft.ApplicationInsights.Web project to support .NET Standard 1.6
- Add a dependency on the Microsoft.ApplicationInsights.Web project in the Microsoft.ApplicationInsights.AspNetCore SDK

> **NOTE:** I am hoping to have this feature in for 2.6.0-beta3, so taking the approach of implementing Option 1, and then Option 2 is on the table here.

## Adding Heartbeats for App Services Automatically
ApplicationInsights.config.install.xdt would be employed to add the created `ITelemetryModule` to the ApplicationInsights.config file, in the case of .NET Full Framework applications. This is already available in the Microsoft.ApplicationInsights.Web nuget configuration and would be trivial.

For .NET Core applications, however, we would have to alter code in Microsoft.ApplicationInsights.AspNetCore to follow the pattern there today regarding AdaptiveSamplingTelemetryProcessor. That is, we would by default add the heartbeat `ITelemetryModule` but add a member to `ApplicationInsightsServiceOptions` that would allow customers to turn it off. (Until we have a way of configuring the application via config, I believe this is the only facility we have available).

SergeyKanzhelev:
[...]

PR #869, opened 2018-03-27
Add Heartbeat properties for App Services

d3r3kk:
Fix Issue #868.
Add heartbeat properties when App Services environment is detected. Keep App Services environment in sync using existing telemetry initializer used for same purpose.

   - No change to public API
- [x] Design discussion issue #868


d3r3kk:
@SergeyKanzhelev can you please have a look at this and let me know if this satisfies the topics we covered in issue #868? I believe I captured all the items discussed there, but I do still need to find a way to allow for updates in non-internal scenarios (AspNetCore).

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
```diff
@@ -0,0 +1,150 @@
...
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+    using Microsoft.ApplicationInsights.WindowsServer.Implementation;
+
+    /// <summary>
+    /// Provides default values for the heartbeat feature of Application Insights that
+    /// are specific to Azure App Services (Web Apps, Functions, etc...).
+    /// </summary>
+    internal class AppServicesHeartbeatTelemetryModule : ITelemetryModule, IDisposable
```
SergeyKanzhelev: must be public so it can be instantiated from code

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
```diff
@@ -0,0 +1,150 @@
...
+        };
+
+        private static AppServicesHeartbeatTelemetryModule instance;
+        private object lockObject = new object();
+        private bool isInitialized = false;
+
+        /// <summary>
+        /// Initializes a new instance of the<see cref="AppServicesHeartbeatTelemetryModule" /> class.
+        /// </summary>
+        internal AppServicesHeartbeatTelemetryModule()
```
SergeyKanzhelev: must be public so it can be instantiated from code

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
```diff
@@ -0,0 +1,150 @@
...
+        internal static readonly KeyValuePair<string, string>[] WebHeartbeatPropertyNameEnvVarMap = new KeyValuePair<string, string>[]
+        {
+            new KeyValuePair<string, string>("appSrv_SiteName", "WEBSITE_SITE_NAME"),
+            new KeyValuePair<string, string>("appSrv_wsStamp", "WEBSITE_HOME_STAMPNAME"),
+            new KeyValuePair<string, string>("appSrv_wsHost", "WEBSITE_HOSTNAME")
+        };
+
+        private static AppServicesHeartbeatTelemetryModule instance;
+        private object lockObject = new object();
+        private bool isInitialized = false;
```
SergeyKanzhelev: does this double initialization control matter for this module? There is no harm in initializing twice. Furthermore - this protection in module creates a false perception that nobody will call `Initialize` twice. In reality it may easily happen with re-creating of entire `TelemetryConfiguration`

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// Environment variables and the Application Insights heartbeat field names that accompany them.
+        /// </summary>
+        internal static readonly KeyValuePair<string, string>[] WebHeartbeatPropertyNameEnvVarMap = new KeyValuePair<string, string>[]
+        {
+            new KeyValuePair<string, string>("appSrv_SiteName", "WEBSITE_SITE_NAME"),
+            new KeyValuePair<string, string>("appSrv_wsStamp", "WEBSITE_HOME_STAMPNAME"),
+            new KeyValuePair<string, string>("appSrv_wsHost", "WEBSITE_HOSTNAME")
+        };
+
+        private static AppServicesHeartbeatTelemetryModule instance;
```
SergeyKanzhelev: why singleton?

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// <summary>
+        /// Initializes a new instance of the<see cref="AppServicesHeartbeatTelemetryModule" /> class.
+        /// </summary>
+        internal AppServicesHeartbeatTelemetryModule()
+        {
+            AppServicesHeartbeatTelemetryModule.Instance = this;
+        }
+
+        /// <summary>
+        /// Gets a value to provide internal access to the only instance of this class that *should* be available.
```
SergeyKanzhelev: I don't understand why it is a singleton. Can you provide more details

Src/WindowsServer/WindowsServer.Shared/Implementation/WindowsServerEventSource.cs
```diff
@@ -326,6 +326,29 @@ public void AzureInstanceMetadataFailureSettingDefaultPayload(string exception,
...
+            this.WriteEvent(
+                33,
+                envVarName ?? "unknown",
+                exceptionStr ?? "unknown-exception",
+                this.ApplicationName);
+        }
+
+        [Event(34,
+            Message = "Could not gain access to the Heartbeat Manager instance during initialization. Exception raised: {0}",
+            Level = EventLevel.Verbose)]
```
SergeyKanzhelev: Make it warning at least. This error is very unlikely to happen. It indicates that something is broken in SDK. So we want to have it at least at warning level.

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
```diff
@@ -0,0 +1,150 @@
...
+                {
+                    // get the variable, then expand it (otherwise we get the name we queried for in the value)
+                    string hbeatValue = Environment.GetEnvironmentVariable(kvp.Value);
+                    if (!string.IsNullOrEmpty(hbeatValue))
+                    {
+                        hbeatValue = Environment.ExpandEnvironmentVariables(hbeatValue);
+                        string hbeatKey = kvp.Key.ToString();
+                        if (isUpdateOperation)
+                        {
+                            hbeatManager.SetHeartbeatProperty(hbeatKey, hbeatValue);
```
SergeyKanzhelev: Would you expect `hbeatManager` to be `null` ever?

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
```diff
@@ -0,0 +1,150 @@
...
+                        hbeatManager = hman;
+                    }
+                }
+            }
+            catch (Exception hearbeatManagerAccessException)
+            {
+                WindowsServerEventSource.Log.HeartbeatManagerAccessFailure(hearbeatManagerAccessException.Message);
+            }
+
+            return hbeatManager;
```
SergeyKanzhelev: You do not check for `null` here. It may be a good idea though

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
```diff
@@ -0,0 +1,150 @@
...
+        {
+            var telemetryModules = TelemetryModules.Instance;
+            Dictionary<string, string> hbeatProps = new Dictionary<string, string>();
+
+            foreach (var kvp in WebHeartbeatPropertyNameEnvVarMap)
+            {
+                try
+                {
+                    // get the variable, then expand it (otherwise we get the name we queried for in the value)
+                    string hbeatValue = Environment.GetEnvironmentVariable(kvp.Value);
```
SergeyKanzhelev: this may throw security exception when it's not full trust appdomain. It is good idea to have a global `try{}catch` in the `Initialize`

SergeyKanzhelev:
I don't see how you are addressing synchronization with telemetry initializer and slot swap issue

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
d3r3kk: To allow access to the 'Update' method. I don't have a better mechanism at this moment, but I think we can define a better one given some time.

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
d3r3kk: I don't need to protect it, no. I do need to know when it gets re-initialized but that's all.

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
d3r3kk: Done. Totally missed this one. Thanks.

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
d3r3kk: We need a way to allow for the update of the heartbeat values associated to AppServices. I made this a singleton to allow other modules to signal this module to update the values. Perhaps there is a better mechanism for this, but I am not aware of one at this time. (I have thoughts on how to improve it in the future though, but it might take some time).

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
d3r3kk: Agreed. Done.

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
d3r3kk: Agreed. Done.

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
d3r3kk: Yes, I've added an early exit to counter this possibility.

Src/WindowsServer/WindowsServer.Shared/Implementation/WindowsServerEventSource.cs
d3r3kk: Agreed. I've gone over the other calls and altered their levels (including some I had set previously for the Azure IMS data as well).

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
d3r3kk: Done

Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
```diff
@@ -31,21 +31,35 @@ public AzureWebAppRoleEnvironmentTelemetryInitializer()
...
+
+            // ensure heartbeat values are up to date...
+            if (string.IsNullOrEmpty(this.lastNodeValue))
+            {
+                this.lastNodeValue = nodeName;
+            }
+            else if (!nodeName.Equals(this.lastNodeValue, StringComparison.Ordinal))
+            {
+                // if the AppServices heartbeat telemetry module exists, signal it to update the values in heartbeat
+                AppServicesHeartbeatTelemetryModule.Instance?.UpdateHeartbeatWithAppServiceEnvVarValues();
```
d3r3kk: @SergeyKanzhelev, here is where I signal the `AppServicesHeartbeatTelemetryModule` to update the values in the heartbeat properties. This is the reason I chose to go with the Singleton pattern. Perhaps there is a better way to go about this?

d3r3kk:
@SergeyKanzhelev ping?

d3r3kk:
@SergeyKanzhelev @cijothomas updated PR with tests, can be merged without any need for follow up PR.

- NOTE: I've added logic in this PR to update environment variables from the Initializer. That will work, but there is a danger that we would potentially miss updated environment variables (the value is potentially cached, please let me know if that is the case when using `InternalContext` as we do) and there might be perf issues with the way I propose here. Please see PR #870 for an alternate that I'm a bit happier with.

Src/WindowsServer/WindowsServer.Shared.Tests/AppServicesHeartbeatTests.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// <returns>Dictionary with expected environment variable names as the key, current environment variable content as the value.</returns>
+        private Dictionary<string,string> GetEnvVarsAssociatedToModule()
+        {
+            Dictionary<string, string> envVars = new Dictionary<string, string>();
+            foreach (var kvp in AppServicesHeartbeatTelemetryModule.WebHeartbeatPropertyNameEnvVarMap)
+            {
+                envVars.Add(kvp.Value, Environment.GetEnvironmentVariable(kvp.Value));
+                if (string.IsNullOrEmpty(envVars[kvp.Value]))
+                {
+                    Environment.SetEnvironmentVariable(kvp.Value, kvp.Key);
```
cijothomas: The effect is setting the ENV variable persists even after the test? This may have unintended consequences in the next test. 
If modifying any ENV, make sure to set it back to original after the test

Src/WindowsServer/WindowsServer.Shared.Tests/WindowsServer.Shared.Tests.projitems
```diff
@@ -9,6 +9,7 @@
     <Import_RootNamespace>Microsoft.ApplicationInsights.WindowsServer</Import_RootNamespace>
   </PropertyGroup>
   <ItemGroup>
+    <Compile Include="$(MSBuildThisFileDirectory)AppServicesHeartbeatTests.cs" />
```
cijothomas: AppServicesHeartbeatTelemetryModuleTests might be better name.

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
```diff
@@ -0,0 +1,181 @@
...
+            {
+                WindowsServerEventSource.Log.AppServiceHeartbeatSetCalledWithNullManager();
+            }
+            else
+            {
+                foreach (var kvp in WebHeartbeatPropertyNameEnvVarMap)
+                {
+                    try
+                    {
+                        // get the variable, then expand it (otherwise we get the name we queried for in the value)
```
cijothomas: I am not sure i understood this. Why not just read ENV variable like done everywhere else? 
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs#L65

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
```diff
@@ -0,0 +1,181 @@
...
+    /// are specific to Azure App Services (Web Apps, Functions, etc...).
+    /// </summary>
+    public sealed class AppServicesHeartbeatTelemetryModule : ITelemetryModule, IDisposable
+    {
+        /// <summary>
+        /// Environment variables and the Application Insights heartbeat field names that accompany them.
+        /// </summary>
+        internal static readonly KeyValuePair<string, string>[] WebHeartbeatPropertyNameEnvVarMap = new KeyValuePair<string, string>[]
+        {
+            new KeyValuePair<string, string>("appSrv_SiteName", "WEBSITE_SITE_NAME"),
```
cijothomas: Aren't these values appended to all telemetry items by the corresponding TelemetryInitializers? Why do we send these values (duplicated) into HeartBeat again?

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
```diff
@@ -0,0 +1,181 @@
...
+            // ensure our Instance variable is not kept around.
+            AppServicesHeartbeatTelemetryModule.instance = null;
+        }
+
+        /// <summary>
+        /// Signal the AppServicesHeartbeatTelemetryModule to update the values of the 
+        /// Environment variables we use in our heartbeat payload.
+        /// </summary>
+        /// <returns>A value indicating whether or not an update to the environment variables occurred.</returns>
+        public bool UpdateHeartbeatWithAppServiceEnvVarValues()
```
cijothomas: Instead of a TelemetryInitializers telling this module to update, can this module update itself (re-read ENV vars) every 'n' minute? And maybe inform the TelemetryInitializer that ENV values changed so that it'll also pick new values.

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
```diff
@@ -0,0 +1,181 @@
...
+
+        /// <summary>
+        /// Initialize the default heartbeat provider for Azure App Services. This module
+        /// looks for specific environment variables and sets them into the heartbeat 
+        /// properties for Application Insights, if they exist.
+        /// </summary>
+        /// <param name="configuration">Unused parameter.</param>
+        public void Initialize(TelemetryConfiguration configuration)
+        {
+            this.isInitialized = this.UpdateHeartbeatWithAppServiceEnvVarValues();
```
cijothomas: this.isInitialized -- Where is this used?

cijothomas:
Need to ensure the tests are run in CI builds. I searched for newly added tests but couldn't find them in test results for this CI build.

Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
```diff
@@ -31,21 +31,35 @@ public AzureWebAppRoleEnvironmentTelemetryInitializer()
...
+                nodeName = LazyInitializer.EnsureInitialized(ref this.nodeName, this.GetNodeName);
+                telemetry.Context.GetInternalContext().NodeName = nodeName;
+            }
+
+            // ensure heartbeat values are up to date...
+            if (string.IsNullOrEmpty(this.lastNodeValue))
+            {
+                this.lastNodeValue = nodeName;
+            }
+            else if (!nodeName.Equals(this.lastNodeValue, StringComparison.Ordinal))
```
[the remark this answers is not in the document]
d3r3kk: Agreed. Please see PR #870 where I fix this issue you call out here, and where I am adding some protection for performance as well. (My intention is to merge that PR into this one, as I prefer that solution).

Src/WindowsServer/WindowsServer.Shared.Tests/AppServicesHeartbeatTests.cs
d3r3kk: Agreed. Also, in PR #870 see that I've added a ClassInitializer/ClassCleanup method to handle environment changes.

Src/WindowsServer/WindowsServer.Shared.Tests/WindowsServer.Shared.Tests.projitems
d3r3kk: Ok.

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
d3r3kk: Agreed. There doesn't ever seem to be any need to handle environment variable expansion for this value. PR #870 does not attempt expansion.

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
d3r3kk: Heartbeat and telemetry initializers are two separate solutions. I am not 100% aware of all the downstream consumers of the TelemetryInitializers, but I am keenly aware of the heartbeat ones (app 2 infra). I want the app 2 infra feature to only look in 1 single place for the values it requires. I believe the duplication caused by heartbeats (default 15-minute cadence) is minimal enough to support the end purpose.

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
d3r3kk: Agreed. I was thinking that a subscriber model would work really nicely actually - that way the heartbeat module and the initializer module could simply 'subscribe' to the environment variable monitor (see PR #870, that's the terminology I use there) and get updated only as necessary. 

For this PR, I wanted to keep it as concise as possible to get the job done. I would certainly be open to creating such a subscriber in our next sprint.

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
d3r3kk: Line 94 - in the call to AddAppServiceEnvironmentVariablesToHeartbeat. The first time (isInitialized=false) we use AddHeartbeatProperty, each subsequent time we use SetHeartbeatProperty.

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
Dmitry-Matveev: `ExpandEnvironmentVariable` will make this method slower, let's skip if unnecessary.

Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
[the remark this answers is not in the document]
d3r3kk: Merged #870 here. Fixed.

Src/WindowsServer/WindowsServer.Shared.Tests/AppServicesHeartbeatTests.cs
d3r3kk: Merged #870 here where I fixed this issue by using unique env vars in place of the standard during testing. Note that I left the original testing for the TelemetryInitializer intact (although it really should be updated to do the same).

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
d3r3kk: After discussions with Dmitry I will create an alternate PR to this one that contains a subscriber model solution. Although due to time constraints we still may opt for this solution.

d3r3kk:
@cijothomas 
>Need to ensure the tests are run in CI builds. I searched for newly added tests but couldn't find them in test results for this CI build.

There was some iteration on builds as I added the tests, perhaps some of them got added as 'New Tests' in prior builds from the one you checked (I looked, and in the .1 build for that day there was 1 newly added test).

**HOWEVER**

Tests in the `WindowsServer.Test.Shared` project where not being added into the `WindowsServer.Net45.Tests` project *at all*. (`WindowsServer.NetCore.Tests` did load them, so they were still covered, just not fully). I have corrected this oversight in this PR as well.

d3r3kk:
This PR is replaced by PR #873.

PR #870, opened 2018-03-30
Add monitor to keep track of current value of App Services env vars

d3r3kk:
Corrects aspects of issue #868, further to PR #869.

Use to keep the value of WEBSITE_SITE_NAME and other App Services values current during runtime, as these values can change when swapping App Service applications from one slot to the next.

Initial PR to bring discussion on the fix, and get some feedback before pushing to Heartbeat update as well.

### DIFF
--- a/Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
+++ b/Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
@@ -1,8 +1,6 @@
 ﻿namespace Microsoft.ApplicationInsights.WindowsServer
 {
     using System;
-    using System.Threading;
-
     using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.Extensibility;
     using Microsoft.ApplicationInsights.Extensibility.Implementation;
@@ -18,9 +16,6 @@
 
         /// <summary>Predefined suffix for Azure Web App Hostname.</summary>
         private const string WebAppSuffix = ".azurewebsites.net";
-
-        private string nodeName;
-        private string roleName;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="AzureWebAppRoleEnvironmentTelemetryInitializer" /> class.
@@ -38,8 +33,7 @@
         {
             if (string.IsNullOrEmpty(telemetry.Context.Cloud.RoleName))
             {
-                string name = LazyInitializer.EnsureInitialized(ref this.roleName, this.GetRoleName);
-                telemetry.Context.Cloud.RoleName = name;
+                telemetry.Context.Cloud.RoleName = this.GetRoleName();
             }
 
             if (string.IsNullOrEmpty(telemetry.Context.GetInternalContext().NodeName))
@@ -61,8 +55,9 @@
 
         private string GetNodeName()
         {
-            AppServiceEnvVarMonitor.GetUpdatedEnvironmentVariable(WebAppHostNameEnvironmentVariable, ref this.nodeName);
-            return this.nodeName;
+            string nodeName = string.Empty;
+            AppServiceEnvVarMonitor.GetUpdatedEnvironmentVariable(WebAppHostNameEnvironmentVariable, ref nodeName);
+            return nodeName;
         }
     }
 }
--- a/Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
+++ b/Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
@@ -44,8 +44,7 @@
 
             if (string.IsNullOrEmpty(telemetry.Context.GetInternalContext().NodeName))
             {
-                string name = LazyInitializer.EnsureInitialized(ref this.nodeName, this.GetNodeName);                
-                telemetry.Context.GetInternalContext().NodeName = name;
+                telemetry.Context.GetInternalContext().NodeName = this.GetNodeName();
             }
         }
 
@@ -62,7 +61,8 @@
 
         private string GetNodeName()
         {
-            return Environment.GetEnvironmentVariable(WebAppHostNameEnvironmentVariable) ?? string.Empty;
+            AppServiceEnvVarMonitor.GetUpdatedEnvironmentVariable(WebAppHostNameEnvironmentVariable, ref this.nodeName);
+            return this.nodeName;
         }
     }
 }
--- a/Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
+++ b/Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
@@ -11,7 +11,7 @@
     internal static class AppServiceEnvVarMonitor
     {
         // Environment variables tracked by this monitor. (internal to allow tests to modify them)
-        internal static Dictionary<string, string> CheckedValues = new Dictionary<string, string>()
+        internal static readonly Dictionary<string, string> CheckedValues = new Dictionary<string, string>()
         {
             { "WEBSITE_SITE_NAME", string.Empty },
             { "WEBSITE_HOME_STAMPNAME", string.Empty },
--- a/Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
+++ b/Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
@@ -1,0 +1,57 @@
+﻿namespace Microsoft.ApplicationInsights.WindowsServer.Implementation
+{
+    using System;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Utility to monitor the value of environment variables which may change 
+    /// during the run of an application. Checks the environment variables 
+    /// intermittently.
+    /// </summary>
+    internal static class AppServiceEnvVarMonitor
+    {
+        internal static TimeSpan CheckInterval = TimeSpan.FromSeconds(30);
+        internal static DateTime LastCheckTime = DateTime.MinValue;
+
+        /// <summary>
+        /// Environment variables tracked by this monitor.
+        /// </summary>
+        internal static Dictionary<string, string> CheckedValues = new Dictionary<string, string>()
+        {
+            { "WEBSITE_SITE_NAME", string.Empty },
+            { "WEBSITE_HOME_STAMPNAME", string.Empty },
+            {  "WEBSITE_HOSTNAME", string.Empty }
+        };
+
+        /// <summary>
+        /// Get the latest value assigned to an environment variable.
+        /// </summary>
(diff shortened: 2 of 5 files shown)


Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
```diff
@@ -0,0 +1,57 @@
...
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Utility to monitor the value of environment variables which may change 
+    /// during the run of an application. Checks the environment variables 
+    /// intermittently.
+    /// </summary>
+    internal static class AppServiceEnvVarMonitor
+    {
+        internal static TimeSpan CheckInterval = TimeSpan.FromSeconds(30);
```
pharring: Nit: `readonly`

Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
```diff
@@ -0,0 +1,57 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Simply update the stored environment variables if the last time we 
+        /// checked from now is greater than the check interval.
+        /// </summary>
+        private static void CheckVariablesIntermittent()
+        {
+            DateTime checkTime = DateTime.Now;
```
pharring: Always use `DateTime.UtcNow` to avoid an expensive local timezone lookup.

Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
```diff
@@ -0,0 +1,57 @@
...
+            if (checkTime - LastCheckTime > CheckInterval)
```
pharring: For bonus perf points, instead of `LastCheckTime` compute and store `NextCheckTime` (i.e. the time after which you should check again) and avoid the subtraction in the common case and make it an addition in the rare case.
i.e.
```C#
DateTime now = DateTime.UtcNow;
if (now > NextCheckTime)
{
    NextCheckTime = now + CheckInterval;
   // etc.
}
```

SergeyKanzhelev:
@d3r3kk can you please confirm you actually tried it and this code works before I'll be reviewing it? From the fast glance this is NOT the fix that suppose to be working.

Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    /// <summary>
+    /// Utility to monitor the value of environment variables which may change 
+    /// during the run of an application. Checks the environment variables 
+    /// intermittently.
+    /// </summary>
+    internal static class AppServiceEnvVarMonitor
+    {
+        internal static TimeSpan CheckInterval = TimeSpan.FromSeconds(30);
+        internal static DateTime LastCheckTime = DateTime.MinValue;
```
pharring: Nit: `private` (unless you need to change it in unit tests.)

Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        /// <summary>
+        /// Environment variables tracked by this monitor.
+        /// </summary>
+        internal static Dictionary<string, string> CheckedValues = new Dictionary<string, string>()
```
pharring: Nit: `private static readonly` (unless you need to change it in unit tests.)

Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            { "WEBSITE_SITE_NAME", string.Empty },
+            { "WEBSITE_HOME_STAMPNAME", string.Empty },
+            {  "WEBSITE_HOSTNAME", string.Empty }
```
pharring: Nit: extra space (alignment)

d3r3kk:
Thanks for looking this over @pharring, exactly the feedback I was looking for (keeping perf in mind for this task is pretty important I think).

@SergeyKanzhelev I am looking for feedback if this is a decent idea to pursue or not. In principle it *should* work with the idea presented, but I understand if you need to keep your PR review time streamlined - apologies for the noise! I will ensure it works in a test now and ping you for feedback once I've proven it out further.

SergeyKanzhelev:
I gave a link to PR in Azure functions that fixing this issue. Problem there is that sandbox doesn't update every env variable after swap. So I'd like you to make sure it works

Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
d3r3kk: Done

Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
d3r3kk: Done - good tip, thanks.

Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
d3r3kk: Awesome. Done.

Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
d3r3kk: `CheckInterval` can be `private`, `NextCheckTime` internal to allow tests to work - added a line comment to call this intention out.

Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
d3r3kk: Agreed, done. I am always thrown by the readonly keyword with containers...

Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
d3r3kk: Done, this was breaking the build (stylecop).

d3r3kk:
@SergeyKanzhelev Please have a look at this now that I've got it a bit more put together. I've added some test methods to ensure they do what we want, and that the change satisfies the same demands the Web Functions PR did.

Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
```diff
@@ -0,0 +1,60 @@
...
+            { "WEBSITE_SITE_NAME", string.Empty },
+            { "WEBSITE_HOME_STAMPNAME", string.Empty },
+            { "WEBSITE_HOSTNAME", string.Empty }
+        };
+
+        // When is the next time we will allow a check to occur? (internal to allow tests to modify this to avoid waits)
+        internal static DateTime NextCheckTime = DateTime.MinValue;
+
+        // how often we allow the code to re-check the environment
+        private static readonly TimeSpan CheckInterval = TimeSpan.FromSeconds(30);
```
cijothomas: Let user configure it?

d3r3kk: Could do, not sure how much value would be there. I propose we raise an issue and chat about it for the next release.

Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
```diff
@@ -62,7 +55,9 @@ private string GetRoleName()
 
         private string GetNodeName()
         {
-            return Environment.GetEnvironmentVariable(WebAppHostNameEnvironmentVariable) ?? string.Empty;
+            string nodeName = string.Empty;
+            AppServiceEnvVarMonitor.GetUpdatedEnvironmentVariable(WebAppHostNameEnvironmentVariable, ref nodeName);
```
Dmitry-Matveev: This is no longer lazy initialized and we've added ~3 extra stack frames to every request/dependency in the customer app. Variable will be read synchronously every 30 seconds (may be fine if we do not `Expand..()` it, though).
We may think of an asynchronous approach here. Discussed offline with Derek.

Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// intermittently.
+    /// </summary>
+    internal static class AppServiceEnvVarMonitor
+    {
+        // Environment variables tracked by this monitor. (internal to allow tests to modify them)
+        internal static readonly Dictionary<string, string> CheckedValues = new Dictionary<string, string>()
+        {
+            { "WEBSITE_SITE_NAME", string.Empty },
+            { "WEBSITE_HOME_STAMPNAME", string.Empty },
+            { "WEBSITE_HOSTNAME", string.Empty }
```
Dmitry-Matveev: In comments for environment variable extraction in [this PR](https://github.com/Azure/azure-webjobs-sdk/pull/1382/files), it's noted that WEBSITE_HOSTNAME is not a reliable source:

> //...refactored so that it did not use WEBSITE_HOSTNAME, which is determined to be unreliable for functions during slot swaps. 

Do we want to rely on that field, or we may use some tricks, e.g. from the same PR to build something more reliable from what we have.

Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
Dmitry-Matveev: Let's not overexpose API we are not certain in and may want to pull in the very next PR. We can make it configurable later if we stick with this solution, we cannot unmake public API once it's in stable release without the breaking change.

Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
d3r3kk: I say we should resort to these tactics if we have an issue or have proof that App Services are affected by the same failing that Azure Functions are. I have put an issue together to verify that we can detect slot swap using only WEBSITE_HOSTNAME, and to update to using the WEBSITE_SITE_NAME + WEBSITE_SLOT_NAME if the need arises. Note that I've included the WEBSITE_SLOT_NAME in the list of environment variables that the monitor watches for now.

d3r3kk:
I have merged this change into PR #869. Please move comments and issues to that PR.

Issue #872, opened 2018-04-03
Verify that App Services slot swaps are being detected in Web SDK


d3r3kk:
There is some concern that slot swap detection is not reliable in the Web SDK for App Services. We have seen evidence that the slot swap was not being detected properly in Azure Functions, and we must ensure that other App Services (Web Apps, for instance) are not affected in the same way.

See comments in PR #870 regarding the use of WEBSITE_HOSTNAME for further insight & details.

## Repro Steps
1. Create test web application in Azure (use the New Devops Project workflow)
2. Add a deployment slot to the application's CD pipeline
3. Instrument the application with the latest beta Web SDK
4. Perform a slot swap and watch App Insights Analytics to ensure WEBSITE_HOSTNAME is being recorded properly.

##  Actual Behavior
**TO BE VERIFIED**

## Expected Behavior
The nodeName property of application insights telemetry will update to reflect the correct slot name within 30 seconds of the slot change.

## Version Info 
SDK Version : latest beta.
.NET Version :	4.5 (.NET full framework) or 1.3 (.NET Standard) 						
Hosting Info (IIS/Azure WebApps/ etc) : WebApps

d3r3kk:
See PR Azure/azure-webjobs-sdk#1382 for further detail.

MS-TimothyMothra:
@d3r3kk will you be able to test this next week?
Please follow up with @Dmitry-Matveev

d3r3kk:
Will investigate by EOD tomorrow.

d3r3kk:
# Environment values investigation

*Can we use the environment to detect slot-swaps in Application Insights?*

**tl;dr** No, we cannot. The environment variables on the host are not propagated to the process after a slot swap.

## Findings at Runtime

### What does the environment say in a demo app?

The environment variables are not updating how we need them to, in order to detect slot - swaps.
I wrote a demo ASP.NET app to show the current values of the environment variables before/after a slot swap.

Here's the variables in the 'dev slot' when I deploy my app into the DEV environment in my release definition in VSTS:

**Environment Variable Name** | **Value**
------------------------------- | ----------------------------
WEBSITE_SITE_NAME | dek-app2infra
WEBSITE_HOME_STAMPNAME | waws-prod-sn1-115
WEBSITE_HOSTNAME | **dek-app2infra-dev.azurewebsites.net**
[...]

PR #873, opened 2018-04-03
Add App Services Heartbeat with environment variable monitor

d3r3kk:
Fix Issue #868 and #872.

Add heartbeat properties when App Services environment is detected. Keep App Services-specific environment variables in sync across slot-swaps (and anything else that would update environment variables).

       - No change to public API
- [x] Design discussion issue #868, #872


d3r3kk:
This is option B, option A is PR #870, and we can go with either PR for 2.6.0-beta3.
Please review this code and we can discuss benefits of this subscriber vs. the prior polling solution.
Note that I am still working on getting the unit tests constructed and passing at the time I wrote this comment.

Src/WindowsServer/WindowsServer.Shared/Implementation/EnvironmentVariableMonitor.cs
```diff
@@ -0,0 +1,111 @@
...
+        // how often we allow the code to re-check the environment
+        protected TimeSpan checkInterval;
+
+        // timer object that will periodically update the environment variables
+        private readonly Timer environmentCheckTimer;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EnvironmentVariableMonitor" /> class.
+        /// </summary>
+        public EnvironmentVariableMonitor(IEnumerable<string> envVars, TimeSpan checkInterval)
```
pharring: Is there any advantage to pass the initial `envVars` to this constructor?
I mean, the monitor can build up its own list of monitored environment variables via the "GetOrAdd" within GetCurrentEnvironmentVariableValue.
So, why not simplify it and only monitor those variables that are actually queried via the public API.

pharring: Constructor of an abstract class should be protected

pharring: Oh, I guess the difference is in the "OnEnvironmentVariableUpdated" call-back. You can get a call-back even for variables you didn't actually call GetCurrentEnvironmentVariableValue on.

d3r3kk: Yes, I am optimizing for the Azure App Services scenario here.

d3r3kk: Yes of course, thanks for the catch.

Src/WindowsServer/WindowsServer.Shared/Implementation/EnvironmentVariableMonitor.cs
```diff
@@ -0,0 +1,111 @@
...
+        {
+            this.CheckedValues = new ConcurrentDictionary<string, string>();
+            this.checkInterval = checkInterval;
+
+            foreach (string varName in envVars)
+            {
+                this.CheckedValues.TryAdd(varName, Environment.GetEnvironmentVariable(varName));
+            }
+
+            this.environmentCheckTimer = new Timer(this.CheckVariablesIntermittent, null, this.checkInterval, TimeSpan.FromMilliseconds(-1));
```
Dmitry-Matveev: If `this.checkInterval` can ever be 0 (i hope not), then there is a race condition possible - `this.CheckVariablesIntermittent` can be executed in its entirety before the assignment above happens, so `this.environmentCheckTimer()` will still be null. The last thing `this.CheckVariablesIntermittent()` does is to change `this.environmentCheckTimer` that may cause NullReference.
If I remember correctly, the right pattern was to create `Timer` like `Timer(callback, null, Timeout.Infinite, Timeout.Infinite)` and immediately follow up with `timer.Change(this.checkInterval, TimeoutInfinite)`.
Again, I think this only applies if you may have really-really low value as `this.checkInterval`.

Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
```diff
@@ -62,7 +83,14 @@ private string GetRoleName()
 
         private string GetNodeName()
         {
-            return Environment.GetEnvironmentVariable(WebAppHostNameEnvironmentVariable) ?? string.Empty;
+            string nodeName = string.Empty;
+            AppServiceEnvVarMonitor.Instance.GetCurrentEnvironmentVariableValue(this.WebAppHostNameEnvironmentVariable, ref nodeName);
```
Dmitry-Matveev: Is this the quickest read possible out of ConcurrentDictionary (I see that internally it uses `GetOrAdd()`) - is there anything that reads faster than that for ConsurrentDictionary with no locks whatsoever?
We are OK if several items are still having old data, but having all ongoing requests/dependencies and other telemetry items to sync on read of that dictionary may lead to a bottleneck here.

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
```diff
@@ -0,0 +1,161 @@
...
+
+        /// <summary>
+        /// Signal the AppServicesHeartbeatTelemetryModule to update the values of the 
+        /// Environment variables we use in our heartbeat payload.
+        /// </summary>
+        public void UpdateHeartbeatWithAppServiceEnvVarValues()
+        {
+            try
+            {
+                var hbeatManager = this.GetHeartbeatPropertyManager();
```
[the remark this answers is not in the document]
d3r3kk: Right now, we are pretty much guaranteed that the HeartbeatPropertyManager is present (since it is the `DiagnosticsTelemetryModule` itself). And for the lifetime of the app it will either be there or it won't, no change. I think I will change this to cache it.

Src/WindowsServer/WindowsServer.Shared/Implementation/EnvironmentVariableMonitor.cs
d3r3kk: Ok, I wasn't aware of that possibility. I will make the change to a 2-step initialization here.

Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
d3r3kk: Good catch to call this out. However, I think I can put any fears to rest. 

Notice that by the time this call is being made, the cached value from the environment has already been stored for this update cycle, the key is already in place. This all works together to ensure the `GetOrAdd` call is simply a `TryGetValue` call ([that has no lock on it as you can see here](https://github.com/dotnet/corefx/blob/e0802d8b6010d4bb4538b15f228eee45378c4cc4/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs#L496)). 

However, in the scenario you suggest (in some future class that we decide to tax the `EnvironmentVariableMonitor` class as much as possible), it absolutely could be possible to create a nasty bottleneck.

Even then, as far as I am aware, GetOrAdd is the *best, fastest* way to deal with updates of this nature. 

`ConcurrentDictionary` actually does the only thing I can think of that *might* be faster, [breaking the problem down into a `TryGetValue` and then using a call to `TryUpdateValue` *only when things have changed*.](https://github.com/dotnet/corefx/blob/e0802d8b6010d4bb4538b15f228eee45378c4cc4/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs#L1030)

In the scenario that you site though, where everything is all being updated at once, we still have a bottleneck (albeit one that is handled very nicely by `ConcurrentDictionary` without us having to write and maintain any complex code).

d3r3kk: I probably could have made this shorter:

> ... We are OK if several items are still having old data...

This is exactly the path that we follow today. 
1. Update all environment variables in a loop, detecting any change. 
2. If we detected a change, notify subscribers to collect their new cached values, ensuring only `TryGetValue` operations on *already tracked* items.

But thanks for making me read into how `ConcurrentDictionary` works under the hood. I do feel better about using it now. I was worried it was locking things even in the case of reads!

Src/WindowsServer/WindowsServer.Shared/Implementation/EnvironmentVariableMonitor.cs
d3r3kk: I think I will enforce a minimum check interval as well - 5 seconds.

Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
```diff
@@ -0,0 +1,59 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.ComponentModel;
+
+    internal delegate void MonitoredAppServiceEnvVarUpdated();
+
+    /// <summary>
+    /// Utility to monitor the value of environment variables which may change 
+    /// during the run of an application. Checks the environment variables 
+    /// intermittently.
```
cijothomas: intermittently .> at regular fixed intervals may sound better.

Src/WindowsServer/WindowsServer.Shared/Implementation/EnvironmentVariableMonitor.cs
```diff
@@ -0,0 +1,114 @@
...
+            }
+
+            if (shouldTriggerOnUpdate)
+            {
+                this.OnEnvironmentVariableUpdated();
+            }
+
+            if (this.isEnabled)
+            {
+                this.environmentCheckTimer.Change(this.checkInterval, TimeSpan.FromMilliseconds(-1));
```
SergeyKanzhelev: single exception will stop timer and there will be no notification to the customer. This will be hard to troubleshoot.

Src/WindowsServer/WindowsServer.Shared/Implementation/WindowsServerEventSource.cs
```diff
@@ -326,6 +326,71 @@ public void AzureInstanceMetadataFailureSettingDefaultPayload(string exception,
...
+        {
+            this.WriteEvent(
+                34,
+                exceptionStr ?? "unknown-exception",
+                this.ApplicationName);
+        }
+        
+        [Event(35,
+            Message = "App Services Heartbeat Provider: Accessing the Hearbeat Manager failed as it is not in the list of available modules.",
+            Level = EventLevel.Informational)]
```
cijothomas: This could be Warning just like the above is warning.

Src/WindowsServer/WindowsServer.Shared/Implementation/EnvironmentVariableMonitor.cs
```diff
@@ -0,0 +1,114 @@
...
+
+    /// <summary>
+    /// Utility to monitor the value of environment variables which may change 
+    /// during the run of an application. Checks the environment variables 
+    /// intermittently.
+    /// </summary>
+    internal abstract class EnvironmentVariableMonitor : IDisposable
+    {
+        // Environment variables tracked by this monitor.
+        protected readonly ConcurrentDictionary<string, string> CheckedValues;
```
SergeyKanzhelev: Why is it a concurrent collection? Do you plan to run multiple instances of timer callback at the same time?

Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
```diff
@@ -1,54 +1,75 @@
...
+        /// same across instances of same slot. Marked as internal to support testing.
+        /// </summary>
+        internal string WebAppHostNameEnvironmentVariable = "WEBSITE_HOSTNAME";
 
         /// <summary>Predefined suffix for Azure Web App Hostname.</summary>
         private const string WebAppSuffix = ".azurewebsites.net";
 
+        /// <summary>
+        /// Value used for keeping track of when the hostname changes (slot swaps occur). We use this
+        /// to notify the other class that makes use of these environment variables.
```
cijothomas: misplaced comment?

Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
```diff
@@ -1,54 +1,75 @@
...
         /// </summary>
         public AzureWebAppRoleEnvironmentTelemetryInitializer()
         {
             WindowsServerEventSource.Log.TelemetryInitializerLoaded(this.GetType().FullName);
+            AppServiceEnvVarMonitor.Instance.MonitoredAppServiceEnvVarUpdatedEvent += this.UpdateEnvironmentValues;
         }
 
         /// <summary>
-        /// Initializes <see cref="ITelemetry" /> device context.
+        /// Initializes <see cref="ITelemetry" /> device context and helps keep any heartbeat values in sync as well.
```
cijothomas: This initializer does not have any knowledge of heartbeat right? The comment look incorrect.

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
```diff
@@ -0,0 +1,158 @@
...
+    /// </summary>
+    public sealed class AppServicesHeartbeatTelemetryModule : ITelemetryModule, IDisposable
+    {
+        /// <summary>
+        /// Environment variables and the Application Insights heartbeat field names that accompany them.
+        /// </summary>
+        internal readonly KeyValuePair<string, string>[] WebHeartbeatPropertyNameEnvVarMap = new KeyValuePair<string, string>[]
+        {
+            new KeyValuePair<string, string>("appSrv_SiteName", "WEBSITE_SITE_NAME"),
+            new KeyValuePair<string, string>("appSrv_SiteName", "WEBSITE_SLOT_NAME"),
```
cijothomas: appSrv_SlotName?

Src/WindowsServer/WindowsServer.Shared.Tests/EnvironmentVariableMonitorTests.cs
```diff
@@ -0,0 +1,105 @@
...
+        [TestMethod]
+        public void EnsureInstanceWorksAsIntended()
+        {
+            Assert.NotNull(AppServiceEnvVarMonitor.Instance);
+        }
+
+        [TestMethod]
+        public void EnsureEnvironmentVariablesAreCapturedImmediately()
+        {
+            var envVars = GetCurrentAppServiceEnvironmentVariableValues();
```
cijothomas: No cleanup of ENV vars?

Src/WindowsServer/WindowsServer.Shared.Tests/AzureWebAppRoleEnvironmentTelemetryInitializerTest.cs
```diff
@@ -21,77 +21,98 @@ public void AzureWebAppRoleEnvironmentTelemetryInitializerSetsRoleName()
...
-            Environment.SetEnvironmentVariable("WEBSITE_HOSTNAME", null);
+            Environment.SetEnvironmentVariable(testVarName, null);
         }
 
         [TestMethod]
         public void AzureWebAppRoleEnvironmentTelemetryInitializerDoesNotOverrideNodeName()
         {
-            Environment.SetEnvironmentVariable("WEBSITE_HOSTNAME", "TestRoleName.azurewebsites.net");
-            
+            var testVarName = "WEBSITE_" + Guid.NewGuid().ToString() + "_HOSTNAME";
```
cijothomas: Why do we change the ENV vars?

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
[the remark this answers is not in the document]
d3r3kk: Done.

Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
d3r3kk: Maybe I was diving too deep.

In the 'happy path', we always read from the local variables:

>From AzureWebAppRoleEnvironmentTelemetryInitializer.cs, line 45

    public void Initialize(ITelemetry telemetry)
    {
        if (this.updateEnvVars) // <= this is set in the update callback when change is detected
        {
            this.roleName = this.GetRoleName();
            this.nodeName = this.GetNodeName();
            this.updateEnvVars = false; // <= this ensures we don't check again until another change
        }
    ...

*Only when a change in the environment is detected* do we ever go and get the information from the monitor.

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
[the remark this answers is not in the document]
d3r3kk: No, that was a mistake, thank you for catching this.

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
```diff
@@ -0,0 +1,158 @@
...
+        /// <summary>
+        /// Initialize the default heartbeat provider for Azure App Services. This module
+        /// looks for specific environment variables and sets them into the heartbeat 
+        /// properties for Application Insights, if they exist.
+        /// </summary>
+        /// <param name="configuration">Unused parameter.</param>
+        public void Initialize(TelemetryConfiguration configuration)
+        {
+            this.UpdateHeartbeatWithAppServiceEnvVarValues();
+            AppServiceEnvVarMonitor.Instance.MonitoredAppServiceEnvVarUpdatedEvent += this.UpdateHeartbeatWithAppServiceEnvVarValues;
```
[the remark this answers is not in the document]
d3r3kk: Agreed. I am going to make an assumption that the presence of one of these environment variables means we are running under Azure App Services.

Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvVarMonitor.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    internal sealed class AppServiceEnvVarMonitor : EnvironmentVariableMonitor
```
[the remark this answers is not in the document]
d3r3kk: Will update this now, before we go stable. Consistency is good.

Src/WindowsServer/WindowsServer.Shared/Implementation/EnvironmentVariableMonitor.cs
```diff
@@ -0,0 +1,114 @@
...
+        protected void CheckVariablesIntermittent(object state)
+        {
+            bool shouldTriggerOnUpdate = false;
+
+            var iter = this.CheckedValues.GetEnumerator();
+            while (iter.MoveNext())
+            {
+                var kvp = iter.Current;
+
+                string envValue = Environment.GetEnvironmentVariable(kvp.Key);
```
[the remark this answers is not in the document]
d3r3kk: Agreed. I capture the `SecurityException` explicitly and disable the `EnvironmentVariableMonitor` completely when it occurs. No need to continue generating exceptions/logs for this.

Src/WindowsServer/WindowsServer.Shared/Implementation/EnvironmentVariableMonitor.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// </summary>
+        protected abstract void OnEnvironmentVariableUpdated();
+
+        /// <summary>
+        /// Check and update the variables being tracked and if any updates are detected,
+        /// raise the OnEnvironmentVariableUpdated event. Restart the timer to check again
+        /// in the configured interval once complete.
+        /// </summary>
+        /// <param name="state">Variable left unused in this implementation of TimerCallback.</param>
+        protected void CheckVariablesIntermittent(object state)
```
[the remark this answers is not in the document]
d3r3kk: Global `try{}catch` added, I disable the monitor in the specific case a `SecurityException` is caught when getting the environment variable value, but I only log a message for any other exception type.

Src/WindowsServer/WindowsServer.Shared/Implementation/EnvironmentVariableMonitor.cs
d3r3kk: Logged messages to the `WindowsServerEventSource` that should help with this.

Src/WindowsServer/WindowsServer.Shared/Implementation/EnvironmentVariableMonitor.cs
d3r3kk: For this abstract class I set out to run it on a single timer callback, and I even guard specifically against re-entrancy with how I use the timer. 

However, because it is possible that some subscriber to the event might exist on different threads, I wanted to ensure those cases were covered.

Perhaps I was being over cautious here.

d3r3kk: @SergeyKanzhelev Would you like me to change this for this PR?

Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
d3r3kk: Yes, removed.

Src/WindowsServer/WindowsServer.Shared/Implementation/WindowsServerEventSource.cs
d3r3kk: All these messages are Warning I think.

Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
d3r3kk: Yes, thanks for catching this. Leftover from when I updated it last.

Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs
d3r3kk: Thanks, fixed.

Src/WindowsServer/WindowsServer.Shared.Tests/EnvironmentVariableMonitorTests.cs
d3r3kk: Thanks for the catch. Oversight on my part (did it in one place, not the other). Updated to use `[TestInitialize]` and `[TestCleanup]` here as well.

Src/WindowsServer/WindowsServer.Shared.Tests/AzureWebAppRoleEnvironmentTelemetryInitializerTest.cs
d3r3kk: We now have 3 classes that all use the *same environment variables* and testing each of them means we update/clear/change that small set of environment variables as we go. If we run the tests in parallel, tests will have that value changed unexpectedly during their execution by other tests.

d3r3kk:
Build failure exceeding time limit of 60 minutes...

Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvironmentVariableMonitor.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            "WEBSITE_SITE_NAME",
+            "WEBSITE_SLOT_NAME",
+            "WEBSITE_HOME_STAMPNAME",
+            "WEBSITE_HOSTNAME",
+            "WEBSITE_OWNER_NAME"
+        };
+
+        // singleton pattern, this is the one instance of this class allowed
+        private static readonly AppServiceEnvironmentVariableMonitor SingletonInstance = new AppServiceEnvironmentVariableMonitor();
```
cijothomas: Is it possible to avoid this static initialization? If the app is not running on WebApp, then the entire effort of this class ctr() and reading multiple env vars is wasted and leading to increasing application startup time. This maybe very minimal but .net core apps could be running in containers and restarted very often.
Perhaps move the entire logic to a lazy initializer?
(Lets track it as separate issue)

Src/WindowsServer/WindowsServer.Shared/Implementation/WindowsServerEventSource.cs
```diff
@@ -326,6 +326,94 @@ public void AzureInstanceMetadataFailureSettingDefaultPayload(string exception,
                 this.ApplicationName);
         }
 
+        [Event(33,
+            Message = "App Services Heartbeat Provider: Failed to obtain Azure App Services environment variable '{0}'. Exception raised: {1}",
+            Level = EventLevel.Informational)]
```
cijothomas: Warning

Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvironmentVariableMonitor.cs
d3r3kk: Agreed. Let's track it in a separate issue. I've added it as #875.

d3r3kk:
@cijothomas I have altered the AI_WebSDK_All-external build to time out at 90 minutes, we were pushing the upper limit of that build prior to this change. Let's discuss ways we can reduce this going forward, perhaps parallelize the builds across a few machines?

cijothomas:
@d3r3kk AI_WebSDK_All-external takes a long time because it has to download docker images everytime as it runs on 'throw away' machine. Not much we can do here. The internal builds are fast enough. -external is meant for PR from forked repos only. (which are rare)

PR #884, opened 2018-04-11
new script: GenerateReleaseMetadata

MS-TimothyMothra:
New script needed to support new release pipeline.

### DIFF
--- a/GenerateReleaseMetadata.ps1
+++ b/GenerateReleaseMetadata.ps1
@@ -1,0 +1,194 @@
+Param(
+    [Parameter(Mandatory=$true,HelpMessage="Path to Artifact files (nupkg):")]
+    [string]
+    $artifactsPath,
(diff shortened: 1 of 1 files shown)


d3r3kk:
We are going to share these scripts across repos correct? Shouldn't we add all release specific scripts to a separate repo perhaps? Then, from that repo we can create versioned release-script packages (zipped, or whatever). Each SDK could then grab the release-script package that we've configured for them and keep the concerns separated and central...

MS-TimothyMothra:
@d3r3kk absolutely we should!
Actually, most of the scripts in our repos are identical and should be migrated to a central repo.

For this week, I need this merged now. That migration will need to be our next big project.

Dmitry-Matveev:
@d3r3kk, this (separate repo) sounds like a logical next step. However; in the best traditions of work in our repos, let's not jump ahead of ourselves. Baby steps will lead us the long way :)

Issue #820, opened 2018-02-14
Override Http DependencyTelemetry "name" Property


michaelwstark:
I'm using the Application Insights DependencyCollector for collecting information our my service dependencies.  I wanted to be able to set the "name" property on these events similar to how I can for incoming request telemetry in order to categorize the calls without the parameters using the relative path uri templates in my source.  I've looked over the Activity class mentioned in some other threads, and it looks like I could add it if I wanted, but I saw some additional benefit in overriding the value so that each call wouldn't be unique in other areas of the Azure Portal.

Is this possible today without activities to override the name field?

My core concern is being able to slice and dice the data better, and the name field with the parameter values detailed within in the relative path make it difficult to out of the box.

michaelwstark:
I considered a TelemetryInitializer -- that is how we achieve the same for incoming requests.  However, I don't seem to have access to the request context for the outgoing http request, and can't seem to find an easy way to grab it -- putting in on a thread context seems dangerous.  Do you have any suggestions for how I can easily get the relative url of the outgoing request?

lmolkova:
@michaelwstark 

I think this is related to the general [issue of enriching dependency telemetry based on http request/response](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/587#issuecomment-364272438) you've commented before

Perhaps there is a bit hacky way to achieve exactly what you want with `TelemetryIntitializer`.
Raw URL is stored in  `DependencyTelmetry.Data` -  you may access it in the initializer, parse and update telemetry name.
`Name` property already has verb and relative path
`Target` represent a host name
[docs](https://docs.microsoft.com/en-us/azure/application-insights/application-insights-data-model-dependency-telemetry)
![image](https://user-images.githubusercontent.com/2347409/36332753-51eba67e-1328-11e8-81e1-c2f3b019d943.png)

More advanced scenario, when you want to use something from HTTP request that is not in the DependencyTelemetry instance, this would not work. However you may access HTTP *request* properties (not the response) by subscribing to the HTTP diagnostics source, in Start event you receive raw request and may store some properties in Activity or your custom AsyncLocal. You may use it safely in the TelemetryInititlaizer. This approach is briefly described [here](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/587#issuecomment-364272438)

[...]

Issue #747, opened 2017-11-16
Add custom dimensions to http dependencies from the request context


sanjaysingh:
For http dependencies of a given web request, there are some headers in the outgoing http request that I want to extract and put into the custom dimensions of the dependency telemetry. 
I am using .Net 4.5 ASPNet WebAPI and on AppInsight version 2.4.0. 

1. Http request comes to WebApi
2. Web Api internally makes more or more calls to an external http web service. It attaches some custom header to each of those calls that represent a very specific thing about that request.
3. Dependency is correctly getting logged. In that dependency, i want to extract the custom that was sent to the external service and write as custom dimension.

 I raised another issue [729](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/729) where i need to extract some headers from incoming request context and attach that to the dependencies. In that case i was able to find no so elegant work around of using CallContext to propagate this and it is working fine. In this case however CallContext seems risky because that outbound calls are all async and this can overstep on the CallContext properties.

I do not want to write an additional dependency telemetry and am happy with the default behavior. Just need this little bit of customization.

When i looked at the code to see if there is any extensibility hooks in the http depenency processor, it seems too closed to do anything. Microsoft.ApplicationInsights.DependencyCollector.Implementation.HttpProcesing seems perfect fit to add this customization but there is no hooks to do any customization. I have seen suggestions of attaching to diagnostics source yourself and then doing all this plumbing; that is non trivial.

I feel this AppInsight SDK needs to restructure these dependency collectors in way that it allows for consumers to plugin some extra bit of customization without going through these extra hops. The correlation code in the HttpProcessing itself could have been one of the inbuilt plugins!!

Please let me know if i am missing something and if there is an easy way to achieve this trivial requirement that i have.

lmolkova:
This is implemented in #900 

Basically during TelemetryIntitialiers run you have access to HttpRequestMessage/HttpWebRequest and HttpResponseMessage/HttpWebResponse through OperationDetails:

`dependencyTelemetry.TryGetOperationDetail("HttpResponse", out var reponseObject)`

Http keys are:

```csharp
        public const string HttpRequestOperationDetailName = "HttpRequest";
        public const string HttpResponseOperationDetailName = "HttpResponse";
        public const string HttpResponseHeadersOperationDetailName = "HttpResponseHeaders";

```

Be careful: in some cases instead of HttpResponse we can only give you HttpResponseHeaders,

Issue #587, opened 2017-05-18
Support adding arbitrary HTTP response headers from dependencies


martincostello:
It would be beneficial when HTTP dependencies are intercepted to be able to register an interceptor for `HttpResponseMessage` via DI that would allow the integrator to inspect the response's HTTP response headers.

This would allow an integrator to inspect them for values of interest which could be added to the dependency telemetry submitted to Application Insights.

For example, this would allow the activity Id (from `x-ms-activity-id`) and consumed Request Units (from `x-ms-request-charge`) from requests to Azure Cosmos DB to be added to the dependency data ([headers documentation](https://docs.microsoft.com/en-us/rest/api/documentdb/common-documentdb-rest-response-headers)).

For example:

```csharp
public interface IHttpResponseInterceptor
{
    void InspectResponse(HttpResponseMessage response, DependencyTelemetry telemetry)
}

internal class MyHttpResponseInterceptor : IHttpResponseInterceptor
{
    public void InspectResponse(HttpResponseMessage response, DependencyTelemetry telemetry)
    {
        if (response.Headers.TryGetValues("x-ms-activity-id", out IEnumerable<string> values))
        {
            telemetry.Properties["Activity Id"] = string.Join(", ", values);
        }

        if (response.Headers.TryGetValues("x-ms-request-charge", out values))
        {
            telemetry.Properties["RUs"] = string.Join(", ", values);
        }
    }
}

public class Startup
{
    public void ConfigureServices(IServiceCollection services)
    {
        services.AddSingleton<IHttpResponseInterceptor, MyHttpResponseInterceptor>();
    }
}
```

SergeyKanzhelev:
+ @lmolkova to comment

lmolkova:
You can get this with DiagnosticSource: you can subscribe to the particular source depending on the .net flavor you use (.NET Core or Full Framework) and get notifications for each outgoing HTTP request: start and stop.

If you use .NET Full Framework, please refer to the [ApplicationInsights listener](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared/Implementation/HttpDesktopDiagnosticSourceListener.cs) as an reference listener and [subscriber](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared/Implementation/HttpDesktopDiagnosticSourceSubscriber.cs) implementations.

[...]

PR #897, opened 2018-04-26
Populate DependencyTelemetry with Operation Data for Initializers

michaelwstark:
Fix Issue # .
[Consolidated]
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/900

[Original]
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/820
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/747
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/587

Populate DependencyTelemetry with operation data for initializers.

- [X] Changes in public surface reviewed
- [X] Design discussion issue # https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/587
- [X] CHANGELOG.md updated with one line description of the fix, and a link to the original issue.
- [X] The PR will trigger build, unit tests, and functional tests automatically. If your PR was submitted from fork - mention one of committers to initiate the build for you.


michaelwstark:
I assume there is a change to the core SDK nuget package required, but as that hasn't been built yet, I just worked with the update locally.  If there is a better processor for changes which modify both, please point me in the right direction -- appreciated.

Src/DependencyCollector/Shared/HttpCoreDiagnosticSourceListener.cs
```diff
@@ -350,6 +351,8 @@ internal void OnActivityStop(HttpResponseMessage response, HttpRequestMessage re
             if (response != null)
             {
                 this.ParseResponse(response, telemetry);
+                telemetry.OperationDetails[RemoteDependencyConstants.HttpResponseOperationDetailName] = response;
+                telemetry.OperationDetails[RemoteDependencyConstants.HttpResponseHeadersOperationDetailName] = response.Headers;
```
lmolkova: why do we need to populate response AND headers? I'd prefer them to be mutually exclusive. 
In case of .NET Core there is no case when a response is missing, but headers are present

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnostics/SqlClientDiagnosticSourceListener.cs
```diff
@@ -116,7 +118,14 @@ public void Dispose()
                                 Data = command.CommandText,
                                 Success = true
                             };
-                            
+
+                            // Populate the operation details for intializers
+                            telemetry.OperationDetails[RemoteDependencyConstants.SqlCommandOperationDetailName] = command;
+                            if (connection != null)
+                            {
+                                telemetry.OperationDetails[RemoteDependencyConstants.SqlConnectionOperationDetailName] = connection;
```
lmolkova: The connection is available through the command, why do we need to explicitly add it to the dictionary?

lmolkova:
@anpete 
Could you please help us review these SQL dependency collection changes?

Our goal is to enable scenario when users add arbitrary dynamic properties to their dependency telemetry. In HTTP case it could be adding request or response header value to the telemetry.

Could you please help us understand whether it is relevant for SQL and what kind of information could be useful/not useful.

Basically, we proxy DiagnostiSource payload to users through telemetry instance (but this info is not serialized and is never sent to the backend as it). It's up to users to access these objects and modify telemetry based on them.

This is a very advanced scenario we don't expect the absolute majority of users to implement.

Src/DependencyCollector/Shared/HttpCoreDiagnosticSourceListener.cs
michaelwstark: That was actually how I originally had it (until I hit the corner case where the response was not present).  My rationale for always including them was to make development of a initializer easier -- if the developer is only interested in headers, headers is always populated.  This is not a stance of principal or conviction -- you are much more familiar with what the developer experience of the initializers should be, so I will defer to you.  Please ack that you would prefer to only have the headers when the response is not available and I will make the change after I finish an update on the SDK pull request.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnostics/SqlClientDiagnosticSourceListener.cs
michaelwstark: Similar rationale as the response\headers comment above.  I was not sure that the command would always be populated based on the failure cases, and I wanted to make it easier to develop the initializer if someone was only interested in properties on the connection.  I will wait for confirmation on the comment above, and align the two to be consistent.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnostics/SqlClientDiagnosticSourceListener.cs
```diff
@@ -301,7 +306,7 @@ public void Dispose()
                                 Type = RemoteDependencyConstants.SQL,
                                 Target = string.Join(" | ", connection.DataSource, connection.Database),
                                 Data = operation,
-                                Success = true
+                                Success = true,
```
cijothomas: intentional?

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnostics/SqlClientDiagnosticSourceListener.cs
```diff
@@ -116,7 +118,10 @@ public void Dispose()
                                 Data = command.CommandText,
                                 Success = true
                             };
-                            
+
+                            // Populate the operation details for intializers
+                            telemetry.SetOperationDetail(RemoteDependencyConstants.SqlCommandOperationDetailName, command);
```
cijothomas: need similar change here as well:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessingBase.cs
This contains SQL dependency collection logic with Profiler installed.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Version 2.7.0-beta1
 - [Added support to collect Perf Counters for .NET Core Apps if running inside Azure WebApps] (https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/889)
+- [Add operation details for HTTP and SQL operation to the dependency telemetry.](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/587)
```
cijothomas: If possible create a new issue in gh, which contain detail about the specific fix. The original issue is quite long discussion, and it'll be hard to figure out which part from that is being implemented here.

cijothomas: Also helpful if an example TelemetryInitializer can be provided which utilizes this new feature.

cijothomas:
@SergeyKanzhelev  and @Dmitry-Matveev - Please share your thoughts on making this under a feature flag, and having them OFF by default. I'd assume that this is only useful if customer has their own telemetry initializer which makes use of this. Those customers can turn this feature ON, so regular users dont see increased memory usage from AI SDK.

michaelwstark:
@cijothomas - While we are waiting for @SergeyKanzhelev and @Dmitry-Matveev to respond to whether this needs to be placed under a feature flag, can you point me to a current example?  I presume it would be something similar to ExcludeComponentCorrelationHttpHeadersOnDomains or IncludeDiagnosticSourceActivities on the current dependency collection module, but am not completely sure.

Src/DependencyCollector/Shared/Implementation/SqlClientDiagnostics/SqlClientDiagnosticSourceListener.cs
michaelwstark: Not intentional, but it is a coding style from my projects.  I'll remove the trailing comma when we hear back regarding the feature flag.

CHANGELOG.md
cijothomas: Either is fine. Inline in a GH issue is easier to find, in my opinion though.

cijothomas:
@michaelwstark Yes you are pretty close. More relevant example is `DisableDiagnosticSourceInstrumentation` flag on `DependencyTrackingTelemetryModule` which turns on/off DiagnosticSource based dependency tracking. 
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs#L57

CHANGELOG.md
michaelwstark: K.  I'll code up a sample as I was intending to use this feature and include it in a new GH issue.

michaelwstark:
Thanks for the pointer.  I will create a branch off of my current working code with this change in anticipation of @SergeyKanzhelev or @Dmitry-Matveev indicating that we should include a flag for this behavior.

lmolkova:
> Those customers can turn this feature ON, so regular users dont see increased memory usage from AI SDK.

The should be no increase in memory usage. The properties DependencyTelemetry obtains already present in the memory and are cleaned up after dependency call is finished.

E.g. in case of HTTP we add Http request and Http response - there are no new allocations, except for dictionary with 2 items.

lmolkova:
@michaelwstark 

we discussed it with @cijothomas offline.
The concern is that the feature would postpone  GC for the requests/responses since reference would be kept in dependency telemetry until it is sent to the backend (that happens asynchronously).

I wonder, can we clean up the OperationDetails in `TrackDependency` call after initialize is called to address this concern?

[Here](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/9479e37efc5221f3b8e238696be1050128d21b2c/src/Microsoft.ApplicationInsights/TelemetryClient.cs#L370)
```csharp
        public void TrackDependency(DependencyTelemetry telemetry)
        {
            if (telemetry == null)
            {
                telemetry = new DependencyTelemetry();
            }

            this.Track(telemetry);

!!!         telemetry.OperationDetails.Clear();
        }
```

It totally makes sense to me, as Track happens syncronously with dependency call and request/response should not be accessed after that.

CHANGELOG.md
michaelwstark: Here is a consolidated issue with a code sample for intended use:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/900

Dmitry-Matveev:
@michaelwstark, I'm with Cijo on this one - due to increased memory usage for default scenario, I'd suggest to make this feature opt-in rather than on by default. Performant apps (let's say 500+ RPS) may all of a sudden become less performant if we store all request/responses/headers for 30 seconds (15000 * 3 objects). Super-performant apps (1000+ RPS) may struggle even if we come up with some optimization to release object sooner - it will still exist more than ASP.NET intended it to.
Making this opt-in will require a switch or flag that you are discussing with Cijo already.

lmolkova:
Chatted with @Dmitry-Matveev offline (he had to run for a meeting so cannot reply right now).
He is fine with the approach when we clean up details immediately after tracking telemetry (sync with response events). He agrees that in this case feature flag is not needed.

Dmitry-Matveev:
Thanks, @lmolkova! @michaelwstark, in case we do not copy nor keep response/request objects after the request is over - we are good without the feature flag.

michaelwstark:
@Dmitry-Matveev Thanks!  @lmolkova - I think that all of the outstanding feedback has been resolved, please let me know what else needs to be done prior to merging these PRs into their respective repos.  I noticed there are some VSTS checks which never provide any feedback -- anything I need to do there?

cijothomas:
@michaelwstark thanks for addressing the suggestions!
The VSTS checks are for CI Builds, which are not auto-triggered for forked repo pull requests. I'll manually trigger them for you and if all tests pass, we are good.
The changes in base SDK need to be first merged and published before the changes in this repo can be merged. I will start in base sdk build trigger first.

cijothomas:
@MS-TimothyMothra to ensure this is merged **AFTER** base 2.7.0-beta1 is published to myget.

MS-TimothyMothra:
Sorry, I misunderstood and prep-ed this for Beta2.
Reverted my change.

michaelwstark:
@MS-TimothyMothra has the Beta1 of the core SDK been published to myget?  When do we expect this PR to merge?  Thanks in advance.

MS-TimothyMothra:
Sorry @michaelwstark, the 2.7 Beta1 got delayed to June 1st because we're working on hotfixes for 2.6 Stable.
Feel free to contact me directly if you need more information

MS-TimothyMothra:
@michaelwstark It looks like some of your new tests are failing.

- TestZeroAndNonZeroContentResponseDiagnosticSource
- TestBasicDependencyCollectionDiagnosticSourceLegacyHeaders

Will you have time to look at these today?

michaelwstark:
@MS-TimothyMothra.  Yup.  I'll be able to take a look today.

lmolkova:
I've added these two recently:

TestZeroAndNonZeroContentResponseDiagnosticSource
TestBasicDependencyCollectionDiagnosticSourceLegacyHeaders

Probably failures are result of auto-merge and new tests are missing something.

michaelwstark:
One test is a negative case for response, and the validate call requires hint not to validate response data.  The other leveraged a new Boolean value in the validate pipeline which was not named which conflicted with the one I added.  Re-running unit tests after the update and will push shortly if they pass.

michaelwstark:
@MS-TimothyMothra you should be good to proceed now.  Let me know if you hit any more issues.

MS-TimothyMothra:
Thanks @michaelwstark ! Queuing your build now.

Issue #797, opened 2018-01-23
Activity is lost in POST requests (i.e. requests do not correlate to dependencies and other nested telemetry)


lmolkova:
## Repro Steps
1. create a new MVC/WebAPI app in VS, update AI SDKs to 2.4 or 2.5
2. Change default ValueController (or create a new controller) to log Activity current in GET and POST requests, e.g. 

```C#
   public class ValuesController : ApiController
    {
        private readonly static TelemetryClient tc = new TelemetryClient();

        public Task<string> Get(int id)
        {
            return Task.Run(() =>
            {
                tc.TrackTrace($"GET: Current Activity is '{Activity.Current?.Id}'");
                return Activity.Current?.Id;
            });
        }

        // POST api/values
        public Task<string> Post([FromBody]string value)
        {
            return Task.Run(() =>
            {
                tc.TrackTrace($"POST: Current Activity is '{Activity.Current?.Id}'");
                return Activity.Current?.Id;
            });
        }
```

##  Actual Behavior
In Get request, Activity.Current is correct therefore trace telemetry is correlated to request telemetry
In Post request Activity.Current is null, trace telemetry does not have operation Id or parent Id.

## Expected Behavior
Activity.Current is not null in post request, nested telemetry is correlated to requests.

## Note
Controller method being sync/async does not change behavior.

## Version Info 
SDK Version : 2.4, 2.5
.NET Version :	NET 4.5.1, 4.7				
How Application was onboarded with SDK(VisualStudio/StatusMonitor/Azure Extension) :
OS : VS
Hosting Info (IIS/Azure WebApps/ etc) : local, Azure Web App

## Workaround
Add and configure one of the following to the app (whatever is more convenient):
* [WebApi ActionFilter](https://docs.microsoft.com/en-us/aspnet/mvc/overview/older-versions-1/controllers-and-routing/understanding-action-filters-cs) OnActionExecuting method
* [DelegatingHandler](https://www.infoworld.com/article/2991458/application-architecture/how-to-work-with-message-handlers-in-web-api.html)
* [ApiControllerActionInvoker](http://www.c-sharpcorner.com/article/webapi-pipeline-revealed-a-true-practical-approach/)

Regardless, which way you choose, add following code **before** calling base SendAsync/InvokeActionAsync

```C#
    using System.Diagnostics;

....
            if (Activity.Current == null)
            {
[...]

PR #898, opened 2018-05-01
Report intermediate requests for lost and restored Activity

lmolkova:
Fix Issue #797 for the case when .NET 4.7.1 is installed on the machine (Azure WebApps case).

1. When request processing starts, Microsoft.AspNet.TelemetryCorrelation module creates a new Activity and fires Start Event. ApplicationInsights receives this event and creates RequestTelmetery.
RequestTelemetry gets all correlation Ids from the Activity module created. (Nothing has changed here)

2. Right before the ASP.NET (or WCF) phase in request processing, the module checks if Activity was lost (it happens in some cases including POST requests). In versions 1.0.0-1.0.1 TelemetryCorrelation module attempted to restore an Activity by creating a child one and did not notify AI about it.

It resulted in following E2E viewer behavior:
image

As the request is a grand-parent of dependency, and nothing has reported in the middle, a correct correlation between them is not possible.

3. With TelemetryCorrelation version 1.0.3, the module improves its ability to restore the Activity by using new 4.7.1. [OnExecuteRequestStep](https://msdn.microsoft.com/en-us/library/system.web.httpapplication.onexecuterequeststep(v=vs.110).aspx) ASP.NET API, see https://github.com/aspnet/Microsoft.AspNet.TelemetryCorrelation/pull/24
It also sens a new event about such activity, so we can track intermediate request and build better UI .

image

This is temporary view, until https://github.com/dotnet/corefx/pull/29208 is released along with .NET Core 2.2. At this point, TelemetryCorrelation will be able to assign `Activity.Current` and dependency would be a direct child of the request.

Next steps (not part of this PR):
- when .NET 4.7.1 is not installed,detect issue and report heartbeat property


Src/Web/Web.Nuget/Package.nuspec
```diff
@@ -19,7 +19,7 @@
       <group targetFramework="net45">
         <dependency id="Microsoft.ApplicationInsights" version="[$coresdkversion$]" />
         <dependency id="Microsoft.ApplicationInsights.WindowsServer" version="$version$" />
-        <dependency id="Microsoft.AspNet.TelemetryCorrelation" version="1.0.1" />
+        <dependency id="Microsoft.AspNet.TelemetryCorrelation" version="1.0.2" />
```
SergeyKanzhelev: nuspec and packages config should be aligned on the required version of this package

lmolkova:
> Will this logic work when TelemetryCorrelation will be fixed? I mean when customer updates TelemetryCorrelation http module, but does not upgrade Application Insights?

Yes, when TelemetryCorrelation will start setting Activity.Current (instead of creating child Activity), ApplicaitonInsights without any change would pick it up.
TelemteryCorrelaiton will stop sending the 'restored activity' event, i.e. AppInsights handler would be effectively no-op

runFunctionalTestsDependencyCollector.cmd
```diff
@@ -6,10 +6,8 @@ CALL buildReleaseFull.cmd
 set BuildRoot=%~dp0..\bin\Release\
 set VSTestPath=%PROGRAMFILES(x86)%\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\CommonExtensions\Microsoft\TestWindow\vstest.console.exe
 
-CALL "%VSTestPath%" "%BuildRoot%Test\E2ETests\E2ETests\DependencyCollectionTests.dll" /TestCaseFilter:"TestCategory=Core20" /logger:trx
```
cijothomas: did you intend to make this change? or accidently checked in!

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -279,6 +281,67 @@ public void Initialize(TelemetryConfiguration configuration)
...
+                Name = string.Format(CultureInfo.InvariantCulture, "Execute request handler ({0})", context.CreateRequestNamePrivate()),
+                Id = activity.Id,
+                Timestamp = activity.StartTimeUtc,
+                Duration = activity.Duration
+            };
+
+            intermediateRequest.Context.Operation.Id = activity.RootId;
+            intermediateRequest.Context.Operation.ParentId = activity.ParentId;
+            intermediateRequest.ResponseCode = null;
+            intermediateRequest.Properties.Add("AI internal", "Execute request handler step");
```
cijothomas: Why do we need it?

runFunctionalTestsDependencyCollector.cmd
lmolkova: oh, did not intend to remove those

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
lmolkova: this is a hint to users that it is some kind of technical, internal AI stuff that leaked into their telemetry, rather than a real request.

PR #903, opened 2018-05-03
Detect issue with POST/PUT/PATCH requests correlation  and report in event source actionable event

lmolkova:
See #797.

We have fixed the issue with #898 for users that have .NET 4.7.1 installed (all Azure WebApp users).
Azure Cloud Services do not have 4.7.1 by default.

For users that do not have 4.7.1 we recommend installing .NET 4.7.1 (without retargeting the app) or follow workaround mentioned in the #797.

We also want to see how many users are affected by the issue, so we detect it and ~~set heartbeat property~~ [**Updated**] send user actionable event source event.

We'll see if need to notify users in UI based on the number of affected users. For now, we just ensuring data is there.

### DIFF
--- a/Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
+++ b/Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
@@ -20,13 +20,14 @@
     public class RequestTrackingTelemetryModule : ITelemetryModule
     {
         private const string IntermediateRequestHttpContextKey = "IntermediateRequest";
-        private readonly IList<string> handlersToFilter = new List<string>();
+        // if HttpApplicaiton.OnRequestExecute is available, we don't attempt to detect any correlation issues
+        private static bool correlationIssuesDetectionComplete = typeof(HttpApplication).GetMethod("OnExecuteRequestStep") != null;
+
         private TelemetryClient telemetryClient;
         private TelemetryConfiguration telemetryConfiguration;
         private bool initializationErrorReported;
-        private bool correlationHeadersEnabled = true;
         private ChildRequestTrackingSuppressionModule childRequestTrackingSuppressionModule = null;
-        
+
         /// <summary>
         /// Gets or sets a value indicating whether child request suppression is enabled or disabled. 
         /// True by default.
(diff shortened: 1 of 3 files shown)


Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -19,12 +20,16 @@
     /// </summary>
     public class RequestTrackingTelemetryModule : ITelemetryModule
     {
-        private readonly IList<string> handlersToFilter = new List<string>();
+        private const string CorrelationStatusHeartbeatProperty = "CorrelationHealthStatus_RequestWithContent";
```
cijothomas: Lets use the minimum required characters. "CorHS_ReqWCont" 
"NA" instead of notsupported :) to minimize payload size

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -19,12 +20,16 @@
...
+
+        // if HttpApplicaiton.OnRequestExecute is available, we don't attempt to detect any correlation issues
+        private static bool correlationIssuesDetectionComplete = typeof(HttpApplication).GetMethod("OnExecuteRequestStep") != null;
```
cijothomas: Is there an exact version of .net framework when this will be available? I believe heartbeat already sents out .NET Framework information - can we use that alone ?

lmolkova: As we have seen in func tests, .NET version does not really matter. What matters is the version of ASP.NET and even then, customers that do not have POST, PUT or PATCH requests (e.g. serving static websites) would never experience the issue. 
We can guess that there are not much of such customers, but I'd prefer to measure it rather than guessing.

lmolkova:
As we have seen in func tests, .NET version does not really matter. What matters is the version of ASP.NET and even then, customers that do not have POST, PUT or PATCH requests (e.g. serving static websites) would never experience the issue.
We can guess that there are not much of such customers, but I'd prefer to measure it rather than guessing.

lmolkova:
@cijothomas, are you ok with it?

lmolkova:
discussed with @Dmitry-Matveev and @cijothomas user actionable event source event is typical way to achieve it and could be checked from internal telemetry and customer telemetry.  And there is no need to build additional UX features for it.

lmolkova:
@cijothomas ping :)

lmolkova:
build issue is transient, related to artifacts publishing and does not repro on a manually triggered build.

PR #907, opened 2018-05-05
enable package signing

MS-TimothyMothra:


Src/WindowsServer/WindowsServer.Nuget/WindowsServer.Nuget.csproj
```diff
@@ -20,6 +20,7 @@
     <TargetFrameworkProfile />
     <NuGetPackageImportStamp>
     </NuGetPackageImportStamp>
+    <DisablePackageAnalysis>true</DisablePackageAnalysis>
```
Dmitry-Matveev: Why do we need this entry on each package - what exactly is disabled if this is true?
Does signing fail without it? If it does - let's have a medium pri issue to comply with package analysis in the upcoming milestone.

MS-TimothyMothra:
I tested it without and it works fine so removing.

PR #908, opened 2018-05-07
Fix 503 WCF test that sometimes fails because of intermediate request

lmolkova:
After merging #898 I found that some Web Func testsfail sometimes in external, hosted build. 

It happens with WCF app tests because intermediate request is sometimes tracked for non-reported handlers.

To fix it, I started tracking intermediate request together with 'main' request. I.e. intermediate requests would never be reported for non-tracked handlers. After ~15 hosted test runs, all web func tests pass.
It should also reduce the noise for users. 

### DIFF
--- a/Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
+++ b/Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
@@ -19,13 +19,14 @@
     /// </summary>
     public class RequestTrackingTelemetryModule : ITelemetryModule
     {
+        private const string IntermediateRequestHttpContextKey = "IntermediateRequest";
         private readonly IList<string> handlersToFilter = new List<string>();
         private TelemetryClient telemetryClient;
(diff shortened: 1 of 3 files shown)


Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -211,6 +212,12 @@ public void OnEndRequest(HttpContext context)
 
             if (this.childRequestTrackingSuppressionModule?.OnEndRequest_ShouldLog(context) ?? true)
             {
+                var intermediateRequest = (RequestTelemetry)context.Items[IntermediateRequestHttpContextKey];
+                if (intermediateRequest != null)
```
Dmitry-Matveev: Is there a guarantee that internal request will reach this OnEndRequest handler faster all the time?
I remember that with TransferRequestHandler we didn't have such a guarantee as per @MS-TimothyMothra.
Without that, the tests may still fail very rarely but customers with high load of requests would sometimes see both reported vs. one reported.

PR #915, opened 2018-05-16
Minor change to cleanup docker script

cijothomas:
Fix Issue # .

### DIFF
--- a/dockercleanup.ps1
+++ b/dockercleanup.ps1
@@ -28,8 +28,7 @@ Write-Host "Removing E2E Images"
 & docker rmi -f e2etests_ingestionservice
 & docker rmi -f e2etests_e2etestwebapi
 & docker rmi -f e2etests_e2etestwebappcore20
-& docker rmi -f e2etests_e2etestwebapp 
-& docker rmi -f e2etests_sql-server
+& docker rmi -f e2etests_e2etestwebapp
 
 Write-Host "Removing dangling images"
 docker images -f "dangling=true" -q | ForEach-Object {docker rmi $_ -f}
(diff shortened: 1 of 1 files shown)


MS-TimothyMothra:
Would it be safer to make that a collection of container names and then for-each the collection to stop and remove?

That might help us avoid another situation where we weren't cleaning up containers.

cijothomas:
@MS-TimothyMothra Not really. The space issue is not caused by those containers - they may show 10GB as size, but of that 99.99% is from base image which we don't clear. These images are cleaned up to not interfere with the next test run, not for space issue.

The likely reason why space wasted is dangling images, which were part of cleanup script - but the script itself was not run.

Write-Host "Removing dangling images"
docker images -f "dangling=true" -q | ForEach-Object {docker rmi $_ -f}

Write-Host "Removing E2E Containers"
docker ps -a -q | ForEach-Object {docker rm $_ -f}

dockercleanup.ps1
```diff
@@ -16,20 +16,22 @@ Write-Host "Stopping E2E Containers"
...
+& docker rm e2etests_azureemulator
 
 Write-Host "Removing E2E Images"
 & docker rmi -f e2etests_ingestionservice
 & docker rmi -f e2etests_e2etestwebapi
 & docker rmi -f e2etests_e2etestwebappcore20
-& docker rmi -f e2etests_e2etestwebapp 
-& docker rmi -f e2etests_sql-server
+& docker rmi -f e2etests_e2etestwebapp
+& docker rmi -f e2etests_azureemulator
```
MS-TimothyMothra: should "e2etests_sql-server_1" be a part of this list?

Issue #913, opened 2018-05-15
Library instrumentation guidance with DiagnosticSource


lmolkova:
We are looking for help from the tracing community to review out library instrumentation guidance

https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/LibraryInstrumentationGuidance.md

Start of the discussion:
https://github.com/StackExchange/StackExchange.Redis/issues/833

We will be happy to hear your feedback about anything, some exiting concerns:
- no configuration
- no contract 
- Tags vs payload usage
- lack of detail
- efficiency/performance
- concrete vs anonymous types

We hope to come up with the guidance that would work well with library owners and tracing systems.
Our philosophy for tracing still has to follow some principles like
- automagical with (almost) zero configuration
- (almost) zero perf impact when nobody listens to, and runtime sampling and verbosity control when there is a listener
- no additional dependencies, so .NET is traceable itself
- ability to build rich UI on top of it and run our important customer scenarios

We'll publish the result in more official place.

/cc @cwe1ss, @NickCraver, @stebet,  @vancem

stebet:
Here are some of my ideas after the discussions that have alreayd taken place in several other issues. By all means chime in and let's try to get everyone on the same page on how to work with this, as I think all of us can see the potential ad would love to see that potential realized.

Here are some points to get things started :)

## What to do with current code that does not adhere to todays guidelines? (ASP.NET, HttpClient, SqlCommand to name a few)

1. ### Convert anonymous payloads into strongly typed objects
    This should be possible with little chance of breaking backwards compatibility as long as the field names match the anonymous object if I'm not mistaken, since reflection is required anyways to fetch the values. As long as they are named the same, things should not break. 

1. ### New versioned events to get up to spec?
    In the case of for example `HttpClient`, could a new event called `System.Net.Http.HttpRequestOut.V2` or somesuch (naming scheme debatable) be created? This is already being done in the [DiagnosticsHandler](https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/src/System/Net/Http/DiagnosticsHandler.cs) which is already sending two events, one of which is deprecated.

    I know it isn't optimal to sprinkle more diagnostic code into code that already has it but that would ideally be a temporary solution until deprecated events can be removed, after being clearly marked as such using the `[Deprecated]` attribute (if using strongly typed objects as above) as well as the possibility to use something like `AppContext` switches to turn them off for performance reasons.

[...]

PR #917, opened 2018-05-19
[WIP] Library instrumentation guidance review

lmolkova:
Fix Issue #913.

Updating library instrumentation guidance per feedback.

This is a work in progress and there are many TODO/TBD items in the doc.
Consumer guide will be started separately.


LibraryInstrumentationGuidance.md
```diff
@@ -1,53 +1,115 @@
...
 
-## Instrumentation 
+## Bare Minimum Instrumentation 
 
-In the simplest case, the operation that is being monitored has to be wrapped by an activity. However in order to minimize performance impact the activity should be only created if there is any listener waiting for it.
+In the simplest case, the operation that is being monitored has to be wrapped by an activity. 
 
 ```csharp
+    // create DiagnosticListener with unique name
     static DiagnosticListener source = new DiagnosticListener("Example.MyLibrary");
```
vancem: I would change Example.MyLibrary, to something like 'NameSpace.ComponentName'  to give people guidance on what a good name looks like

LibraryInstrumentationGuidance.md
```diff
@@ -1,53 +1,115 @@
...
-    // create activity only when requested
-    if (source.IsEnabled() && source.IsEnabled("Example.MyLibrary.MyOperation"))
+    
+    // check if listener is interested in this operation
+    if (source.IsEnabled() && source.IsEnabled("MyOperation"))
     {
-        activity = new Activity("Example.MyLibrary.MyOperation");
-        source.StartActivity(activity, new { Input = input });
+        activity = new Activity("MyOperation");
+        activity.Start();
```
vancem: You can simplify to 

activity = new Activity("MyOperation").Start();

But we probably want to make a helper  'DiagnosticListener.StartIfEnabled(string)' which does the lines above (and inlines), and thus the guidance is to do 
```
Activity activity = source.StartIfEnabled("MyOperation")
```

LibraryInstrumentationGuidance.md
```diff
@@ -1,53 +1,115 @@
...
     object output = null;
     try
     {
         // perform the actual operation
-        output = RunOperation(input);  
+        output = RunOperation(input);
+    }
+    catch (Exception ex)
+    {
+        activity?.AddTag("error", true);
```
vancem: Tags values are strings so this has to be 

activity?.AddTag("error", "true");

LibraryInstrumentationGuidance.md
```diff
@@ -1,53 +1,115 @@
...
     }
     finally
     {
         // stop activity if started
         if (activity != null)
-             source.StopActivity(activity, new { Input = input, Output = output }); 
+             source.StopActivity(activity, new MyStopPayload { Input = input, Output = output });
```
vancem: We should have a discussion about whether we should guide people to use explicit payload classes.   They can improve efficiency and ease of use on the subscriber side, HOWEVER, they also are a mildly painful to use and do couple the receiver to the exact version of the sender.   We do have 'fast' reflection APIs (see PropertyFetch class in DIagnosticSourceEventSource) that we can publically expose.    These are 'pretty fast (< 10 instructions to fetch), and are probably good enough.

If that is not sufficient, and we really want to recommend everyone making their own types, you should include the definition of MyStopPayload in the example and we need give guidance on naming (the type should live in the YOUR_COMPONENT.DiagnosticSource namespace) and we need to describe  versioning, and the fact that using Properties (not fields) is better.

LibraryInstrumentationGuidance.md
vancem: We should probably add special helpers like 'AddErrorTag' for canonical tags as we find they are very generic.   That makes the code simpler and allows Activity to encode it more efficiently in the future.

LibraryInstrumentationGuidance.md
```diff
@@ -1,53 +1,115 @@
...
+        source.StartActivity(activity, new MyStartPayload {Input = input});
+    else
+        activity.Start();
+```
+
+DO NOT cache results of IsEnabled: tracing system may subscribe/unsubscribe at any moment. Results of IsEnabled are dynamic and depend on the context (`Activity.Current`, `Input`).
+
+### Context Propagation
+
+In distributed systems, one user operation may be processed by many applications. To trace operation end-to-end, some context must be propagated along with external call.
```
vancem: This needs more meat.  Basically an example (or leave it out, because as it is it does not help anyone).

vancem:
Thanks @lmolkova for driving this.   I agree with the general community that we need more investment in getting good instrumentation 

Are we saying that we have some examples of this now in Azure code?   If it is open source, we should put the urls here as examples are useful (and I would be interested in seeing them).

LibraryInstrumentationGuidance.md
lmolkova: having 'status' on the Activity would be great once we agree on the contract!

TL;DR;
So, I would  recommend setting a status code and optionally sending exception event.
 
Status should still provide "OK" or "something else" (like [OpenCensus](https://opencensus.io/trace.html) does). Having [strict enum-kind of mapping](https://github.com/census-instrumentation/opencensus-specs/blob/33b9a1af01c0cfb36036471fa9c464c7e7c6494d/trace/HTTP.md#mapping-from-http-status-codes-to-trace-status-codes) could be an option too, but it seems hard to do this mapping for each exception in the library 

Since the result of operation is special and essential, I think we should have a `Status` property (and `SetStatus(string)` method) on the Activity.

-------
I put "error" as a placeholder and do not think we should use it. Boolean value as [OpenTracing](https://github.com/opentracing/specification/blob/master/semantic_conventions.md#span-tags-table) suggests here does not seem to be informative enough: HTTP, gRPC and perhaps other communication would rather provide a status code. [OpenCensus](https://opencensus.io/trace.html) suggests status code + details.

While it is useful to stick to the tags suggested by [OpenTracing](https://github.com/opentracing/specification/blob/master/semantic_conventions.md#span-tags-table), I don't really understand the guidance there as well.

@cwe1ss, could you please advise us, what is the OpenTracing recommendation for errors

I see suggestion to use 
```
event="error"
error.object=<error object instance>
```
where possible. What about generic (non http) status recommendation in the tags?

This recommendation translates to DiagnosticSource 'Exception' event rather than a tag on the Span/Activity.
After working with exception events for a while, I can see an issue with them:
We expected someone would be interested in exceptions only without Activities. It seems tracking ALL exceptions without context is not that useful - it could mean overwhelming amount of handled exceptions and we hear users are interested mostly in unhandled ones.
Because of the same assumptions they [complicate instrumentation](https://github.com/Azure/azure-service-bus-dotnet/blob/dev/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs#L383) a bit.

LibraryInstrumentationGuidance.md
lmolkova: So we had a bit of discussion about it with @NickCraver in https://github.com/StackExchange/StackExchange.Redis/issues/833 and https://github.com/dotnet/corefx/issues/18114.
I believe his concerns are 1) hard to use 2) no contract and compile time checks. 
I agree with the first one, but not the second: we have to keep events and anonymous payloads backward compatible anyway and in reality for end-user application there is no compile time check.
 
It seems we do recommend using explicit types in [DiagnosticListnener guide](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/DiagnosticSourceUsersGuide.md#payloads) already.

Property fetching: we have [Microsoft.Extensions.DiagnosticAdapter](https://github.com/aspnet/EventNotification) and while it does efficient payload parsing, it does not [play well](https://github.com/aspnet/EventNotification/issues/67) with Activities/sampling scenarios and deprecated events.
So we should either invest in DiagnosticAdapter or deprecate it and expose Property Fetcher.

Anyway, I agree we need more info about naming/versioning. I will add guidance from [DiagnosticListnener guide](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/DiagnosticSourceUsersGuide.md#payloads). Eventually we can convert this doc +diagnostic listener +  activity guide to several articles on docs.ms.com

LibraryInstrumentationGuidance.md
cwe1ss: Having "activity.Start()" and "source.StartActivity()" is very confusing anyway and I too think that this should be simplified somehow in the API.

LibraryInstrumentationGuidance.md
cwe1ss: #### Status code

OpenTracing currently only defines the "http.status_code" header and no non-http status codes. As there's no real protocol-independent standard (or none that I know of), I doubt that it makes sense to have a generic "status" tag/property. Instead, I think that libraries that have status codes should add their own tags - e.g. "grpc.status_code".

#### Errors
As far as I understand, the OpenTracing "error" tag is for "Exception"-like cases, yes. Most OpenTracing instrumentation I've seen adds the "Exception" details via a `span.Log( "event"="error", "error.object"=exception )` call (pseudo code). 
Of course, the question is, whether this a case of the "log & throw" anti-pattern. If there's a parent span/activity it might log the same exception and re-throw again. 

Not sure what the best solution is here. The instrumentation code could only set the "error" tag and not log/raise the exception. This means it would rely on some parent span/activity do log it but what if there isn't one? Also, it might make debugging harder as the exception details are associated with an activity/span that did not raise it. Imagine there's a retry-scenario where something fails twice and succeeds on the third attempt. Who's responsible for logging the exceptions?

There's also a difference between OpenTracing and Activity because with OpenTracing's "Log" concept, the exception always stays with the span whereas this is more decoupled in `Activity` where it currently can only be done via a separate DiagnosticSource event.

LibraryInstrumentationGuidance.md
cwe1ss: It would be great if "PropertyFetcher" became a first class citizen. I tried the DiagnosticAdapter project but I removed it again because I thought it was too complicated and of limited use. Using a `switch` on the event names and the PropertyFetcher stuff isn't much less readable.

LibraryInstrumentationGuidance.md
```diff
@@ -1,53 +1,115 @@
...
 > ### *__TODO__ - provide a pointer to a document with more advanced instrumentation examples (WIP)* 
 
+This minimum instrumentation give tracing system a chance to trace something like '<timestamp> [Example.MyLibrary] MyOperation, Id 'abc.1.2.3.', took <duration> ms, with properties [{ "error": true }]'.
+While in majority of cases some other details are desirable (and we will get to it [here](TBD)), this instrumentation gives bare minimum.
+
+### Instrumentation control and sampling considerations
+
+**DO** make `DiagnosticListener` name globally unique.
+Every DiagnosticListener has a name. Tracing system discovers all available sources and subscribes to them based on the name. 
+[DiagnosticSource guide](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/DiagnosticSourceUsersGuide.md#naming-conventions) provides more naming considerations.
```
cwe1ss: As I've said in #913, this document should be merged with the existing UserGuides.

LibraryInstrumentationGuidance.md
cwe1ss: You know my concerns about having hard-coded header names etc. in instrumentation code. We should have a proper discussion about this at some point. Until then, "context propagation" should not be documented IMO.

cwe1ss:
@lmolkova thank you for incorporating my feedback!

LibraryInstrumentationGuidance.md
```diff
@@ -1,53 +1,104 @@
...
+
+**DO** Create one `DiagnosticListener` instance for a particular source (make it static): each time when new DiagnoscticListener is created, all subscribers are notified and about it.
+
+**DO** call 'IsEnabled()' for every operation.
+The first `source.IsEnabled()` call efficiently tell if there is any listener in the system for our source.
```
ReubenBond: s/tell/tells maybe

LibraryInstrumentationGuidance.md
```diff
@@ -1,53 +1,104 @@
...
+
+**DO** call 'IsEnabled(OperationName)' for every operation
+If there is a listener, we ask tracing system if it is interested in 'MyOperation' with `source.IsEnabled(OperationName)`. It enables filtering based on the operation name, but more importantly gives a tracing system a chance to sample this trace out based on the current Activity: tracing system samples events consistently to trace all or nothing for high-level operation. 
+
+**DO** fire 'OperationName.Start' event. Tracing system may need to inject additional properties (e.g. Http headers), add tags or mutate some properties on `Activity`. In [W3C distributed tracing standard](https://w3c.github.io/distributed-tracing), `tracestate` support depends on this extensibility. In many cases tracing system may disable this event, so guard 'Start' event with `IsEnabled` cliche, but make sure `Activity` is started in any case.
+
+**DO NOT** cache results of IsEnabled: tracing system may subscribe/unsubscribe at any moment. Results of IsEnabled are dynamic and depend on the context.
+
+**CONSIDER** providing more context via 'IsEnabled(OperationName, Input)' overload instaed of 'IsEnabled(OperationName)'
```
ReubenBond: instead*

vancem (approved):
I am OK with this guidance.   There is the question of how people find it, but that should not block this checkin

lmolkova:
I have not provided an update in a while...
We've been working on investigating and evaluating partnership with OpenCensus: https://open.microsoft.com/2018/06/13/microsoft-joins-the-opencensus-project/
vendor-agnostic polyglot libs and tools for distributed tracing.

We are looking into providing common guidance for lib instrumentation later this year. **We still recommend instrumenting .NET libraries with DiagnosticSource and Activity** but want to have a consistent story in a polyglot environment, and we are still finalizing details. 

@ReubenBond thanks for the review. I'll fix the typos and merge it.

@vancem thanks! there are a lot of TBDs in the guidance, so it's not really ready for docs.microsoft.com. I'll work on this once we finalize OpenCensus story.

MS-TimothyMothra:
@lmolkova should we close this?

PR #948, opened 2018-07-07
Fix race condition in generic diagnostic listener

lmolkova:
When generic diagnostic listener subscribes to DiagnosticSource, it attempts to use this in its own constructor, while some properties (filters) are not initialized yet.

As a result, if DiagnosticSource is created BEFORE the generic listener, it's guaranteed to miss subscription event and never receive anything.

This PR fixes this issue.

### DIFF
--- a/Src/DependencyCollector/Shared/Implementation/DiagnosticSourceListenerBase.cs
+++ b/Src/DependencyCollector/Shared/Implementation/DiagnosticSourceListenerBase.cs
@@ -16,13 +16,21 @@ namespace Microsoft.ApplicationInsights.DependencyCollector.Implementation
         protected readonly TelemetryClient Client;
         protected readonly TelemetryConfiguration Configuration;
 
-        private readonly IDisposable listenerSubscription;
+        private IDisposable listenerSubscription;
         private List<IDisposable> individualSubscriptions;
 
         protected DiagnosticSourceListenerBase(TelemetryConfiguration configuration)
         {
             this.Configuration = configuration;
             this.Client = new TelemetryClient(configuration);
+        }
+
+        public void Subscribe()
+        {
+            if (this.listenerSubscription != null)
+            {
+                return;
+            }
 
             try
             {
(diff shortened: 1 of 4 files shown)


lmolkova:
Please ignore build failure, it was cancelled, re-started and passed, but github did not pick up successful run

Src/DependencyCollector/Shared/Implementation/DiagnosticSourceListenerBase.cs
```diff
@@ -16,13 +16,21 @@ internal abstract class DiagnosticSourceListenerBase<TContext> : IObserver<Diagn
...
+        private IDisposable listenerSubscription;
         private List<IDisposable> individualSubscriptions;
 
         protected DiagnosticSourceListenerBase(TelemetryConfiguration configuration)
         {
             this.Configuration = configuration;
             this.Client = new TelemetryClient(configuration);
+        }
+
+        public void Subscribe()
```
MS-TimothyMothra: for this public method, can you please add a <summary> describing What this is, Why this is?

Src/DependencyCollector/Shared/Implementation/DiagnosticSourceListenerBase.cs
```diff
@@ -16,13 +16,25 @@ internal abstract class DiagnosticSourceListenerBase<TContext> : IObserver<Diagn
         protected readonly TelemetryClient Client;
         protected readonly TelemetryConfiguration Configuration;
 
-        private readonly IDisposable listenerSubscription;
+        private IDisposable listenerSubscription;
         private List<IDisposable> individualSubscriptions;
 
         protected DiagnosticSourceListenerBase(TelemetryConfiguration configuration)
```
MS-TimothyMothra: Because the Subscribe method is required, should we make a reference to it in the constructor?
I don't know how often a developer creates a new DiagnosticSourceListener (probably not often :) ), but I wonder that they may not know to look for the Subscribe method.

lmolkova: this is an internal class, it's only us who call this method and just once.  I can add the comment, though

PR #952, opened 2018-07-18
Use W3C trace Id format for Activities that do not have parents

lmolkova:
We are going to gradually enable W3C support in ApplicaitonInsights.

While the details and backward compatibility conditions are still not clear, we want to make sure that customers have enough time to transition to W3C format and this change is the first step in this transition.

This change generates Activity RootId in the format of [W3C traceId](https://github.com/w3c/distributed-tracing/blob/master/trace_context/HTTP_HEADER_FORMAT.md#trace-id) when there are no parents.

I.e. if the service has received Request-Id (or legacy) header from upstream, it will use it, no matter which format it follows.

However, if there were no such header in the incoming request, the new operation id will comply with W3C spec for traceId. 

This is a temporary workaround that could be removed once W3C supported in .NET (Activity).

Sampling tests shows that distribution of sampling score for the new traceId generation algorithm is uniform.


Src/DependencyCollector/Shared/Implementation/ClientServerDependencyTracker.cs
```diff
@@ -59,19 +60,22 @@ internal static DependencyTelemetry BeginTracking(TelemetryClient telemetryClien
...
+                // we want to generate Activity Ids in the W3C compatible format.
+                // While .NET changes to Activity are pending, we want to ensure trace starts with W3C compatible Id
+                // as early as possible, so that everyone has a chance to upgrade and have compatibility with W3C systems once they arrive.
+                // So if there is no parent Activity (i.e. this request has happened in the background, without parent scope), we'll override 
+                // the current Activity with the one with properly formatted Id. This workaround should go away
+                // with W3C support on .NET https://github.com/dotnet/corefx/issues/30331
                 if (currentActivity == null)
                 {
-                    activity.SetParentId(telemetry.Id);
+                    activity.SetParentId(StringUtilities.GenerateTraceId());
```
cijothomas: do you need to stop this in EndTracking?

Src/Common/ActivityExtensions.cs
```diff
@@ -0,0 +1,31 @@
+﻿namespace Microsoft.ApplicationInsights.Common
+{
+    using System.Diagnostics;
+
+    internal static class ActivityExtensions
+    {
+        public static Activity UpdateParent(this Activity original, string newParentId)
```
cijothomas: Unittest this method

Src/DependencyCollector/Shared/Implementation/ClientServerDependencyTracker.cs
lmolkova: nope, it's stopped on line 81

Src/Common/ActivityExtensions.cs
lmolkova: done

lmolkova:
@MS-TimothyMothra could you please have a look?

Issue #970, opened 2018-08-08
ServiceBus correlation is broken in 2.7.1


lmolkova:
ServiceBus correlation is broken in 2.7.1: request (message processing) is not correlated to the dependencies (send operation)

PR #971, opened 2018-08-08
Fix broken correlation for ServiceBus

lmolkova:
Fix Issue #970  .

When we generate W3C compatible trace id, we should check that activity does not have a parent and parent id, and only then assign the new operation id.


Src/DependencyCollector/Shared/HttpCoreDiagnosticSourceListener.cs
```diff
@@ -303,7 +303,7 @@ internal void OnActivityStart(HttpRequestMessage request)
             // So if there is no parent Activity (i.e. this request has happened in the background, without parent scope), we'll override 
             // the current Activity with the one with properly formatted Id. This workaround should go away
             // with W3C support on .NET https://github.com/dotnet/corefx/issues/30331
```
MS-TimothyMothra: If you intend to remove this at a later date, consider marking this with a TODO so we can find it later.
Or even better, open a cleanup issue and assign it to the 3.0 milestone! :)

PR #999, opened 2018-09-13
fix fxcop warnings

MS-TimothyMothra:
- disable TaskScheduler warning
- remove unnecessary empty array allocation
- change methods to static
- fix for Uri warning

no functional changes

### DIFF
--- a/Src/PerformanceCollector/Filtering.Shared/Implementation/Service contract/CalculatedMetricInfo.cs
+++ b/Src/PerformanceCollector/Filtering.Shared/Implementation/Service contract/CalculatedMetricInfo.cs
@@ -4,6 +4,7 @@
     using System.Globalization;
     using System.Linq;
     using System.Runtime.Serialization;
+    using Microsoft.ApplicationInsights.Common;
 
     [DataContract]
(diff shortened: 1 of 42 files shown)


Src/PerformanceCollector/Filtering.Shared/Implementation/Service contract/CalculatedMetricInfo.cs
```diff
@@ -77,7 +77,19 @@ public override string ToString()
...
+
+        private string FilterGroupsToString()
+        {
+            if (this.FilterGroups == null)
+            {
+                return string.Empty;
+            }
+            else
+            {
+                return string.Join(", ", this.FilterGroups.Select(filterGroup => filterGroup.ToString()));
```
Dmitry-Matveev: Do we need a Format Provider for this and other ToString() methods as well?

MS-TimothyMothra: in this instance no..
this is our overridden ToString() method that has a custom string builder. Inside our custom method we use CultureInfo.InvariantCulture.

PR #1000, opened 2018-09-13
fix fxcop disposable

MS-TimothyMothra:
the following classes had disposable fields but were not themselves disposable.

### DIFF
--- a/Src/PerformanceCollector/Perf.Shared/Implementation/QuickPulse/IQuickPulseServiceClient.cs
+++ b/Src/PerformanceCollector/Perf.Shared/Implementation/QuickPulse/IQuickPulseServiceClient.cs
@@ -5,7 +5,7 @@
 
     using Microsoft.ApplicationInsights.Extensibility.Filtering;
 
-    internal interface IQuickPulseServiceClient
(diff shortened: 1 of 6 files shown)


Src/PerformanceCollector/Perf.Shared/Implementation/QuickPulse/IQuickPulseServiceClient.cs
```diff
@@ -5,7 +5,7 @@
 
     using Microsoft.ApplicationInsights.Extensibility.Filtering;
 
-    internal interface IQuickPulseServiceClient
+    internal interface IQuickPulseServiceClient : IDisposable
```
Dmitry-Matveev: If Dispose() is .NET version specific (e.g. we need it only for .NET core) - then you can make this inheritance conditional on #ifdef NETSTANDARD13 or similar. This will remove two empty implementations from the clients on another .Net versions.

MS-TimothyMothra: conditional inheritance sounds like a slippery slope. I would like to avoid this

PR #1008, opened 2018-09-25
convert Web project to PackageReference

MS-TimothyMothra:
convert Web project to PackageReference.
updated DependenceCollector and WindowsServer nuspec to new location of Web.dll
fixed content structure of HostingStartup project.
temporarily suppressing WarningsAsError in Web project.


Src/DependencyCollector/NuGet/DependencyCollector.NuGet.csproj
```diff
@@ -16,7 +16,7 @@
     <NoWarn>2008</NoWarn>
     <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\..\</SolutionDir>
     <PackageSpecFile>$(MSBuildProjectDirectory)\Package.nuspec</PackageSpecFile>
-    <PackageVersionFile>$(BinRoot)\$(Configuration)\Src\Web\Web.Net45\Microsoft.AI.Web.dll</PackageVersionFile>
+    <PackageVersionFile>$(BinRoot)\$(Configuration)\Src\Web\Web\net45\Microsoft.AI.Web.dll</PackageVersionFile>
```
SergeyKanzhelev: no need in this property after the switch. It is used by old way NuGets were packaged

Src/Web/Web/Web.csproj
```diff
@@ -0,0 +1,65 @@
...
+  <Import Project="..\..\..\Common.props" />
+  <Import Project="..\..\..\Nupkg.props" />
+
+  <PropertyGroup>
+    <RootNamespace>Microsoft.ApplicationInsights.Web</RootNamespace>
+    <AssemblyName>Microsoft.AI.Web</AssemblyName>
+    <DocumentationFile>$(OutputPath)\$(AssemblyName).XML</DocumentationFile>
+    <TargetFramework>net45</TargetFramework>
+    <Prefer32Bit>false</Prefer32Bit>
+    <DefineConstants>$(DefineConstants);NET45;ALLOW_AGGRESSIVE_INLIGNING_ATTRIBUTE</DefineConstants>
```
SergeyKanzhelev: no need to define `NET45` in new project types btw

MS-TimothyMothra: I think it's best to leave it for now unless we want to investigate the full impact.
Our Common projects are still using the NET45 preprocessor. 
In the future, we could change our preprocessors to be NET45 by default and the if condition could be NETSTANDARD.

Issue #1009, opened 2018-09-26
Current Activity is null for event 


lmolkova:
We have a bunch of errors like:

Current Activity is null for event = 'System.Net.Http.HttpRequestOut.Stop'

it also affects ServiceBus.Receive and Process.

It prevents telemetry from being collected.

I was not able to reproduce it and cannot imagine a valid case where this could happen.  It indicates some bug.

lmolkova:
Ok, so I was able to find the root cause of this issue and repro it locally.

Apparently, it happens when multiple apps are hosted in the same process (somewhat typical for ServiceFabric and possible on AspNet Core).

In this case, each app initializes its own AppInsights pipeline and multiple modules start working in parallel. 

Because of changes done in https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/952, we in some cases started to stop an Activity on the forst app. Causing all other hosts to fail with null Current Activity and preventing duplicate depednency collection.

Multi-tracking is handled here: https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/621

I'm closing this issue as it will be addressed by fixing https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/621, which I'm working on.

PR #1010, opened 2018-09-26
Partially mitigate issue with null Current Activity in stop event, log more details

lmolkova:
See #1009

This change 
1. mitigates issue for http: we can still track the dependency but have to start tracking it in Start event.
2. adds more logging for the case when Activity is null in Stop event so we can understand why it happens and fix it.


Src/DependencyCollector/Shared/HttpCoreDiagnosticSourceListener.cs
```diff
@@ -312,12 +322,44 @@ internal void OnActivityStart(HttpRequestMessage request)
...
+            {
+                if (!dependency.Properties.ContainsKey(item.Key))
+                {
+                    dependency.Properties[item.Key] = item.Value;
+                }
+            }
+
+            this.client.Initialize(dependency);
+
+            dependency.SetOperationDetail(RemoteDependencyConstants.HttpRequestOperationDetailName, request);
```
lmolkova: should first call setoperationdetail, than initialize

lmolkova:
I'm closing this PR, root cause found and this issue is not harmful (on the contrary, someone can consider it a feature).

There are some nice test improvements here. I'll send another PR with them.

PR #1013, opened 2018-09-27
Prevent duplicate dependency collection when multiple apps run in the same process

lmolkova:
See https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/621.

ServiceFabric supports running multiple apps on the same host. Each app has its own AppInsights pipeline configured. i.e. there are multiple active dependency collectors and diagnostics listeners. Each of them receives even about dependency calls and tracks duplicates.

This change allows only one listener to be 'active': i.e. track telemetry.
When it is disposed, a new active listener is chosen between survivors.

New class `ActiveSubscriptionManager` was introduced (public, but non-browsable), that I want to reuse in AspNetCore SDK to prevent duplicate requests collection.

- [ ] Changes in public surface reviewed: **NO**
- [x] The PR will trigger build, unit tests, and functional tests automatically. If your PR was submitted from fork - mention one of committers to initiate the build for you.


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## Version 2.8.0-beta3
```
lmolkova: please advise what is our plan for beta3 and whether i need to change it to 2.9-beta1

MS-TimothyMothra: there was an informal discussion about a beta3, but i think the plan is to go on to 2.9-beta1.
Lets get this cleared up during standup today.

Src/Common/ActiveSubsciptionManager.cs
```diff
@@ -0,0 +1,63 @@
...
+    using System.Linq;
+    using System.Threading;
+
+    /// <summary>
+    /// Keeps one active subscription to specific DiagnosticSource per process.
+    /// Helps manage subsciptions in scenarios where multiple apps hosted in the same process.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+#if DEPENDENCY_COLLECTOR
+    public
```
SergeyKanzhelev: why should this class be public?

Src/DependencyCollector/Shared/HttpCoreDiagnosticSourceListener.cs
```diff
@@ -111,11 +117,20 @@ public void OnError(Exception error)
...
-            const string ErrorTemplateValueParse = "Event {0}: cannot parse '{1}' as type {2}";
-
             try
             {
+                // It's possible to host multiple apps (ASP.NET Core or generic hosts) in the same process
+                // Each of this apps has it's own DependencyTrackingModule and corresponding Http listener.
+                // We should ignore events for all of them except one
+                if (!SubscriptionManager.IsActive(this))
+                {
+                    DependencyCollectorEventSource.Log.NotActiveListenerNoTracking(evnt.Key, Activity.Current?.Id);
```
SergeyKanzhelev: this is a part of a solution. It ensures subscription only happens once and picks one "victim" to send requests thru.

I think better solution will be to check application name in the listener and only process requests that belong to the application that created the subscriber

lmolkova: All applications created subscribers. All of them are valid. Moreover, any listener could die at any moment and we still need to report telemetry through the next victim.

The limitation of this is that all applications have to have the same config/ikey. We discussed it briefly in the Microsoft/ApplicationInsights-aspnetcore#621 and everyone seems to be on board with it.

Src/Common/ActiveSubsciptionManager.cs
lmolkova: because we want to reuse it in the AspNetCore SDK for request deduplication

MS-TimothyMothra: Can we move this to the DependencyCollector project instead of Common?
The mixed accessibility levels seem wrong. :)

Src/DependencyCollector/Shared.Tests/ActiveSubsciptionManagerTests.cs
```diff
@@ -0,0 +1,162 @@
...
+            dlSubscription.Attach(subs);
+
+            Assert.IsTrue(dlSubscription.IsActive(subs));
+
+            dlSubscription.Detach(subs);
+            Assert.IsFalse(dlSubscription.IsActive(subs));
+        }
+
+        [TestMethod]
+        public void AttachAndDetachMultiple()
```
Dmitry-Matveev: Minor: This test does not seem to `Detach()`.

Src/DependencyCollector/Shared.Tests/ActiveSubsciptionManagerTests.cs
```diff
@@ -0,0 +1,162 @@
...
+
+            dlSubscription.Detach(subs2);
+            Assert.IsTrue(dlSubscription.IsActive(subs1));
+
+            dlSubscription.Detach(subs3);
+            Assert.IsTrue(dlSubscription.IsActive(subs1));
+        }
+
+        [TestMethod]
+        public void AttachNamyAndDetachActive()
```
Dmitry-Matveev: Minor: Many

Src/DependencyCollector/Shared/Implementation/ActiveSubsciptionManager.cs
```diff
@@ -0,0 +1,56 @@
...
+    using System.ComponentModel;
+    using System.Linq;
+    using System.Threading;
+
+    /// <summary>
+    /// Keeps one active subscription to specific DiagnosticSource per process.
+    /// Helps manage subsciptions in scenarios where multiple apps hosted in the same process.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public class ActiveSubsciptionManager
```
Dmitry-Matveev: Does it have to be public (even with `EditorBrowsableState.Never`)?

Src/DependencyCollector/Shared/HttpCoreDiagnosticSourceListener.cs
Dmitry-Matveev: Same configuration is a pretty heavy requirement if not for the nature of Service Fabric environment where you may want AI configuration out of the single host scope. This code, however, will run in non-service Fabric scenarios as well - do we have a scenario where a customer will encounter multiple listeners outside of SF? Will the requirement for same config still be fine in such a case?

Src/DependencyCollector/Shared/Implementation/ActiveSubsciptionManager.cs
lmolkova: yes, as I'm going to reuse it in AspNetCore SDK for request deduplication

PR #1016, opened 2018-09-27
consolidate test dependencies

MS-TimothyMothra:
I'm consolidating the dependencies of our functional tests.
Moving all the references to the BaseSDK to single files will simplify the updates necessary for monthly releases.

to validate these changes, i want to see that the # of passed tests match that of the develop branch

### DIFF
--- a/Test/Web/FunctionalTests/TestApps/WebAppFW45Sampled/WebAppFW45Sampled.csproj
+++ b/Test/Web/FunctionalTests/TestApps/WebAppFW45Sampled/WebAppFW45Sampled.csproj
@@ -22,7 +22,6 @@
     <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\..\</SolutionDir>
     <WcfConfigValidationEnabled>True</WcfConfigValidationEnabled>
     <TargetFrameworkProfile />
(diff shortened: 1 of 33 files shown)


Test/Web/FunctionalTests/TestApps/WebAppFW45Sampled/WebAppFW45Sampled.csproj
```diff
@@ -120,6 +112,9 @@
     <Content Include="packages.config">
       <SubType>Designer</SubType>
     </Content>
+    <None Include="$(SolutionDir)\..\test.common.packages.config">
```
SergeyKanzhelev: does it actually work? So two packages config will be used by single project?

MS-TimothyMothra: everything compiles and all the tests pass. that was my metric for success :)

SergeyKanzhelev: I don't understand how it can be. It may be build ordering thing when you got lucky. Hopefully those projects will be re-written on new csproj format and this will not be an issue any longer.

I'm ok to merge if you need this change.

PR #1024, opened 2018-10-02
Ignore unobserved task exceptions explicitly marked as observed

### LINKED ISSUE #1026: Allow users to ignore specific UnobservedTaskExceptions
Sometimes you encounter `UnobservedTaskException`s that are outside of your control (e.g. in third-party libraries or in the framework itself).
Currently you can either have them mixed in with the fixable `UnobservedTaskException`s which lowers signal-to-noise ratio, or you must remove the entire `UnobservedExceptionTelemetryModule`.

It would be useful to have some way to ignore specific `UnobservedTaskException`s without throwing the baby with the bathwater.

i3arnon:
Fixes #1026

This allows users to ignore specific exceptions (for example, from third party libraries) by subscribing to `UnobservedTaskException` and marking the exceptions as observed via `UnobservedTaskExceptionEventArgs.SetObserved`.

Without this users need to remove the entire `UnobservedExceptionTelemetryModule` in order to ignore floods of uninteresting exceptions outside their control.


SergeyKanzhelev:
Can you please create an issue with the description of the change and the use example. It will work as a documentation for the feature on GitHub,

Also - update CHANGELOG.md with the reference to that issue.

Thank you for contribution!

i3arnon:
@SergeyKanzhelev ping?

cijothomas:
@SergeyKanzhelev Can we merge this?

CHANGELOG.md
```diff
@@ -56,6 +56,7 @@
 	If you are upgrading, and have added/modified TelemetryProcessors, make sure to copy them to the default sink section.
 - [Microsoft.AspNet.TelemetryCorrelaiton package update to 1.0.4](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/991)
 - Add vmScaleSetName field to heartbeat properties collected by AzureInstanceMetadataTelemetryModule to allow navigation to right Azure VM Scale Set
+- [Allow users to ignore specific UnobservedTaskExceptions](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/1026)
```
SergeyKanzhelev: so the description say it "allows to ignore" when the code simply ignores by defult. Should there be a setting that will enable to restore the original behavior?

SergeyKanzhelev:
@cijothomas I left a comment. Please decide. Good feature, I would recommend to enhance it with the config setting to restore the original behavior.

cijothomas:
@i3arnon Can you make a public bool field on the module, and Ignore the exception based on its value? 
If you are not actively working on this, please let us know - I will merge this, as it is for now, and add the public bool field as a followup PR.

Thanks for your contribution!

CHANGELOG.md
i3arnon: @SergeyKanzhelev it only ignores exceptions explicitly marked as `Observed` by the user subscribing to the event and calling `SetObserved` for that exception. Other exceptions are treated as before..

SergeyKanzhelev: I was asking - if someone get used to the behavior when Observer exceptions are still being collected, this change will break the expectation. So I am suggesting to add a setting that will keep the previous behavior.

i3arnon: But there is one.. the `Observed` setting :)
If they previously observed the exception they probably didn't want another handler, and if they do, they can just not set it as observed anymore (which how the next handler is controlled).
Practically, this is such a niche edge case (even more than this one is) so I would leave that as is unless someone complains..

PR #1040, opened 2018-10-10
testing fix for signing

MS-TimothyMothra:
second attempt to fix signing..

- fix issue with DLLs not being signed. (manual import Microbuild.Core)
- fix issue with NUPKG not being signed. (manually import Microbuild.Plugins.Signing)

### DIFF
--- a/Src/DependencyCollector/DependencyCollector/DependencyCollector.csproj
+++ b/Src/DependencyCollector/DependencyCollector/DependencyCollector.csproj
@@ -1,4 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\..\..\packages\MicroBuild.Core.0.3.0\build\MicroBuild.Core.props" />
+  <Import Project="$(MicroBuildPluginDirectory)\MicroBuild.Plugins.Signing.1.1.16\build\MicroBuild.Plugins.Signing.props" Condition="Exists('$(MicroBuildPluginDirectory)\MicroBuild.Plugins.Signing.1.1.16\build\MicroBuild.Plugins.Signing.props')" />
+  
   <Import Project="..\..\..\Common.props" />
   <Import Project="..\..\..\Nupkg.props" />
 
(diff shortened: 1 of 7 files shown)


Src/DependencyCollector/DependencyCollector/DependencyCollector.csproj
```diff
@@ -1,4 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\..\..\packages\MicroBuild.Core.0.3.0\build\MicroBuild.Core.props" />
+  <Import Project="$(MicroBuildPluginDirectory)\MicroBuild.Plugins.Signing.1.1.16\build\MicroBuild.Plugins.Signing.props" Condition="Exists('$(MicroBuildPluginDirectory)\MicroBuild.Plugins.Signing.1.1.16\build\MicroBuild.Plugins.Signing.props')" />
```
Dmitry-Matveev: This fix will break as soon as plugin is updated to 1.1.17. We'd need to consider a more permanent fix by then.

Src/HostingStartup/HostingStartup/HostingStartup.csproj
```diff
@@ -1,4 +1,5 @@
 <Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\..\..\packages\MicroBuild.Core.0.3.0\build\MicroBuild.Core.props" />
```
Dmitry-Matveev: Do HostingStartup and Web need that fix? I remember they were importing those through common.

Src/DependencyCollector/DependencyCollector/DependencyCollector.csproj
MS-TimothyMothra: I believe we would be isolated from this... 1.1.17 is a dependency of MicroBuild.Core 0.3.0
When MicroBuild.Core gets updated, we would also have to fix this.
But i hope to have this problem resolved before then :)

Src/HostingStartup/HostingStartup/HostingStartup.csproj
MS-TimothyMothra: They work without the extra import. Very confusing.

Issue #988, opened 2018-09-07
Misspelling of activity name in ClientServerDependencyTracker


EvanMulawski:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/af20672f42145b837bf50b48e30e8427ca817e2d/Src/DependencyCollector/Shared/Implementation/ClientServerDependencyTracker.cs#L13

lmolkova:
thanks for the contribution!

PR #1044, opened 2018-10-11
Revert "Fixes misspelling #988"

MS-TimothyMothra:
Reverts Microsoft/ApplicationInsights-dotnet-server#992

Dev changes should not go against the MASTER branch

### DIFF
--- a/Src/DependencyCollector/Shared/Implementation/ClientServerDependencyTracker.cs
+++ b/Src/DependencyCollector/Shared/Implementation/ClientServerDependencyTracker.cs
@@ -10,7 +10,7 @@
 
     internal static class ClientServerDependencyTracker
     {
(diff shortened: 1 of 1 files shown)


Src/DependencyCollector/Shared/Implementation/ClientServerDependencyTracker.cs
```diff
@@ -10,7 +10,7 @@
 
     internal static class ClientServerDependencyTracker
     {
-        internal const string DependencyActivityName = "Microsoft.ApplicationInsights.Web.Dependency";
+        internal const string DependencyActivityName = "Microsoft.ApplivationInsights.Web.Dependency";
```
Dmitry-Matveev: Is there any code that already took a dependency on the previous Activity Name? (E.g. consumes activity and validates it against the misspelled name?) I assume not by looking at an `internal` in the definition.

Issue #1046, opened 2018-10-16
Hearbeat package from Azure App Service missing correct resource group name


iusafaro:
Heartbeat package generated from Azure App Service not necessarily has the correct information about resource group:
Current implementation relies on the format of the WEBSITE_OWNER_NAME environment variable exposed on App Service: <subId>+<resGroup>-<location>webspace, but there can be cases where <resGroup> is not the name of actual resource group. For instance names don't match after resource was moved to the different resource group, in this case WEBSITE_OWNER_NAME is not updated.

Recently App Service exposed different variable that we should add to heartbeat payload and use it for resource group identification: WEBSITE_RESOURCE_GROUP

iusafaro:
Added

PR #1047, opened 2018-10-16
Add WEBSITE_RESOURCE_GROUP to hearbeat properties

iusafaro:
Fix Issue #1046  .
App Services expose different variable that we should add to heartbeat payload and use it for resource group identification: WEBSITE_RESOURCE_GROUP

### DIFF
--- a/Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvironmentVariableMonitor.cs
+++ b/Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvironmentVariableMonitor.cs
@@ -27,7 +27,8 @@
             "WEBSITE_SLOT_NAME",
             "WEBSITE_HOME_STAMPNAME",
             "WEBSITE_HOSTNAME",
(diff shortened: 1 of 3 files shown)


Src/WindowsServer/WindowsServer.Shared/Implementation/AppServiceEnvironmentVariableMonitor.cs
```diff
@@ -27,7 +27,8 @@ internal sealed class AppServiceEnvironmentVariableMonitor : EnvironmentVariable
             "WEBSITE_SLOT_NAME",
             "WEBSITE_HOME_STAMPNAME",
             "WEBSITE_HOSTNAME",
-            "WEBSITE_OWNER_NAME"
+            "WEBSITE_OWNER_NAME",
+            "WEBSITE_RESOURCE_GROUP"
```
MS-TimothyMothra: Should we add comments to document where we expect these EnvVariables to be used?
Looking at this list, it would not be clear what applications or services support which variables.

Issue #1090, opened 2018-10-26
Sql dependency tracking broken in 2.8.0+. Dependency operation is not stopped and becomes parent of subsequent operations


ArturDorochowicz:
This is a bug in versions 2.8.0+ (checked 2.8.0 and 2.8.1). It does not happen in 2.7.2 (also checked 2.5.1).

## Repro Steps
Create a console application with code like this:
```csharp
    public class Sql
    {
        public static void Run()
        {
            var connString = @"Data Source=(LocalDb)\MSSQLLocalDB; Initial Catalog=master; Integrated Security=SSPI";

            for (var i = 0; i < 5; ++i)
            {
                ExecSql(i, connString);
            }
        }

        static void ExecSql(int i, string connectionString)
        {
            using (var conn = new SqlConnection(connectionString))
            {
                conn.Open();
                conn.GetSchema("Tables");
            }

            Console.WriteLine($"Activity id after sql {i}: {Activity.Current?.Id}");
        }
    }
```

Full repro solution attached.
[ConsoleApp1.zip](https://github.com/Microsoft/ApplicationInsights-dotnet-server/files/2520572/ConsoleApp1.zip)

##  Actual Behavior
When using 2.8.0 the app prints:
```
Activity id after sql 0: |47da6e291f386f42b6fb519694553b3c.dc340c80_
Activity id after sql 1: |47da6e291f386f42b6fb519694553b3c.dc340c80_1.
Activity id after sql 2: |47da6e291f386f42b6fb519694553b3c.dc340c80_1.1.
Activity id after sql 3: |47da6e291f386f42b6fb519694553b3c.dc340c80_1.1.1.
Activity id after sql 4: |47da6e291f386f42b6fb519694553b3c.dc340c80_1.1.1.1.
```

Note that activity remains present after running a sql dependency and, due to this, the next operation (here another sql dependency) becomes its child. It continues for subsequent operations.

## Expected Behavior
When using 2.7.2 the app prints:
```
Activity id after sql 0:
Activity id after sql 1:
Activity id after sql 2:
Activity id after sql 3:
Activity id after sql 4:
```

This is the expected output, i.e. there's no activity left over after running a sql dependency.

## Version Info 
SDK Version : 2.8.0, also 2.8.1
.NET Version :	net framework 4.6.1, also 4.7.1	
How Application was onboarded with SDK(VisualStudio/StatusMonitor/Azure Extension) : Visual Studio
[...]

PR #1091, opened 2018-10-27
Fix sql activity not stopped by EventSource Sql listener

lmolkova:
Fix Issue #1090 .

### DIFF
--- a/Src/DependencyCollector/Shared/Implementation/FrameworkSqlProcessing.cs
+++ b/Src/DependencyCollector/Shared/Implementation/FrameworkSqlProcessing.cs
@@ -1,8 +1,7 @@
-﻿using System.Diagnostics;
-
-namespace Microsoft.ApplicationInsights.DependencyCollector.Implementation
+﻿namespace Microsoft.ApplicationInsights.DependencyCollector.Implementation
 {
     using System;
+    using System.Diagnostics;
     using System.Globalization;
     using Microsoft.ApplicationInsights.Common;
     using Microsoft.ApplicationInsights.DataContracts;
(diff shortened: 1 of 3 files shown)


Src/DependencyCollector/Shared/Implementation/FrameworkSqlProcessing.cs
```diff
@@ -59,19 +60,26 @@ public void OnBeginExecuteCallback(long id, string dataSource, string database,
...
                 if (telemetryTuple == null)
                 {
-                    bool isCustomCreated = false;
                     var telemetry = ClientServerDependencyTracker.BeginTracking(this.telemetryClient);
                     telemetry.Name = resourceName;
                     telemetry.Target = string.Join(" | ", dataSource, database);
                     telemetry.Type = RemoteDependencyConstants.SQL;
                     telemetry.Data = commandText;
-                    this.TelemetryTable.Store(id, new Tuple<DependencyTelemetry, bool>(telemetry, isCustomCreated));
+                    this.TelemetryTable.Store(id, new Tuple<DependencyTelemetry, bool>(telemetry, false));
```
MS-TimothyMothra: is this boolean value ever NOT false?
I only see one example below of an if statement checking this value.

lmolkova: this value was explicitly set to false few lines above, so it has always been false. Resharper is so nice, it tells me about that ;)

lmolkova: https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared/Implementation/FrameworkSqlProcessing.cs#L62

MS-TimothyMothra: What I meant was, do we need it if it never changes? :)

lmolkova: Oh, sorry. Now i get your question.
It's not used in sql processing, but we store telemetry items in `CacheBasedOperationHolder` between start and stop. And We also use this class for http processing, where `isCustomCreated` was needed at least some time ago. 
This class requires us to have a tuple of `<DependencyTelemetry, bool>`.
It is a good idea to refactor it and get rid of the flag (I'm almost sure it's not used anymore), but I'd prefer not to do it in the scope of this PR.

PR #1119, opened 2018-11-30
add postbuild targets

MS-TimothyMothra:
Fix: https://github.com/Microsoft/ApplicationInsights-dotnet/issues/1021

### DIFF
--- a/Src/DependencyCollector/DependencyCollector/Microsoft.ApplicationInsights.DependencyCollector.targets
+++ b/Src/DependencyCollector/DependencyCollector/Microsoft.ApplicationInsights.DependencyCollector.targets
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <Target Name="CopyApplicationInsightsConfigToOutputDirectory" AfterTargets="CopyFilesToOutputDirectory" >
+    <!--Currently we cannot set the config's "Copy to Output Directory" property to "PreserveNewest.
+          Workaround is to use this targets file to manually copy into output path."-->
+
+    <Copy SourceFiles="ApplicationInsights.config" DestinationFolder="$(OutputPath)" ContinueOnError="true" />
+    <Message Text="Microsoft.ApplicationInsights.DependencyCollector: ApplicationInsights.config has been copied to $(OutputPath)" Importance="high" />
+  </Target>
+</Project>
--- a/Src/DependencyCollector/DependencyCollector/Microsoft.ApplicationInsights.DependencyCollector.targets
+++ b/Src/DependencyCollector/DependencyCollector/Microsoft.ApplicationInsights.DependencyCollector.targets
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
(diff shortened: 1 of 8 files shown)


pharring:
Have you tried using a .props file (instead of .targets) and setting the property there instead?

MS-TimothyMothra:
@pharring I didn't realize that property would be accessible in a .props file.
I found an example [here](https://stackoverflow.com/questions/21143817/set-content-files-to-copy-local-always-in-a-nuget-package/30386836#30386836). I'll test it this afternoon.

Src/DependencyCollector/DependencyCollector/Microsoft.ApplicationInsights.DependencyCollector.targets
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <Target Name="CopyApplicationInsightsConfigToOutputDirectory" AfterTargets="CopyFilesToOutputDirectory" >
```
SergeyKanzhelev: can you please set affected files as a property of this target

Src/DependencyCollector/DependencyCollector/Microsoft.ApplicationInsights.DependencyCollector.targets
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
SergeyKanzhelev: are we forced to specify ToolsVersion here?

Src/DependencyCollector/DependencyCollector/Microsoft.ApplicationInsights.DependencyCollector.targets
SergeyKanzhelev: Is `CopyFilesToOutputDirectory` executes for console apps?

MS-TimothyMothra: sorry, i don't understand what you're asking here

MS-TimothyMothra: I can test specifically with console apps. Would we need to take any action if `CopyFilesToOutputDirectory` is not executed?

MS-TimothyMothra: I confirmed that this is executed for Console apps.

I would argue though that we would be unaffected by any app type that does not copy files to an output directory. :)

Src/DependencyCollector/DependencyCollector/Microsoft.ApplicationInsights.DependencyCollector.targets
MS-TimothyMothra: removing

MS-TimothyMothra:
@pharring I haven't been able to find a setting that modifies the file's properties. 
If you know of one, please do share!

MSBuild Tasks can let me execute scripts to do this, but I would need time to investigate a script. 
I'm not sure if this is worth the time right now.

pharring:
Does the following work?
```xml
<Project ToolsVersion="15.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <ItemGroup>
    <Content Include="ApplicationInsights.config">
    <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
    </Content>
  </ItemGroup>
</Project>
```

PR #1136, opened 2019-01-22
remove reference to TelemetryCorrelation package from DependencyCollector

lmolkova:
Reference was added by accident in https://github.com/Microsoft/ApplicationInsights-dotnet-server/commit/fed761e3710f39089d201a85da1300030becd82d

DepednencyCollector does not use anything from the Microsoft.AspNet.TelemetryCorrelation package.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## Version 2.9.0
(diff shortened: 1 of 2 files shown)


MS-TimothyMothra:
thanks

PR #1138, opened 2019-01-24
Move W3C classes to base SDK so everyone can use them

lmolkova:
W3C implementation is used by Web and AspNetCore SDKs.

We are going to enable W3C for some 1DS users that do not necessarily use Web SDK (and reference DependencyCollector that declares W3C implementation).

Without this change, we would require everyone to install base + dependency collector just to onboard W3C. With this change, it is still an explicit choice to enable it.

W3C is moved to base, Implementation in dependency collector is deprecated with error.


Src/Common/StringUtilities.cs
```diff
@@ -47,19 +42,21 @@ public static string EnforceMaxLength(string input, int maxLength)
         /// https://github.com/w3c/distributed-tracing/blob/master/trace_context/HTTP_HEADER_FORMAT.md#trace-id
         /// </summary>
         /// <returns>Random 16 bytes array encoded as hex string</returns>
+        [Obsolete("Use Microsoft.ApplicationInsights.Extensibility.W3C.W3CUtilities.GenerateTraceId instead.", true)]
```
Dmitry-Matveev: Will it break compilation? It can be fine if we didn't have this feature in any stable (I remember we didn't), so migrating between stables won't suddenly stop compiling, right?

Src/Common/W3C/W3CActivityExtensions.cs
```diff
@@ -1,284 +1,114 @@
...
-        private static readonly Regex TraceIdRegex = new Regex("^[a-f0-9]{32}$", RegexOptions.Compiled);
-        private static readonly Regex SpanIdRegex = new Regex("^[a-f0-9]{16}$", RegexOptions.Compiled);
-
         /// <summary>
         /// Generate new W3C context.
         /// </summary>
         /// <param name="activity">Activity to generate W3C context on.</param>
         /// <returns>The same Activity for chaining.</returns>
-        [Obsolete("Not ready for public consumption.")]
+        [Obsolete("Microsoft.ApplicationInsights.Extensibility.W3C.W3CActivityExtensions.GenerateW3CContext instead.", true)]
```
Dmitry-Matveev: You may need to add "Use" here and in the other cases below.

Src/Common/W3C/W3CUtilities.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    using System;
+    using System.ComponentModel;
+    using System.Globalization;
+
+    /// <summary>
+    /// W3C distributed tracing utilities.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public class W3CUtilities
```
Dmitry-Matveev: Do we need this public? Removal will be hard once we ship in stable.

Src/Common/StringUtilities.cs
cijothomas: I'd also mention the name of the nuget package as well  Microsoft.ApplicationInsights

Src/Common/W3C/W3CUtilities.cs
lmolkova: sorry, file was added by mistake

PR #1140, opened 2019-01-29
Prepare develop for stable 2.9.0

Dmitry-Matveev:
Merging into develop first. @lmolkova mentioned that I'd need to pick all develop commits into master anyway.

### DIFF
--- a/Nupkg.props
+++ b/Nupkg.props
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <PropertyGroup>
+  <PropertyGroup Condition="$(OS) == 'Windows_NT'">
     <!-- Including this file will generate both the *.nupkg and *.symbols.nupkg -->
     <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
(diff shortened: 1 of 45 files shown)


Nupkg.props
```diff
@@ -15,7 +15,7 @@
     <Authors>Microsoft</Authors>
     <Owners>Microsoft,AppInsightsSdk</Owners>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
-    <PackageLicenseUrl>https://go.microsoft.com/fwlink/?LinkID=510709</PackageLicenseUrl>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageProjectUrl>https://go.microsoft.com/fwlink/?LinkId=392727</PackageProjectUrl>
```
MS-TimothyMothra: going to want that same WindowsNT check on line 3.

Dmitry-Matveev: I thought we do not have Linux builds for web. Let's update right now anyway for the future gains.

Dmitry-Matveev:
Changed Build Pipeline to use .NET Core 2.1.500 tools (should understand LicenseExpression vs. LicenseUrl)

PR #1161, opened 2019-03-21
 A telemetry initializer that will gather configuration information.

paulomorgado:
A telemetry initializer that will gather configuration information.

### DIFF
--- a/Src/WindowsServer/WindowsServer.Shared.Tests/ConfigurationTelemetryInitializerTest.cs
+++ b/Src/WindowsServer/WindowsServer.Shared.Tests/ConfigurationTelemetryInitializerTest.cs
@@ -1,0 +1,196 @@
+﻿namespace Microsoft.ApplicationInsights.WindowsServer
+{
+    using System;
+    using System.Globalization;
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+#if !NETCORE
+    using Microsoft.ApplicationInsights.WindowsServer.Azure;
+    using Microsoft.ApplicationInsights.WindowsServer.Azure.Emulation;
+#endif
+    using Microsoft.ApplicationInsights.WindowsServer.Implementation;
+
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using Assert = Xunit.Assert;
+
+    [TestClass]
+    public class ConfigurationTelemetryInitializerTest
+    {
+        private const string TestRoleName = nameof(TestRoleName);
+        private const string TestRoleInstance = nameof(TestRoleInstance);
+
+        [TestMethod]
+        public void ConfigurationTelemetryInitializerSetsRoleName()
+        {
+            var telemetryItem = new EventTelemetry();
+
+            var initializer = new ConfigurationTelemetryInitializer
+            {
+                RoleName = TestRoleName,
+            };
+
+            initializer.Initialize(telemetryItem);
+
+            Assert.Equal(TestRoleName, telemetryItem.Context.Cloud.RoleName);
+        }
+
+        [TestMethod]
+        public void ConfigurationTelemetryInitializerSetsRoleNameFromEnvironmentVariable()
+        {
+            string roleNameTestVarName, roleInstanceTestVarName;
+            GetAndInitializeEnvironment(out roleNameTestVarName, out roleInstanceTestVarName);
+
+            var telemetryItem = new EventTelemetry();
+
+            var initializer = new ConfigurationTelemetryInitializer
+            {
+                RoleName = $"%{roleNameTestVarName}%",
+            };
+
+            initializer.Initialize(telemetryItem);
+
+            Assert.Equal(TestRoleName, telemetryItem.Context.Cloud.RoleName);
+
+            ClearEnvironmentVariables(roleNameTestVarName, roleInstanceTestVarName);
+        }
+
+        [TestMethod]
+        public void ConfigurationTelemetryInitializerSetsRoleInstanceNameAndNodeName()
+        {
+            string roleNameTestVarName, roleInstanceTestVarName;
+            GetAndInitializeEnvironment(out roleNameTestVarName, out roleInstanceTestVarName);
+
+            var telemetryItem = new EventTelemetry();
+
+            var initializer = new ConfigurationTelemetryInitializer
+            {
+                RoleInstance = TestRoleInstance,
+            };
+
+            initializer.Initialize(telemetryItem);
+
+            Assert.Equal(TestRoleInstance, telemetryItem.Context.Cloud.RoleInstance);
+            Assert.Equal(TestRoleInstance, telemetryItem.Context.GetInternalContext().NodeName);
+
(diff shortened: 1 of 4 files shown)


Src/WindowsServer/WindowsServer.Shared/ConfigurationTelemetryInitializer.cs
```diff
@@ -0,0 +1,114 @@
...
+        public void Initialize(ITelemetry telemetry)
+        {
+            if (this.updateEnvVars)
+            {
+                this.roleName = this.GetRoleName();
+                this.roleInstanceName = this.GetRoleInstanceName();
+                this.updateEnvVars = false;
+            }
+
+            if (this.roleName is string && string.IsNullOrEmpty(telemetry.Context.Cloud.RoleName))
```
pharring: Can `this.roleName is string` ever be false if it's declared as a string?

Src/WindowsServer/WindowsServer.Shared/ConfigurationTelemetryInitializer.cs
```diff
@@ -0,0 +1,114 @@
...
+        private string GetRoleInstanceName() => ExpandEnvironmentVariables(this.RoleInstance);
+
+        private static string ExpandEnvironmentVariables(string definition)
+        {
+            if (string.IsNullOrWhiteSpace(definition))
+            {
+                return default(string);
+            }
+
+            var expandedDefinition = Regex.Replace(definition, @"(?<e>%%)|%(?<v>[^%]+)%", ExpandEnvironmentVariable, RegexOptions.ExplicitCapture);
```
pharring: This deserves a comment (because reg-exps are hard to read). What does the replacement do?

Src/WindowsServer/WindowsServer.Shared.Tests/ConfigurationTelemetryInitializerTest.cs
```diff
@@ -0,0 +1,196 @@
...
+            var initializer = new ConfigurationTelemetryInitializer();
+
+            initializer.Initialize(telemetryItem);
+
+            Assert.Null(telemetryItem.Context.Cloud.RoleName);
+            Assert.Null(telemetryItem.Context.Cloud.RoleInstance);
+            Assert.Null(telemetryItem.Context.GetInternalContext().NodeName);
+        }
+
+        private static void GetAndInitializeEnvironment(out string roleNameTestVarName, out string roleInstanceTestVarName)
```
pharring: Consider returning an `IDisposable` that does the clear. Then callers can use `using` to ensure clean-up even when exceptions are thrown.

Src/WindowsServer/WindowsServer.Shared/ConfigurationTelemetryInitializer.cs
```diff
@@ -0,0 +1,114 @@
...
+using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.ApplicationInsights.Extensibility.Implementation;
+using Microsoft.ApplicationInsights.WindowsServer.Implementation;
+
+namespace Microsoft.ApplicationInsights.WindowsServer
+{
+    /// <summary>
+    /// A telemetry initializer that will gather configuration information.
+    /// </summary>
+    public class ConfigurationTelemetryInitializer : ITelemetryInitializer
```
pharring: "Configuration" is an overloaded term. What configuration is this? Perhaps it should be called `CloudRoleInitializer` or something along those lines.

Src/WindowsServer/WindowsServer.Shared/ConfigurationTelemetryInitializer.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        private string roleName;
+        private string roleInstanceName;
+        private volatile bool updateEnvVars = true;
```
pharring: `volatile` is probably overkill since there's no interlocking on this value.
Also, suggest reversing the 'polarity' so that it can be initialized to `false` (which is the default) and set to `true` when values have been read.

Src/WindowsServer/WindowsServer.Shared/ConfigurationTelemetryInitializer.cs
pharring: Is this just shorthand for a (not-)null check? If so, `!= null` seems more direct (not sure if the C# compiler/JIT will _actually,_ do a type check), clear and concise.

Src/WindowsServer/WindowsServer.Shared/ConfigurationTelemetryInitializer.cs
pharring: Also, what's wrong with `Environment.ExpandEnvironmentVariables`?

Src/WindowsServer/WindowsServer.Shared/ConfigurationTelemetryInitializer.cs
paulomorgado: I this case, it's more a matter of style.

But a variable of type `string` can be `null` which is not a `string`.

A construct like `stringThatMayBeNill is string stringThatIsDefinitelyNotNull` guarantees that `stringThatIsDefinitelyNotNull` is definitely not `null`.

This does not happen with `string` but any type can redefine the `==` and `!=` operators and the `true` and `false` operators in a way that `value == null` is not guaranteed `value` is actually `null`. On the other hand, the compiler guarantees that `value is null` will be resolved as `object.ReferenceEquals(value, null)`.

Src/WindowsServer/WindowsServer.Shared/ConfigurationTelemetryInitializer.cs
paulomorgado: I can add a comment, if you want to.

I have nothing against `Environment` anything. That's why I don't understand why `AppServiceEnvironmentVariableMonitor` is being used.

Src/WindowsServer/WindowsServer.Shared/ConfigurationTelemetryInitializer.cs
paulomorgado: Naming is one of the hardest tasks in software development. `CloudRoleInitializer`? Since it can also initialize the role instance.

Src/WindowsServer/WindowsServer.Shared/ConfigurationTelemetryInitializer.cs
paulomorgado: Totally agree. I just followed the implementation of [`AzureWebAppRoleEnvironmentTelemetryInitializer`](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs#L26).

Src/WindowsServer/WindowsServer.Shared.Tests/ConfigurationTelemetryInitializerTest.cs
paulomorgado: Don't you think that's overkill? But using a `try`/`finally` like other tests would be a good idea.

PR #1172, opened 2019-04-16
Performance: mikp's new class "PostSamplingTelemetryProcessor" for attach scenarios

MS-TimothyMothra:
The private fork of our SDK is causing me some issues.
I'm trying to get as many of these changes into the actual SDK as possible.

- introduces new telemetry processor: `PostSamplingTelemetryProcessor`
  - if RequestTelemetry.Url is null, set to context.Request's unvalidated url.
  - will set Request.Source to the source's appId.
  - Both of these behaviors exist in 'RequestTrackingTelemetryModule.cs', but have been disabled in favor of being applied after sampling
- This class will **not** be added to the default SDK and will not affect SDK customers.
- This class will **only** be used by attach scenarios.

May be more appropriate to rename this class `PostSamplingRequestTrackingTelemetryProcessor` because it specifically deferrs behavior from the `RequestTrackingTelemetryModule` to execute after Sampling.

### DIFF
--- a/Src/Web/Web.Shared.Net/PostSamplingTelemetryProcessor.cs
+++ b/Src/Web/Web.Shared.Net/PostSamplingTelemetryProcessor.cs
@@ -1,0 +1,77 @@
+﻿namespace Microsoft.ApplicationInsights.Web
+{
+    using System;
+
+    using Microsoft.ApplicationInsights.Channel;
+    using Microsoft.ApplicationInsights.Common;
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+    using Microsoft.ApplicationInsights.Web.Implementation;
+
+    /// <summary>
+    /// PostSamplingTelemetry processor which ensures that the RequestTelemetry.Url is initialized
+    /// in the context of HttpContext.Current.Request
+    /// </summary>
+    public class PostSamplingTelemetryProcessor : ITelemetryProcessor
+    {
+        private readonly ITelemetryProcessor nextProcessorInPipeline;
+        private TelemetryConfiguration telemetryConfiguration;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PostSamplingTelemetryProcessor"/> class.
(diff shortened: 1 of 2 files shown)


Src/Web/Web.Shared.Net/PostSamplingTelemetryProcessor.cs
```diff
@@ -0,0 +1,77 @@
...
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+    using Microsoft.ApplicationInsights.Web.Implementation;
+
+    /// <summary>
+    /// PostSamplingTelemetry processor which ensures that the RequestTelemetry.Url is initialized
+    /// in the context of HttpContext.Current.Request
+    /// </summary>
+    public class PostSamplingTelemetryProcessor : ITelemetryProcessor
```
cijothomas: These are public classes, and will be released as part of regular nugets, so lets ship only those things which we really want. Its not easy to remove once shipped.

MS-TimothyMothra: If it's good enough for Attach customers, it should be good enough for all SDK customers  :)

What would be required for this to be a part of the full SDK?

cijothomas: URL is already populated by RequestCollectionModules which is shipped officially.
Modifying existing RequestCollectionModules to NOT populate URL, and instead rely on this new TP is a behavior change- Tel.Initializers will not see the URL. And also any TP added before this TP wont see URL. And there is no guarantee that TP is run in same context as incoming Request, so HttpContext.Current could be null.

We can discuss such change, but i'd do it only when we are changing major version.

MS-TimothyMothra: That's a good point. I'm going to close this PR and open an issue to discuss further.

MS-TimothyMothra: https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/1173

Issue #1173, opened 2019-04-16
Performance Enhancement: Request Tracking Properties & Post Sampling Execution Deferral 


MS-TimothyMothra:
@Mikhail-Pranovich identified performance issues in RequestTrackingTelemetryModule.
Specifically:

1. Setting the `requestTelemetry.Url` from the context
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/3bee86d3f7ec3e81efe6c5ad3ebdbe37f79929bb/Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs#L164-L167

2. Setting the `requestTelemetry.Source` to the SourceAppId
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/3bee86d3f7ec3e81efe6c5ad3ebdbe37f79929bb/Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs#L176-L203

**Proposal**
1. I'm proposing a new property: `RequestTrackingTelemetryModule.DisableTrackingProperties` (open to naming suggestions)

```
[EditorBrowsable(EditorBrowsableState.Never)]
public bool DisableTrackingProperties {get; set;}
```

This would disable setting this properties in the RequestTrackingTelemetryModule.

2. Settings these properties would be deferred to a TelemetryProcessor: `PostSamplingRequestTrackingTelemetryModule`.
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/8d10958ca6e7e3d6130b8d565851a927de6e975b/Src/Web/Web.Shared.Net/PostSamplingTelemetryProcessor.cs#L16-L76

**Risks**
@cijothomas had a good comment:

> URL is already populated by RequestCollectionModules which is shipped officially.
> Modifying existing RequestCollectionModules to NOT populate URL, and instead rely on this new TP is a behavior change- Tel.Initializers will not see the URL. And also any TP added before this TP wont see URL. And there is no guarantee that TP is run in same context as incoming Request, so HttpContext.Current could be null.
> 
> We can discuss such change, but i'd do it only when we are changing major version.

**Open Question**
- What changes would be required to take this change as part of the official SDK?
  - Attach wants these changes already, and cannot wait for 3.0

MS-TimothyMothra:
I'm shocked that populating the `Request.Unvalidated.Url` adds a significant performance overhead.

I also see that removing it adds additional side effects...

[...]

PR #1174, opened 2019-04-16
Mikp/web/private perf build

MS-TimothyMothra:
Fix Issue #1173 

- removes `RequestTelemetry.Url` from `RequestTrackingTelemetryModule.
- removes `RequestTelemetry.Source` from `RequestTrackingTelemetryModule.
- Adds new `PostSamplingTelemetryProcessor` to populate the previously removed properties. (Defered execution)
- restores `RequestTelemetry.Url` to `QuickPulseTelemetryProcessor`


Src/Web/Web.Shared.Net/PostSamplingTelemetryProcessor.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        private TelemetryConfiguration TelemetryConfiguration => this.telemetryConfiguration ?? (this.telemetryConfiguration = TelemetryConfiguration.Active);
+
+        /// <inheritdoc />
+        public void Process(ITelemetry item)
+        {
+            if (item is RequestTelemetry requestTelemetry)
+            {
+                var context = System.Web.HttpContext.Current;
```
MS-TimothyMothra: should protect this with a null check

Src/Web/Web.Shared.Net/PostSamplingTelemetryProcessor.cs
```diff
@@ -0,0 +1,77 @@
+﻿namespace Microsoft.ApplicationInsights.Web
```
cijothomas: can we use a namespace like ....extensibility.implementation so that we can change/remove it if needed, without changing major version?
@SergeyKanzhelev knows best about this.

Src/Web/Web.Shared.Net/PostSamplingTelemetryProcessor.cs
```diff
@@ -0,0 +1,77 @@
...
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+    using Microsoft.ApplicationInsights.Web.Implementation;
+
+    /// <summary>
+    /// PostSamplingTelemetry processor which ensures that the RequestTelemetry.Url is initialized
+    /// in the context of HttpContext.Current.Request
+    /// </summary>
+    public class PostSamplingTelemetryProcessor : ITelemetryProcessor
```
cijothomas: lets comment here to indicate this is not to be used by customers, unless they know the implications.

MS-TimothyMothra:
@Dmitry-Matveev can I get you to review this?
If you're okay with these changes, I'll release **alpha5** today

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Version 2.10.0-alpha5
```
SergeyKanzhelev: do we need this note? I thought the idea is to bring features to the `master`

Src/PerformanceCollector/Perf.Shared/QuickPulseTelemetryProcessor.cs
```diff
@@ -180,6 +180,13 @@ public void Process(ITelemetry telemetry)
 
         private static ITelemetryDocument ConvertRequestToTelemetryDocument(RequestTelemetry requestTelemetry)
         {
+#if NET45
+            if (requestTelemetry.Url == null && System.Web.HttpContext.Current != null)
```
SergeyKanzhelev: will this method be called sync or async? I thought there is a separate thread for this conversion

Src/PerformanceCollector/Perf.Shared/QuickPulseTelemetryProcessor.cs
```diff
@@ -180,6 +180,13 @@ public void Process(ITelemetry telemetry)
...
+            {
+                requestTelemetry.Url = System.Web.HttpContext.Current.Request.Unvalidated.Url;
```
SergeyKanzhelev: I'd suggest to keep it in local var for `telemetryDocument`, not mutate the RequestTelemetry. So there will not be a sudden behavior change when live stream UI is opened

Src/Web/Web.Shared.Net/Extensibility/Implementation/PostSamplingTelemetryProcessor.cs
```diff
@@ -0,0 +1,81 @@
...
+
+    using Microsoft.ApplicationInsights.Channel;
+    using Microsoft.ApplicationInsights.Common;
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+    using Microsoft.ApplicationInsights.Web.Implementation;
+
+    /// <summary>
+    /// [This feature is still being evaluated and not recommended for end users.]
```
SergeyKanzhelev: put this into `<remark>` section, not on the top of description

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -52,6 +54,17 @@ public class RequestTrackingTelemetryModule : ITelemetryModule
         /// </summary>
         public bool EnableAccessControlExposeHeader { get; set; } = true;
 
+        /// <summary>
+        /// Gets or sets a value indicating whether requestTelemetry.Url and requestTelemetry.Source are disabled.
```
SergeyKanzhelev: So cookies for user id and session id are cheap to collect?

CHANGELOG.md
MS-TimothyMothra: This was specifically to allow me to release a forked SDK alongside the stable. 
But if we can get this change into `master` that is preferred.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
Mikhail-Pranovich: At least it is not in current testing methodology we use. We can take a look into measuring this as a next step.

Src/PerformanceCollector/Perf.Shared/QuickPulseTelemetryProcessor.cs
Mikhail-Pranovich: It seems to be run in sync with all TelemetryProcessors.Process. @tokaplan , could you please confirm?

Src/PerformanceCollector/Perf.Shared/QuickPulseTelemetryProcessor.cs
Mikhail-Pranovich: TelemetryProcessors are expected to mutate the state. Additional copy might be expensive and error prone. I'd live it as is.

Src/Web/Web.Shared.Net/Extensibility/Implementation/PostSamplingTelemetryProcessor.cs
Mikhail-Pranovich: done

Src/PerformanceCollector/Perf.Shared/QuickPulseTelemetryProcessor.cs
tokaplan: @Mikhail-Pranovich confirmed, this conversion is done sequentially, there is no separate thread.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## Version 2.10.0-beta5
```
cijothomas: beta5 is not planned.

cijothomas:
@MS-TimothyMothra Is this how its going to work:
1. We merge this changes in main branch and ship it.
2. The default behavior wont be affected for SDK customers.
3. In Code-Less attach, via config, PostSampling. etc will be enabled.

Src/PerformanceCollector/Perf.Shared/QuickPulseTelemetryProcessor.cs
```diff
@@ -178,8 +178,15 @@ public void Process(ITelemetry telemetry)
             }
         }
 
-        private static ITelemetryDocument ConvertRequestToTelemetryDocument(RequestTelemetry requestTelemetry)
+        private static ITelemetryDocument ConvertRequestToTelemetryDocument(RequestTelemetry requestTelemetry, IApplicationIdProvider applicationIdProvider)
         {
+#if NET45
+            // some of the requestTelemetry properties might be deffered by using RequestTrackingTelemetryModule.DisableTrackingProperties.
+            // evaluate them now:
+            var request = System.Web.HttpContext.Current?.Request;
```
Dmitry-Matveev: There were come cases when HttpContext.Current is not available or lost in the process of the Http Request execution. Do any of them apply here? We simply return if Context is null - any downsides of that?

This also adds a dependency for Telemetry Processors to run synchronously with the request (as they do now) due to the use of Current, which was not the case before. If we decide to move Processors to async execution in 3.0, it may break. Arguably, 3.0 may have another approach to involve sampling earlier anyway, so this won't be needed...

Src/Web/Web.Shared.Net/Extensibility/Implementation/PostSamplingTelemetryProcessor.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        private TelemetryConfiguration TelemetryConfiguration => this.telemetryConfiguration ?? (this.telemetryConfiguration = TelemetryConfiguration.Active);
+
+        /// <inheritdoc />
+        public void Process(ITelemetry item)
+        {
+            if (item is RequestTelemetry requestTelemetry)
+            {
+                var request = HttpContext.Current?.Request;
```
lmolkova: No reason to believe HttpContext.Current is available when processors run. One async call in our pipeline is enough to break it, also sync processors execution may (and should) change.

lmolkova:
In the long term (second half of 2019) this problem is intended to be solved with sampling on Activity level: https://github.com/dotnet/corefx/issues/36349

We need to make sampling decision on request when we *start* processing it and set it on Acitivity.Current.Recording. Then we can always rely on Activity.Current.Recording when we need to set expensive properties.

In this case there is no need in extra flags and telemetry module, but we might need to expose sampling APIs in base SDK. Also it does not seem right to use HttpContext.Current in processors which make the whole approach error-prone and unreliable.

If this change is not time critical I propose to postpone it and make things better without introducing new things that also subject to break.

Src/Web/Web.Shared.Net/Extensibility/Implementation/PostSamplingTelemetryProcessor.cs
Mikhail-Pranovich: Correct, RequestTrackingUtilities.UpdateRequestTelemetryFromRequest handles that.

Src/PerformanceCollector/Perf.Shared/QuickPulseTelemetryProcessor.cs
Mikhail-Pranovich: There are such potential cases with design change. The more async handling should be considered along with sampling redesign, where it happens before any expensive operations.

MS-TimothyMothra:
@cijothomas Yes, that is the plan. 
I already have a branch with the config required for codeless attach.

@lmolkova This is to fix a problem we're actively having.
Codeless attach is using a forked version of the SDK which is creating a maintenance nightmare.
We need these changes in the SDK and hidden behind a feature flag.

CHANGELOG.md
Mikhail-Pranovich: changed

CHANGELOG.md
Mikhail-Pranovich: changed to vnext

Src/PerformanceCollector/Perf.Shared/QuickPulseTelemetryProcessor.cs
```diff
@@ -463,7 +470,7 @@ private void ProcessTelemetry(ITelemetry telemetry)
                             documentStreams,
                             documentStream => documentStream.RequestQuotaTracker,
                             documentStream => documentStream.CheckFilters(telemetryAsRequest, out groupErrors),
-                            ConvertRequestToTelemetryDocument);
+                            requestTelemetry => ConvertRequestToTelemetryDocument(requestTelemetry, this.config?.ApplicationIdProvider));
```
SergeyKanzhelev: let's not mutate the original request telemetry. The logic with post-processing by itself already hard to understand. It is harder and more error-prone when opening live stream UI may produce different results than when it's not opened.

SergeyKanzhelev: For this reply: 
image

I only suggest to do it for the url population. There will be no copy involved

Src/PerformanceCollector/Perf.Shared/QuickPulseTelemetryProcessor.cs
SergeyKanzhelev: Can `System.Web.HttpContext.Current` throw when not in ASP.NET context? If so - please add try/catch so console apps wouldn't blow out. I'd suggest to do it anyway as I believe this call is protected with some security demands which may not be met on some console apps

Src/PerformanceCollector/PerformanceCollector/Perf.csproj
```diff
@@ -44,6 +44,7 @@
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.5.0" />
     <!--Framework References-->
     <Reference Include="System.Runtime.Caching" />
+    <Reference Include="System.Web" />
```
SergeyKanzhelev: I don't remember whether I asked before. Can this be implemented as some sort of quick pulse extensibility point to eliminate this reference?

SergeyKanzhelev: perhaps a sort of a callback?

Src/Web/Web.Shared.Net/Extensibility/Implementation/PostSamplingTelemetryProcessor.cs
SergeyKanzhelev: nit: You can check that `if (requestTelemetry.Url == null)` before getting the current context. It will be a common mistake to include both processors

Src/Web/Web.Shared.Net.Tests/RequestTrackingTelemetryModuleTest.cs
```diff
@@ -453,7 +453,7 @@ private TelemetryConfiguration CreateDefaultConfig(HttpContext fakeContext, stri
                 InstrumentationKey = TestInstrumentationKey1,
                 ApplicationIdProvider = new MockApplicationIdProvider(TestInstrumentationKey1, TestApplicationId1)
             };
-            configuration.TelemetryInitializers.Add(new Extensibility.OperationCorrelationTelemetryInitializer());
+            configuration.TelemetryInitializers.Add(new Microsoft.ApplicationInsights.Extensibility.OperationCorrelationTelemetryInitializer());
```
SergeyKanzhelev: should both cases be tested?

Src/Web/Web.Shared.Net/Extensibility/Implementation/PostSamplingTelemetryProcessor.cs
```diff
@@ -0,0 +1,45 @@
...
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.Extensibility;
+
+    /// <summary>
+    /// PostSamplingTelemetryProcessor evaluates deferred properties.
+    /// It is intended to be used with <see cref="RequestTrackingTelemetryModule.DisableTrackingProperties"/> after Sampling.
+    /// </summary>
+    /// <remarks>
+    /// This feature is still being evaluated and not recommended for end users.
+    /// This setting is not browsable at this time.
```
SergeyKanzhelev: comment out of place.

lmolkova:
> This is to fix a problem we're actively having. Codeless attach is using a forked version of the SDK which is creating a maintenance nightmare. We need these changes in the SDK and hidden behind a feature flag.

My point is, relying on HttpContext.Current is not a good idea and I'm afraid we are moving nightmare inside the SDK and exposing feature flags/APIs that are not going to be used in the long term, but will have to maintain them forever. Have we evaluated options to avoid this?

MS-TimothyMothra:
> Have we evaluated options to avoid this?

@lmolkova We're very open to suggestions, but I don't think waiting until the end of this year is good enough. 

Yes, HttpContext.Current will be a problem if the processor is run on a different thread, but I've been assured that this doesn't happen in the SDK today.

Src/PerformanceCollector/Perf.Shared/QuickPulseTelemetryProcessor.cs
Mikhail-Pranovich: Makes sense. It appears that Source field is not used by QP after all, also removed its' evaluation here.

Src/PerformanceCollector/PerformanceCollector/Perf.csproj
Mikhail-Pranovich: That would be a design change - not in this iteration

MS-TimothyMothra:
Per offline conversation, I will work on these requested changes Friday.

Src/PerformanceCollector/Perf.Shared/QuickPulseTelemetryProcessor.cs
Mikhail-Pranovich: Good point, fixed

Src/Web/Web.Shared.Net/Extensibility/Implementation/PostSamplingTelemetryProcessor.cs
Mikhail-Pranovich: It will not overwrite the Url & Source if already populated. We can add the check if Url or Source is null, then get request, but this should not be a concern, and I'd rather let pipeline handle that case. Let me know if you want this check

Src/PerformanceCollector/Perf.Shared/QuickPulseTelemetryProcessor.cs
Mikhail-Pranovich: got it, done

Src/Web/Web.Shared.Net/Extensibility/Implementation/PostSamplingTelemetryProcessor.cs
```diff
@@ -0,0 +1,44 @@
...
+    using Microsoft.ApplicationInsights.Extensibility;
+
+    /// <summary>
+    /// PostSamplingTelemetryProcessor evaluates deferred properties.
+    /// It is intended to be used with <see cref="RequestTrackingTelemetryModule.DisableTrackingProperties"/> after Sampling.
+    /// </summary>
+    /// <remarks>
+    /// This feature is still being evaluated and not recommended for end users.
+    /// </remarks>
+    public class PostSamplingTelemetryProcessor : ITelemetryProcessor
```
lmolkova: make it internal and config will still work

Mikhail-Pranovich: this might work for PostSamplingProcessor, however the flags will still have an issue due to [GetProperties](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs#L301). Do you suggest to change it to include BindingFlags.NonPublic ?

PR #1191, opened 2019-05-15
[2.11.0-beta1] guard against malicious headers in quickpulse

MS-TimothyMothra:
### Part 1: Cijo please review

In a previous SDL we introduced `InjectionGuardConstants` to protect from malicious values being injected into the SDK and causing memory overruns. 

I found a place in QuickPulse not doing this check.

(For reference see, mseng bug 651018)

### Part 2: Alex please review

`QuickPulseServiceClient` was retrieving an array of request headers but only using the First() value.
I didn't want to do a length check on an entire array if only the first value would be used.
I changed the implementation to only return the first element in an array.


Src/Common/InjectionGuardConstants.cs
```diff
@@ -50,5 +50,10 @@ static class InjectionGuardConstants
         /// Max number of key value pairs in the tracestate header.
         /// </summary>
         public const int TraceStateMaxPairs = 32;
+
+        /// <summary>
+        /// Max length of incoming Response Header value allowed.
+        /// </summary>
+        public const int QuickPulseResponseHeaderHeaderMaxLength = 1024;
```
cijothomas: minor: HeaderHeader -> Header

MS-TimothyMothra: wow. good catch!

MS-TimothyMothra:
@cijothomas is it safe to merge this now? :)

PR #1198, opened 2019-05-22
[2.11.0-beta1] experimental deferred request properties

MS-TimothyMothra:
Fix Issue #1173.
Second half of this fix.
This will use experimental config to enable deferred request properties.

### TODO
- Wait for BaseSDK 2.11.0-beta1 (contains Experimental)


Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -267,6 +267,13 @@ public void Initialize(TelemetryConfiguration configuration)
             this.telemetryClient = new TelemetryClient(configuration);
             this.telemetryClient.Context.GetInternalContext().SdkVersion = SdkVersionUtils.GetSdkVersion("web:");
 
+            this.DisableTrackingProperties = configuration.EvaluateExperimentalFeature("deferRequestTrackingProperties");
```
lmolkova: > internal bool DisableTrackingProperties { get; set; } = false;

shouldn't they become private now?

MS-TimothyMothra: I think it had some tests that toggled this on/off.
tests require `internal`

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -267,6 +267,13 @@ public void Initialize(TelemetryConfiguration configuration)
...
+            if (this.DisableTrackingProperties)
+            {
+                configuration.TelemetryProcessorChainBuilder.Use(next => new PostSamplingTelemetryProcessor(next));
```
cijothomas: can you validate that this achieves the intended effect? I believe this should be done on default sink's TP instead on common TP.

Src/PerformanceCollector/PerformanceCollector/Perf.csproj
```diff
@@ -37,7 +37,7 @@
   
   <ItemGroup>
     <!--Common Dependencies-->
-    <PackageReference Include="Microsoft.ApplicationInsights" Version="2.10.0" />
+    <PackageReference Include="Microsoft.ApplicationInsights" Version="2.11.0-dev0-build32757" />
```
MS-TimothyMothra: DO NOT MERGE.
Private build to test Experimental changes from BaseSDK

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
MS-TimothyMothra: Done. wrote a bunch of new unit tests. :)

Src/PerformanceCollector/Perf.Shared/QuickPulseTelemetryProcessor.cs
```diff
@@ -114,6 +115,8 @@ public void Initialize(TelemetryConfiguration configuration)
             this telemetry processor will only collect for whichever instrumentation key is specified by the module in StartCollection call.
             */
 
+            this.EvaluateDisabledTrackingProperties = configuration.EvaluateExperimentalFeature("deferRequestTrackingProperties");
```
cijothomas: a line/comment to describe what this experimentalfeature would be really helpful.

Src/PerformanceCollector/Perf.Shared/QuickPulseTelemetryProcessor.cs
```diff
@@ -194,7 +197,7 @@ private ITelemetryDocument ConvertRequestToTelemetryDocument(RequestTelemetry re
             {
                 try
                 {
-                    // some of the requestTelemetry properties might be deffered by using RequestTrackingTelemetryModule.DisableTrackingProperties.
+                    // some of the requestTelemetry properties might be deferred by using RequestTrackingTelemetryModule.DisableTrackingProperties.
```
cijothomas: is this statement stil true in the context of new Experimental Feature suport?

cijothomas: also why is this appearing inside #if net45 only?

Src/PerformanceCollector/Perf.Shared.Tests/QuickPulse/QuickPulseTelemetryProcessorTests.cs
```diff
@@ -2986,5 +2986,165 @@ public void QuickPulseTelemetryProcessorHandlesCalculatedMetricsInThreadSafeMann
...
+        public void VerifyInitializationWhenDeferredIsTrue()
+        {
+            var config = new TelemetryConfiguration();
+            config.ExperimentalFeatures.Add("deferRequestTrackingProperties");
+
+            var spy = new SimpleTelemetryProcessorSpy();
+            var telemetryProcessor = new QuickPulseTelemetryProcessor(spy);
+            telemetryProcessor.Initialize(config);
+
+            Assert.IsTrue(telemetryProcessor.EvaluateDisabledTrackingProperties);
```
cijothomas: why this test inside net45? The actual code is doing this irrespective of net45 or not.

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -265,6 +266,13 @@ public void Initialize(TelemetryConfiguration configuration)
             this.telemetryClient = new TelemetryClient(configuration);
             this.telemetryClient.Context.GetInternalContext().SdkVersion = SdkVersionUtils.GetSdkVersion("web:");
 
+            this.DisableTrackingProperties = configuration.EvaluateExperimentalFeature("deferRequestTrackingProperties");
```
cijothomas: "deferRequestTrackingProperties" - please move this to a const in shared proj. I believe its duplicated in qp and request tracking ..

Src/Web/Web.Shared.Net.Tests/RequestTrackingTelemetryModuleTest.cs
```diff
@@ -478,6 +480,53 @@ public void TrackIntermediateRequestSetsProperties()
             Assert.True(intermediateRequest.Properties.ContainsKey("AI internal"));
         }
 
+        [TestMethod]
+        public void VerifyBehaviorWhenDeferredIsFalse()
+        {
+            var config = this.CreateDefaultConfig(HttpModuleHelper.GetFakeHttpContext());
+            var context = HttpModuleHelper.GetFakeHttpContext();
+
+            // Create and Validate RequestTrackingTelemetryModule
```
cijothomas: Create, Initialize and Validate.

Src/PerformanceCollector/Perf.Shared/QuickPulseTelemetryProcessor.cs
MS-TimothyMothra: I think this is still valid.
The experimental feature is used to set this property, but this is still the property that decides to defer work.

MS-TimothyMothra: I believe this change was only intended for ASP.Net applications, not CORE.
As you can see I'm not changing that in my PR, that already exists.

Src/PerformanceCollector/Perf.Shared/QuickPulseTelemetryProcessor.cs
MS-TimothyMothra: I will leave this comment on the new string constant.

Src/PerformanceCollector/Perf.Shared.Tests/QuickPulse/QuickPulseTelemetryProcessorTests.cs
MS-TimothyMothra: the code path i'm testing is NET45 only.

MS-TimothyMothra: product code does have NET45 condition.

PR #1199, opened 2019-05-22
Upgrade base version to 2.10 stable

lmolkova:
Also fixes upgradeVersion script: it changes 2.9.0-beta4 to 2.10.0 everywhere including Version attribute on assembly:

In the `<Reference Include="Microsoft.ApplicationInsights, Version=2.10.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">` 
version has to be `Version=2.10.0.0`, not `Version=2.10.0`, while package version in packages.config is 2.10.0.

It was leading to build errors and I'm not sure if msbuild became pickier or something else has changed


Src/DependencyCollector/Shared.Tests/Implementation/SqlClientDiagnosticSourceListenerTests.cs
```diff
@@ -141,7 +140,7 @@ public void TracksCommandExecuted()
             Assert.IsTrue((bool)dependencyTelemetry.Success);
             Assert.IsTrue(dependencyTelemetry.Duration > TimeSpan.Zero);
             Assert.IsTrue(dependencyTelemetry.Duration < TimeSpan.FromMilliseconds(500));
-            Assert.IsTrue(DateTimeOffset.UtcNow >= dependencyTelemetry.Timestamp);
+            Assert.IsTrue(Math.Abs((start - dependencyTelemetry.Timestamp).TotalMilliseconds) <= 16);
```
MS-TimothyMothra: why the magic number "16" ?

lmolkova: This is about datetime.now precision on .net fx
https://stackoverflow.com/questions/2143140/c-sharp-datetime-now-precision

lmolkova: Changes in these tests are caused by base sdk which started to set more precise timestamps in https://github.com/microsoft/ApplicationInsights-dotnet/pull/1126

Issue #1153, opened 2019-02-21
Flaky test TestDiagnosticFW45


MS-TimothyMothra:
https://dev.azure.com/mseng/AppInsights/_build/results?buildId=8508196

Failed TestDiagnosticsFW45
Error Message:
Assert.IsTrue failed. AI actionable event was not received
Stack Trace:
at Functional.DiagnosticsTest.TestDiagnosticsFW45() in E:\A\_work\422\s\Test\Web\FunctionalTests\FunctionalTests\DiagnosticsTest.cs:line 69

Issue #1154, opened 2019-02-21
Flaky test Test4Medium200RequestAsync


MS-TimothyMothra:
Failed Test4Medium200RequestAsync
Error Message:
Assert.Fail failed. Read test output. There are errors found in application trace.
Stack Trace:
at Functional.Helpers.SingleWebHostTestBase.StopWebAppHost(Boolean treatTraceErrorsAsFailures) in E:\A\_work\534\s\Test\Web\FunctionalTests\FunctionalTests\TestBase\SingleWebHostTestBase.cs:line 75
at FunctionalTests.Mvc4MediumTrustTest.TestCleanup() in E:\A\_work\534\s\Test\Web\FunctionalTests\FunctionalTests\MVC4_MediumTrustTest.cs:line 46

Issue #1127, opened 2018-12-04
Flaky tests Test452OnNet47_* no status monitor but SDK version is rddp


lmolkova:
Somehow status monitor was enabled in the suite that does not test it therefore SDK version was not right: rddp:2.9.0-7382

2018-12-03T23:55:50.1321438Z Failed   Test452OnNet47_TestSyncHttpDependencyWebApp
2018-12-03T23:55:50.1334752Z Error Message:
2018-12-03T23:55:50.1335693Z  Assert.IsTrue failed. Actual version:rddp:2.9.0-7382
2018-12-03T23:55:50.1335973Z Stack Trace:
2018-12-03T23:55:50.1336397Z    at E2ETests.Test452Base.ValidateDependency(DependencyTelemetry expectedDependencyTelemetry, TelemetryItem`1 actualDependencyTelemetry, String expectedPrefix) in E:\A\_work\39\s\Test\E2ETests\E2ETests\Test452Base.cs:line 1240
2018-12-03T23:55:50.1336546Z    at E2ETests.Test452Base.ValidateBasicDependency(String targetInstanceIp, String targetPath, DependencyTelemetry expectedDependencyTelemetry, String ikey, Int32 count, String expectedPrefix, Int32 additionalSleepTimeMsec) in E:\A\_work\39\s\Test\E2ETests\E2ETests\Test452Base.cs:line 1099
2018-12-03T23:55:50.1336688Z    at E2ETests.Test452Base.TestHttpDependency(String expectedPrefix, String appname, String path, String data, String target, String resultCodeExpected, Boolean successExpected) in E:\A\_work\39\s\Test\E2ETests\E2ETests\Test452Base.cs:line 252
2018-12-03T23:55:50.1336801Z    at E2ETests.Test452Base.TestSyncHttpDependency(String expectedPrefix) in E:\A\_work\39\s\Test\E2ETests\E2ETests\Test452Base.cs:line 258
2018-12-03T23:55:50.1337428Z    at E2ETests.net471.Test452OnNet471.Test452OnNet47_TestSyncHttpDependencyWebApp() in E:\A\_work\39\s\Test\E2ETests\E2ETests\net471\Test452OnNet471.cs:line 75
2018-12-03T23:55:50.1337554Z 

https://dev.azure.com/mseng/AppInsights/_build/results?buildId=7897749

full logs attached
[...]

Issue #1096, opened 2018-10-31
*_HttpDependencyCorrelationInPostRequest flaky test


lmolkova:
https://dev.azure.com/mseng/AppInsights/_build/results?buildId=7634250&view=ms.vss-test-web.test-result-details

lmolkova:
[logs_7898263.zip](https://github.com/Microsoft/ApplicationInsights-dotnet-server/files/2641865/logs_7898263.zip)
+1

 2018-12-04T00:43:58.0088004Z Error Message:
2018-12-04T00:43:58.0088431Z  Assert.AreEqual failed. Expected:<2>. Actual:<1>. 
2018-12-04T00:43:58.0088596Z Stack Trace:
2018-12-04T00:43:58.0088906Z    at E2ETests.Test452Base.<TestHttpDependencyCorrelationInPostRequest>d__19.MoveNext() in E:\A\_work\250\s\Test\E2ETests\E2ETests\Test452Base.cs:line 328
2018-12-04T00:43:58.0089015Z --- End of stack trace from previous location where exception was thrown ---
2018-12-04T00:43:58.0089745Z    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
2018-12-04T00:43:58.0090241Z    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
2018-12-04T00:43:58.0090685Z    at E2ETests.Net462.Test452OnNet462.<Test452OnNet462_HttpDependencyCorrelationInPostRequest>d__14.MoveNext() in E:\A\_work\250\s\Test\E2ETests\E2ETests\net462\Test452OnNet462.cs:line 87
2018-12-04T00:43:58.0091022Z --- End of stack trace from previous location where exception was thrown ---
2018-12-04T00:43:58.0091178Z    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
2018-12-04T00:43:58.0091266Z    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
2018-12-04T00:43:58.0091741Z    at Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.ThreadOperations.ExecuteWithAbortSafety(Action action)
2018-12-04T00:43:58.0091868Z 
[...]

PR #1204, opened 2019-05-22
Use InMemory channel in functional tests

lmolkova:
~~supposedly~~ fixes issues #1153, #1154, #1127, #1096.

func tests use servertelemetrychannel, but it does not guarantee to immediately send telemetry. 
Let's try using InMemory channel and see if it stabilizes tests.

### DIFF
--- a/Test/E2ETests/E2ETests/Test452Base.cs
+++ b/Test/E2ETests/E2ETests/Test452Base.cs
@@ -74,12 +74,13 @@ namespace E2ETests
             GetDockerCompose();
             Trace.WriteLine("Getting docker-compose.exe completed.");
 
-            //DockerUtils.RemoveDockerImage(Apps[AppNameBeingTested].imageName, true);
-            //DockerUtils.RemoveDockerContainer(Apps[AppNameBeingTested].containerName, true);
+            DockerUtils.RemoveDockerImage(Apps[AppNameBeingTested].imageName, true);
+            DockerUtils.RemoveDockerContainer(Apps[AppNameBeingTested].containerName, true);
 
             // Deploy the docker cluster using Docker-Compose
-            //DockerUtils.ExecuteDockerComposeCommand("up -d --force-recreate --build", DockerComposeFileName);
-            DockerUtils.ExecuteDockerComposeCommand("up -d --build", DockerComposeFileName);
+            DockerUtils.ExecuteDockerComposeCommand("up -d --force-recreate --build", DockerComposeFileName);
+            
+            //DockerUtils.ExecuteDockerComposeCommand("up -d --build", DockerComposeFileName);
             Thread.Sleep(5000);
             DockerUtils.PrintDockerProcessStats("Docker-Compose -build");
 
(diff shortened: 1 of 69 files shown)


lmolkova:
All these changes:
- InMemoryChannel in dependency and web func tests
- Forcing docker image recreation 

seem to help a lot reducing the failure rate to ~25%.
still there is some problem that sometimes test app container is not started properly/timely/etc. I'll create another issue for that.

MS-TimothyMothra:
This LGTM, but because @cijothomas is the author, I want to give him a chance to review. :)

Test/Web/FunctionalTests/TestApps/AspNetDiagnostics/ApplicationInsights.config
```diff
@@ -42,9 +42,9 @@
     <Add Type="Microsoft.ApplicationInsights.WindowsServer.UnobservedExceptionTelemetryModule, Microsoft.AI.WindowsServer"/>
     <Add Type="Microsoft.ApplicationInsights.WindowsServer.FirstChanceExceptionStatisticsTelemetryModule, Microsoft.AI.WindowsServer"/>
   </TelemetryModules>
-  <TelemetryChannel Type="Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.ServerTelemetryChannel, Microsoft.AI.ServerTelemetryChannel">
+  <TelemetryChannel Type="Microsoft.ApplicationInsights.Channel.InMemoryChannel, Microsoft.ApplicationInsights">
     <EndpointAddress>http://localhost:4000/v2/track</EndpointAddress>
-    <MaxTelemetryBufferDelay>0:00:01</MaxTelemetryBufferDelay>
+    <DeveloperMode>true</DeveloperMode>
```
cijothomas: if devmode is true, then we dont need flush at all right?

Test/E2ETests/E2ETests/Test452Base.cs
```diff
@@ -74,12 +74,13 @@ public static void MyClassInitializeBase()
...
 
-            //DockerUtils.RemoveDockerImage(Apps[AppNameBeingTested].imageName, true);
-            //DockerUtils.RemoveDockerContainer(Apps[AppNameBeingTested].containerName, true);
+            DockerUtils.RemoveDockerImage(Apps[AppNameBeingTested].imageName, true);
+            DockerUtils.RemoveDockerContainer(Apps[AppNameBeingTested].containerName, true);
 
             // Deploy the docker cluster using Docker-Compose
-            //DockerUtils.ExecuteDockerComposeCommand("up -d --force-recreate --build", DockerComposeFileName);
-            DockerUtils.ExecuteDockerComposeCommand("up -d --build", DockerComposeFileName);
+            DockerUtils.ExecuteDockerComposeCommand("up -d --force-recreate --build", DockerComposeFileName);
```
cijothomas: if we are force-recreating, the the above steps to remove image and container is not required.
but if this is helping anyway - lets just do it.

Test/Web/FunctionalTests/TestApps/AspNetDiagnostics/ApplicationInsights.config
lmolkova: this is Web, there is no flush there. But I guess in dependency tracking functional tests yes, we don't. need it.

In theory, it should with ServerTelemteryChannel isn't it?

cijothomas: Yes devmode=true means we should be flushing instantly after every item automatically.

PR #1211, opened 2019-05-30
Replace non-threadsafe HashSet with ConcurrentDictionary in RequestTrackingTelemetryModule.IsHandlerToFilter

Mikhail-Pranovich:
Fix Issue # .
Fixes IndexOutOfRangeException:
WAC FrontEnd unhandled exception [0] [ExceptionType:System.IndexOutOfRangeException]
System.IndexOutOfRangeException: Index was outside the bounds of the array.
at System.Collections.Generic.HashSet`1.AddIfNotPresent(T value)
at Microsoft.ApplicationInsights.Web.RequestTrackingTelemetryModule.IsHandlerToFilter(IHttpHandler handler)
at Microsoft.ApplicationInsights.Web.RequestTrackingTelemetryModule.NeedProcessRequest(HttpContext httpContext)
...

### DIFF
--- a/Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
+++ b/Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
@@ -28,11 +28,9 @@
         private TelemetryConfiguration telemetryConfiguration;
         private bool initializationErrorReported;
         private ChildRequestTrackingSuppressionModule childRequestTrackingSuppressionModule = null;
-
-        /// <summary>
-        /// Handler types that are not TransferHandlers will be included in request tracking.
-        /// </summary>
-        private HashSet<Type> requestHandlerTypesDoNotFilter = new HashSet<Type>();
+        
+        /// <summary>Tracks if given type should be included in telemetry. ConcurrentDictionary is used as a concurrent hashset.</summary>
+        private ConcurrentDictionary<Type, bool> includedHttpHandlerTypes = new ConcurrentDictionary<Type, bool>();
 
         /// <summary>
         /// Gets or sets a value indicating whether child request suppression is enabled or disabled. 
@@ -371,7 +369,7 @@
             if (handler != null)
             {
                 var handlerType = handler.GetType();
-                if (!this.requestHandlerTypesDoNotFilter.Contains(handlerType))
+                if (!this.includedHttpHandlerTypes.ContainsKey(handlerType))
                 {
                     var handlerName = handlerType.FullName;
                     foreach (var h in this.Handlers)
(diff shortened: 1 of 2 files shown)


MS-TimothyMothra:
In general, this looks good to me.
Is HashSet thread safe? I think that was the main reason for using ConcurrentDictionary here.

cijothomas:
> In general, this looks good to me.
> Is HashSet thread safe? I think that was the main reason for using ConcurrentDictionary here.

https://stackoverflow.com/a/18923091/2624380

MS-TimothyMothra:
> > In general, this looks good to me.
> > Is HashSet thread safe? I think that was the main reason for using ConcurrentDictionary here.
> 
> https://stackoverflow.com/a/18923091/2624380

lol. I was just reading the same article! :)
I haven't seen the exception that @Mikhail-Pranovich referenced, so it would be good to hear what problem he's trying to solve.

Mikhail-Pranovich:
> changelog.md



> And modify changelog.md to add one line about this bug fix.

@cijothomas, Do we need to modify it if the fix is within same version? (i.e. this code was not shipped before)

MS-TimothyMothra:
oooohhhhhhhhhhh. i read this backwards.. it's replacing the HashSet with ConcurrentDictionary. I agree with this change

Src/Web/Web.Shared.Net/RequestTrackingTelemetryModule.cs
```diff
@@ -28,11 +28,9 @@ public class RequestTrackingTelemetryModule : ITelemetryModule
...
         private bool initializationErrorReported;
         private ChildRequestTrackingSuppressionModule childRequestTrackingSuppressionModule = null;
-
-        /// <summary>
-        /// Handler types that are not TransferHandlers will be included in request tracking.
-        /// </summary>
-        private HashSet<Type> requestHandlerTypesDoNotFilter = new HashSet<Type>();
+        
+        /// <summary>Tracks if given type should be included in telemetry. ConcurrentDictionary is used as a concurrent hashset.</summary>
+        private ConcurrentDictionary<Type, bool> includedHttpHandlerTypes = new ConcurrentDictionary<Type, bool>();
```
pharring: readonly?

jirigregor:
Is there any chance to release this fix in some quicker way then usual several months cycle? I consider it as a critial bug causing all of our application crashing randomly and there is no other way how to fix it than restarting entire application (in IIS).

TimothyMothra:
Hi @jirigregor This fix is available now in our 2.11-beta1 release.

Regarding our release cadence, we currently release _something_ monthly. 
Currently, our requirements are that we only ship a patch release if we introduced a regression. NEW bug fixes are picked up in the next version's release cycle.
But I've seen a handful of requests asking for quicker bug fixes. I'll bring this up in our standup today.

Issue #1222, opened 2019-06-20
Support EventCounter Collection


cijothomas:
Starting with .NET Core 3.0 there will a bunch of system-wide counters published using EventCounters. ApplicationInsights SDK need to support collecting of event counters as its replacing PerformanceCounters.

We will support EvenCounter collection NETSTANDARD20 onwards, but nobody is publishing any counters by default, until .net core 3.0. So this will only help collect which user is publishing. Starting with .NET Core 3.0, AspNetCore SDK enables a set of built-in counters.

This module should allow collecting EventCounter from any source/counternames. A pre-defined set of counters will be auto-collected if using Asp.Net Core SDK.

Read more:
Initial List of available counters:
https://github.com/dotnet/diagnostics/blob/master/src/Tools/dotnet-counters/KnownData.cs

EventCounter Tutorial: 
https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.Tracing/documentation/EventCounterTutorial.md

Part of this: https://github.com/microsoft/ApplicationInsights-dotnet-server/issues/1188

cijothomas:
Originally posted: https://github.com/microsoft/ApplicationInsights-aspnetcore/issues/601

alejagapatrick:
https://github.com/microsoft/ApplicationInsights-dotnet-server/issues/1222

PR #1224, opened 2019-06-21
Adds new EventCounterCollectorModule

cijothomas:
Fix Issue #1222 

Adds new Nuget package, new ITelemetryModule for collecting EventCounters.

I am working on adding Unit Tests, please review product code.


Src/EventCounterCollector/EventCounterCollector.Tests/EventCounterCollector.Tests/EventCounterCollectorDiagnoticListener.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Text;
+
+namespace EventCounterCollector.Tests
+{
+    internal class EventCounterCollectorDiagnoticListener : EventListener
+    {
+        public IList<string> EventsReceived { get; private set; }
+        public EventCounterCollectorDiagnoticListener()
+        {
+            this.EventsReceived = new List<string>();
```
lmolkova: make EventsReceived some kind of concurrent collection (e.g. queue). not sure if it is important right away, but it has proven needed/useful in other cases and can save a bunch of time debugging broken tests

Src/EventCounterCollector/EventCounterCollector.Tests/EventCounterCollector.Tests/TestChannel.cs
```diff
@@ -0,0 +1,50 @@
...
+namespace EventCounterCollector.Tests
+{
+    internal class TestChannel : ITelemetryChannel
+    {
+        public bool? DeveloperMode { get; set; }
+        public string EndpointAddress { get; set; }
+
+        public List<ITelemetry> receivedItems;
+
+        public TestChannel(List<ITelemetry> items)
```
lmolkova: same here: make it concurrent collection

Src/EventCounterCollector/EventCounterCollector/EventCounterCollectionModule.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// </summary>
+        internal EventCounterCollectionModule(double refreshIntervalInSecs) : this()
+        {
+            this.refreshInternalInSecs = refreshIntervalInSecs;
+        }
+
+        /// <summary>
+        /// Gets the list of counter names to collect. Each should have the name of EventSource publishing the counter, and counter name.
+        /// </summary>
+        public IList<EventCounterCollectionRequest> Counters { get; private set; }
```
lmolkova: any reason for IList? Could it be IEnumerable?

Src/EventCounterCollector/EventCounterCollector/EventCounterCollectionModule.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// <summary>
+        /// Initialize method is called after all configuration properties have been loaded from the configuration.
+        /// </summary>
+        public void Initialize(TelemetryConfiguration configuration)
+        {
+            try
+            {                
+                if (!this.isInitialized)
+                {
+                    EventCounterCollectorEventSource.Log.ModuleIsBeingInitializedEvent(string.Format(
```
lmolkova: when using EventSource, it is better to format string internally (inside the event). In the majority of cases (there is no listener), event is not actually written and there is no need to allocate extra string. Not very important here, but still best to follow best practices

lmolkova: Also event source formats this message for you and passes args alone

Src/EventCounterCollector/EventCounterCollector/EventCounterListener.cs
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        private void EnableIfRequired(EventSource eventSource)
+        {
+            try
+            {
+                // The EventSourceName is in the list we want to collect some counters from.
+                if (this.countersToCollect.ContainsKey(eventSource.Name))
+                {
+                    Dictionary<string, string> refreshInterval = new Dictionary<string, string>();
```
lmolkova: it seems this could be instance variable - no need to create it per eventSource

Src/EventCounterCollector/EventCounterCollector/EventCounterListener.cs
```diff
@@ -0,0 +1,232 @@
...
+                int actualCount = 0;
+                foreach (KeyValuePair<string, object> payload in eventPayload)
+                {
+                    var key = payload.Key;
+                    if (key.Equals("Name", StringComparison.OrdinalIgnoreCase))
+                    {
+                        var counterName = payload.Value.ToString();
+                        if (this.countersToCollect[eventSourceName].Contains(counterName))
+                        {
+                            metricTelemetry.Name = string.Format(CultureInfo.InvariantCulture, "{0}|{1}", eventSourceName, counterName);
```
lmolkova: it is cheaper to concatenate strings than call string.format. We are on the hot path here

http://geekswithblogs.net/BlackRabbitCoder/archive/2010/05/10/c-string-compares-and-concatenations.aspx

Src/EventCounterCollector/EventCounterCollector/EventCounterCollectionModule.cs
cijothomas: Could be. Followed same approach as existing performance counter modul.

cijothomas: Actually we use count so IList is needed.

Src/EventCounterCollector/EventCounterCollector/EventCounterListener.cs
cijothomas: Not sure what is the recommendation here..https://stackoverflow.com/questions/296978/when-is-it-better-to-use-string-format-vs-string-concatenation
I am unable to find concrete recommendation about one vs other.

Src/EventCounterCollector/EventCounterCollector/EventCounterListener.cs
```diff
@@ -0,0 +1,230 @@
...
+                int actualCount = 0;
+                foreach (KeyValuePair<string, object> payload in eventPayload)
+                {
+                    var key = payload.Key;
+                    if (key.Equals("Name", StringComparison.OrdinalIgnoreCase))
+                    {
+                        var counterName = payload.Value.ToString();
+                        if (this.countersToCollect[eventSourceName].Contains(counterName))
+                        {
+                            metricTelemetry.Name = eventSourceName + "|" + counterName;
```
sywhang: Is this what appears to the users? If so you should use the `DisplayName` property in the payload instead. Our guidance is that `Name` is the property that should be used internally within code, command-line parameters, etc. whereas `DisplayName` property is the one that gets shown on APM/dashboards/etc.

Src/EventCounterCollector/EventCounterCollector.Tests/EventCounterCollector.Tests/EventCounterCollectorDiagnoticListener.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics.Tracing;
+using System.Text;
+
+namespace EventCounterCollector.Tests
+{
+    internal class EventCounterCollectorDiagnoticListener : EventListener
```
sywhang: nit: I think you meant to call this `EventCounterCollectorDiagnosticListener` (missing an s in Diagnostic)

Src/EventCounterCollector/EventCounterCollector/EventCounterListener.cs
cijothomas: Yes! good point. I somehow missed this. Will do a followup PR to address this and actually allow users to specific 'reportAs' as well, so users can decide what the name should appear.

Src/EventCounterCollector/EventCounterCollector/EventCounterListener.cs
```diff
@@ -0,0 +1,230 @@
...
+
+        // Thread-safe variable to hold the list of all EventSourcesCreated.
+        // This class may not be instantiated at the time of EventSource creation, so the list of EventSources should be stored to be enabled after initialization.
+        private ConcurrentQueue<EventSource> allEventSourcesCreated;        
+
+        // EventSourceNames from which counters are to be collected are the keys for this IDictionary.
+        // The value will be the corresponding ICollection of counter names.
+        private IDictionary<string, ICollection<string>> countersToCollect = new Dictionary<string, ICollection<string>>();
+
+        public EventCounterListener(TelemetryClient telemetryClient, IList<EventCounterCollectionRequest> eventCounterCollectionRequests, double refreshIntervalSecs)
```
sywhang: `refreshIntervalSec` shouldn't be a double. Internally we do parse it as float - https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/shared/System/Diagnostics/Tracing/CounterGroup.cs#L57 - but I'd suggest using integer for this since we are planning on removing that undocumented behavior (of parsing it as a float).

sywhang:
Other than the two comments I made above I think things look fine (from a quick glance) - Please feel free to tag me on any follow-up PRs regarding counters/EventListener and I'd be happy to help answer any questions :-)

PR #1232, opened 2019-07-12
Tilee/test fix perf signed

MS-TimothyMothra:
MicroBuild in the conditional ItemGroup has been causing problems. 
This fixes the Perf signing issue and we can remove the manual imports.
Changing other projects for consistency.

### DIFF
--- a/Src/DependencyCollector/DependencyCollector/DependencyCollector.csproj
+++ b/Src/DependencyCollector/DependencyCollector/DependencyCollector.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\..\Product.props" />
 
   <PropertyGroup>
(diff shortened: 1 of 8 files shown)


MS-TimothyMothra:
@cijothomas finally removed those manual imports :)

Src/DependencyCollector/DependencyCollector/DependencyCollector.csproj
```diff
@@ -29,6 +29,10 @@
     <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.9.2">
     <PrivateAssets>All</PrivateAssets>
     </PackageReference>
+  </ItemGroup>
+
+  <ItemGroup>
+    <!--Build Infrastructure-->
```
cijothomas: Release + Windows_NT condition would have always been met right? wonder why this caused issues.

MS-TimothyMothra: Yeah i'm still confused why this worked for every other project except Perf.
This is part of what broke the builds for Dmitry this week so I'm making this change everywhere.

Issue #1212, opened 2019-05-31
Add netstandard 2.0 target on WindowsServer and DependencyCollector


lmolkova:
All .net core libs should target 2.0

lmolkova:
netstandard 2.0 target does not require additional dependencies on

WindowsServer
NETStandard.Library (>= 1.6.1)
System.Runtime.Serialization.Json (>= 4.3.0)
System.Runtime.Serialization.Primitives (>= 4.3.0)

DependecnyCollector:
NETStandard.Library (>= 1.6.1)
System.Data.SqlClient (>= 4.3.1)
System.Diagnostics.StackTrace (>= 4.3.0)

cijothomas:
Thanks for opening this. We can try to do this in next release.

PR #1234, opened 2019-07-15
Adds netstandard2.0 target for dependencycollector.

cijothomas:
UnitTests converted to target 2.1 instead of 1.1. Unit tests were blind about 2.0 before this, and now they will be blind about 1.1 instead. .NET Core 1.1 is already officially out of support, but SDK still has target for 1.1

We may chose to have 1.1 and 2.1 tests, instead of 2.1 alone.
Fix Issue #1212  .

### DIFF
--- a/Src/DependencyCollector/NetCore.Tests/DependencyTrackingTelemetryModuleTestNetCore.cs
+++ b/Src/DependencyCollector/NetCore.Tests/DependencyTrackingTelemetryModuleTestNetCore.cs
@@ -170,29 +170,28 @@
 
                 parent.Stop();
 
-                this.ValidateTelemetryForDiagnosticSource(this.sentTelemetry.Single(), url, request, true, "200", false, parent);
+                this.ValidateTelemetryForDiagnosticSource(this.sentTelemetry.Single(), url, request, true, "200", false, true, parent);
 
                 Assert.AreEqual("k=v", request.Headers.GetValues(RequestResponseHeaders.CorrelationContextHeader).Single());
             }
         }
 
         /// <summary>
-        /// Tests dependency collection when request procession causes exception (DNS issue).
-        /// On .netcore1.1 and before, such dependencies are ot collected
-        /// On .netcore2.0 they are collected, but there is no build infra to support it (https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/572)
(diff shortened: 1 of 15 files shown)


Src/DependencyCollector/NetCore.Tests/DependencyTrackingTelemetryModuleTestNetCore.cs
```diff
@@ -170,29 +170,28 @@ public async Task TestDependencyCollectionWithParentActivity()
...
-        [Timeout(10000)]
+        
         public async Task TestDependencyCollectionDnsIssue()
         {
             using (var module = new DependencyTrackingTelemetryModule())
             {
                 module.Initialize(this.config);
 
-                var request = new HttpRequestMessage(HttpMethod.Get, $"http://{Guid.NewGuid()}");
+                var url = $"http://{Guid.NewGuid()}:5050";
```
MS-TimothyMothra: what is port 5050 ?
would this test still work if we used a non-routable IP address? (ex: 10.0.0.0)

MS-TimothyMothra:
I understand the reason for switching from 1.1 to 2.1, but this makes me just a little bit nervous. 
I would feel better if we added 2.1 now, and removed 1.1 at a later date.

Other than that, this looks good to me. :)

Src/DependencyCollector/NetCore.Tests/DependencyTrackingTelemetryModuleTestNetCore.cs
cijothomas: 5050 was just random!
The goal of this test is to simulate a clientside exception - the easiest way to trigger that is to fail in DNS itself. Its not easy to simulate any other network situation in unit tests.

cijothomas:
@MS-TimothyMothra we were running without 2.0 tests until now :(  Since 1.1 is deprecated officially, we can live without testing 1.1. Anyway if all insist for 1.1 tests, I am happy to get it added.
(I am planning to soon add 3.0 tests, so want to get rid of 1.1 sooner...)

cijothomas:
Plan is to anyway cut support for 1.1 by the time 3.0 releases. So it is okay to shift unit tests to target 2.1 instead of 1.1.

Issue #1088, opened 2018-10-24
SEHException from QuickPulseTelemetryModule.StateThreadWorker


dennis-yemelyanov:
When using Application Insights in .NET console application, sometimes the process crashes with below exception. Although the exception itself comes from .NET code, it looks like this code is called from QuickPulseTelemetryModule. Any ideas what might be causing this error?

> Description: The process was terminated due to an unhandled exception.
> Exception Info: System.Runtime.InteropServices.SEHException
>    at System.Threading.Monitor.ObjWait(Boolean, Int32, System.Object)
>    at System.Threading.Monitor.Wait(System.Object, Int32, Boolean)
>    at System.Threading.Monitor.Wait(System.Object, Int32)
>    at System.Threading.ManualResetEventSlim.Wait(Int32, System.Threading.CancellationToken)
>    at System.Threading.Tasks.Task.SpinThenBlockingWait(Int32, System.Threading.CancellationToken)
>    at System.Threading.Tasks.Task.InternalWait(Int32, System.Threading.CancellationToken)
>    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task)
>    at Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.QuickPulse.QuickPulseTelemetryModule.StateThreadWorker(System.Threading.CancellationToken)
>    at Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.Implementation.QuickPulse.Helpers.QuickPulseThreadModuleScheduler+State.Worker(System.Object)
>    at System.Threading.ThreadHelper.ThreadStart_Context(System.Object)
>    at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
>    at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
>    at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
>    at System.Threading.ThreadHelper.ThreadStart(System.Object)

## Version Info 
SDK Version : 2.7.2
.NET Version : 4.6.2	 						
How Application was onboarded with SDK(VisualStudio/StatusMonitor/Azure Extension) : Visual Studio
OS : Windows 10
[...]

PR #1238, opened 2019-07-19
QuickPulse threads modified to have safeguards for never throwing 

cijothomas:
1. QuickPulse threads modified to have safeguards for never throwing exception. 
2. NetCore implementation moved to share same Thread based implementation as NetFramework. Previously NetCore used Task based implementation. (I believe at that time Thread package was not referred in netcore projects)

Fix Issue #1088  .


Src/PerformanceCollector/Perf.Shared/Implementation/QuickPulse/Helpers/QuickPulseThreadModuleScheduler.cs
```diff
@@ -47,7 +48,23 @@ public void Dispose()
...
             {
-                (state as Action<CancellationToken>)?.Invoke(this.cancellationTokenSource.Token);
+                try
+                {
+                    (state as Action<CancellationToken>)?.Invoke(this.cancellationTokenSource.Token);
+                }
+                catch(Exception ex)
+                {
+                    // This is a Thread, and we don't want any exception thrown ever from this part as this would cause application crash.
+                    try
```
lmolkova: from what  i found, there is another place like this https://github.com/microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/PerformanceCollector/Perf.Shared/QuickPulseTelemetryModule.cs#L310

lmolkova: and here 

https://github.com/microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/PerformanceCollector/Perf.Shared/QuickPulseTelemetryModule.cs#L520

Src/PerformanceCollector/Perf.Shared/Implementation/QuickPulse/Helpers/QuickPulseTaskModuleScheduler.cs
```diff
@@ -1,54 +0,0 @@
-﻿namespace Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.Implementation.QuickPulse.Helpers
```
lmolkova: no issue happened on functions v2 because we used tasks. Now we are switching to threads ad it may start happening.

The reason for threads is we pretend it does guarantee QP to run during thread pool starvation. But you knbow what - QP runs on top of HttpClient and even tough it's a spearate thread, HttpClient  methods are async and run in thread pool threads.

So I guess the whole intent to run QP ina separate thread does not make sense and I'd rather prefer for everything to run on top of tasks.

Src/PerformanceCollector/Perf.Shared/Implementation/QuickPulse/Helpers/QuickPulseThreadModuleScheduler.cs
cijothomas: As discussed offline, all these are run inside same thread worker where we try--catch everything,

Src/PerformanceCollector/Perf.Shared/Implementation/QuickPulse/Helpers/QuickPulseTaskModuleScheduler.cs
cijothomas: I believe the original reason why Threads were used was to ensure QP was able to accumulate data and aggregate in fixed intervals. Sending data to backend using httpclient maybe delayed, but the data would be aggregated in the way expected by QP backend. Similar reasoning was used in MetricAggregator as well to get dedicated thread, to ensure aggregation occur in fixed interval.

@tokaplan Do you have any additional thoughts here? Originaly QP for .net core didn't use Threads because threads nuget was not referenced in the project.

cijothomas: @netgh0st as well to share any additional comments. I think @netgh0st did the .net core support.

tokaplan: Story time.

Once upon a time, around 4am, I was leaning forward in my chair, staring at a glaring screen. After quite a few weeks of designing, prototyping, and redesigning, I had finally finished the very first version of QuickPulse! All telemetry items passing through the SDK pipeline were being registered, accumulated, and all metrics were being aggregated correctly. The only thing remaining was to see how the solution behaved under load, as a complex lock-free multi-threaded solution was likely to produce noticeable bugs under stress.

The load test didn't go well, as data stopped coming in at merely a few hundred requests per second. Further observations showed that the 
```C#
await Task.Delay(TimeSpan.FromSeconds(1)).ConfigureAwait(false)
```

call, which was supposed to return after 1 second, was only returning after 10, even 15 seconds!

I was a sad, sad panda, kind of like this one:
image

I spent many sleepless hours investigating thread pools and thread priorities, finally arriving at the simple yet efficient solution with a dedicated thread.

I had to spend the next couple of weeks trying to convince SDK people that I really did need a dedicated thread, that 
```C#
Thread.Sleep(TimeSpan.FromSeconds(1))
```
did _not_ consume resources while sleeping, and that the mere fact of allocating a _Thread_ object didn't do that either to any significant degree.

Keep the dedicated thread.

Save the whales.

tokaplan: Also, regarding the original .NET Core implementation - it didn't use Threads, but it used a flag which guaranteed that the task would be scheduled on a separate thread. That's why it worked. I believe @netgh0st talked to an expert about it.

tokaplan: Yeah that occurred to me as well at the time, but in practice it was the way I described. Maybe HttpClient uses I/O completion ports/threads which have separate handling, maybe there's a separate thread pool... I tested for regular .NET Framework.

cijothomas: I guess i will keep the current implementation as is, given that lot of research was done to arrive at original decision, and its not in current scope to re-validate earlier design,
I will abandon this and resubmit a PR which will address a single issue - handle all exception from Thread to ensure we dont crash due to unhandled exceptions.

Thanks all for sharing insights :)

cijothomas:
Abandoning this in favor of this: https://github.com/microsoft/ApplicationInsights-dotnet-server/pull/1246

Issue #1253, opened 2019-09-05
QuickPulse does not distinguish multiple processes running on the same host


lmolkova:
QuickPulse sends machine name + role instance to the service and service and UX distinguish services based on machine names rather than role instance, role instance serves as a friendly name only. 

When there is more than one instance on the same host, QP shows a single node for the host regardless of how many roles are reported within the node.

It is not possible to override the machine name QP uses as it is taken from `Environment.MachineName` and not configurable

https://github.com/microsoft/ApplicationInsights-dotnet-server/blob/66a8883df0f15f8e836446c6fafb8cbe6a2d5765/Src/PerformanceCollector/Perf.Shared/QuickPulseTelemetryModule.cs#L358

## Proposal

`QuickPulseTelemetryModule` should be configurable.

PR #1254, opened 2019-09-05
Quick pulse make machine name configurable

lmolkova:
Fix Issue https://github.com/microsoft/ApplicationInsights-dotnet-server/issues/1253 .


Src/PerformanceCollector/Perf.Shared.Tests/QuickPulse/QuickPulseTelemetryModuleTests.cs
```diff
@@ -50,6 +48,25 @@ public void QuickPulseTelemetryModuleIsInitializedBySdk()
...
+            }
+        }
+
+        [TestMethod]
+        public void QuickPulseTelemetryModuleServerIdCanBeChanged()
+        {
+            using (var qp = new QuickPulseTelemetryModule())
+            {
+                qp.ServerId = "my-server-name";
+                Assert.AreEqual("my-server-name", qp.ServerId);
```
tokaplan: As much as I enjoy testing the C# compiler, maybe call Initialize() before asserting just to make it more fun :)

Issue #1215, opened 2019-06-11
Support correlation-context in absence of Request-Id or traceparent


lmolkova:
Correlation-Context is used by partner teams in Microsoft to propagate arbitrary key-value pairs along with request flow and in addition to tracing identifiers (Request-Id, cV or traceparent).

When 1P app instrumented with cV attempts to work with app that is instrumented with ApplicationInsights, AppInsights ignores Correlation-Context. 
It happens because Correlation-Context [according to spec](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/HttpCorrelationProtocol.md) is an extension on top of Request-Id (or traceparent) and is not read when Request-Id is not present.

We should read Correlation-Context regardless of Request-Id (traceprent) presence.

PR #1255, opened 2019-09-09
Parse correlation context in absence of reuqest-id or traceparent

lmolkova:
Fix Issue #1215 .

### DIFF
--- a/Src/Common/WebHeaderCollectionExtensions.cs
+++ b/Src/Common/WebHeaderCollectionExtensions.cs
@@ -128,8 +128,8 @@
 
         public static void ReadActivityBaggage(this NameValueCollection headers, Activity activity)
         {
-            Debug.Assert(headers != null);
-            Debug.Assert(activity != null);
+            Debug.Assert(headers != null, "Headers must not be null");
+            Debug.Assert(activity != null, "Activity must not be null");
 
             int itemsCount = 0;
             var correlationContexts = headers.GetValues(RequestResponseHeaders.CorrelationContextHeader);
@@ -180,7 +180,8 @@
                     }
 
                     currentLength += nextComma + 1;
-                } while (itemsCount < CorrelationContextMaxPairs && currentLength < initialLength);
+                }
+                while (itemsCount < CorrelationContextMaxPairs && currentLength < initialLength);
             }
         }
 
--- a/Src/Common/WebHeaderCollectionExtensions.cs
+++ b/Src/Common/WebHeaderCollectionExtensions.cs
@@ -12,6 +12,8 @@
     internal static class WebHeaderCollectionExtensions
     {
         private const string KeyValuePairSeparator = "=";
+        private const int CorrelationContextHeaderMaxLength = 8192;
+        private const int CorrelationContextMaxPairs = 180;
 
         /// <summary>
         /// For the given header collection, for a given header of name-value type, find the value of a particular key.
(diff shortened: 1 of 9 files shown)


Src/DependencyCollector/Shared.Tests/HeaderCollectionManipulationTests.cs
```diff
@@ -173,40 +177,263 @@ public void GetHeaderValueNoMax()
             Assert.AreEqual("k2=v2", values.Last());
         }
 
-        [Xunit.Theory]
-        [Xunit.InlineData(12)] // k1=v1,k2=v2,".Length
-        [Xunit.InlineData(11)] // k1=v1,k2=v2".Length
-        [Xunit.InlineData(15)] // k1=v1,k2=v2,k3=".Length
-        [Xunit.InlineData(13)] // k1=v1,k2=v2,k".Length
```
lmolkova: apparently these tests are not always run on (at least) my machine as they are xunit while the rest is mstest

Src/Common/WebHeaderCollectionExtensions.cs
```diff
@@ -124,6 +126,67 @@ public static IEnumerable<string> GetHeaderValue(this NameValueCollection header
...
+                    ReadOnlySpan<char> kvp = nextSegment.Slice(0, nextComma).Trim();
+
+                    var separatorInd = kvp.IndexOf('=');
+                    if (separatorInd > 0 && separatorInd < kvp.Length - 1)
+                    {
+                        var separatorIndNext = kvp.Slice(separatorInd + 1).IndexOf('=');
+                        // check there is just one '=' in key-value-pair
+                        if (separatorIndNext < 0)
+                        {
+                            activity.AddBaggage(kvp.Slice(0, separatorInd).Trim().ToString(),
```
TimothyMothra: nit: this line is difficult to read because of all the nested parenthesis. Consider changing this to use named parameters.

Src/Common/WebHeaderCollectionExtensions.cs
```diff
@@ -124,6 +126,67 @@ public static IEnumerable<string> GetHeaderValue(this NameValueCollection header
             return null;
         }
 
+        public static void ReadActivityBaggage(this NameValueCollection headers, Activity activity)
+        {
+            Debug.Assert(headers != null, "Headers must not be null");
+            Debug.Assert(activity != null, "Activity must not be null");
+
```
cijothomas: Do we need to check !activity.Baggage.Any()
https://github.com/microsoft/ApplicationInsights-aspnetcore/blob/develop/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs#L621

Src/Common/WebHeaderCollectionExtensions.cs
```diff
@@ -124,6 +126,67 @@ public static IEnumerable<string> GetHeaderValue(this NameValueCollection header
             return null;
         }
 
+        public static void ReadActivityBaggage(this NameValueCollection headers, Activity activity)
```
TimothyMothra: nit: i would add a comment to this method to make it clear how it's parsing headers. adding an example header is helpfull to visualize the internal operations.

cijothomas:
To double confirm, web sdk just populates correlation-context and populates Activity baggage from them. Decorating application insights telemetry with these is done by OperationCorrelationInitializer from base sdk.
https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/src/Microsoft.ApplicationInsights/Extensibility/OperationCorrelationTelemetryInitializer.cs#L68

Src/Common/WebHeaderCollectionExtensions.cs
lmolkova: sure we can in debug.assert, but we call this method from the places we control and there we make sure there is no baggage

PR #1285, opened 2019-10-04
Adding a flag to enable/disable collection of SQL Command text

stebet:
Adding a boolean flag to the DependencyTrackingTelemetryModule which instructs the SQL processors to enable/disable collecting SQL command text.

Also added tests to verify the functionality.

Perhaps there is a better place for that flag than on the module itself since it's propagated down to the other classes, but it made sense to put it there.

This relates to https://github.com/microsoft/ApplicationInsights-aspnetcore/issues/980 as well as
#1282 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Version 2.12.0-beta1
  - [Enhancement to how QuickPulseTelemetryModule shares its ServiceEndpoint with QuickPulseTelemetryProcessor.](https://github.com/microsoft/ApplicationInsights-dotnet-server/pull/1266)
- - [Adding a flag to DependencyTrackingTelemetryModule to enable/disable collection of SQL Command text](https://github.com/microsoft/ApplicationInsights-dotnet-server/pull/1285)
+ - [Adding a flag to DependencyTrackingTelemetryModule to enable/disable collection of SQL Command text.](https://github.com/microsoft/ApplicationInsights-dotnet-server/pull/1285)
+   Collecting SQL Command Text will now be opt-in, so this value will default to false. This is a change from the current behavior on .NET Core. To see how to collect SQL Command Text see here for details: https://docs.microsoft.com/en-us/azure/azure-monitor/app/asp-net-dependencies#advanced-sql-tracking-to-get-full-sql-query
 
 ## Version 2.11.0
  - Updated Base SDK to 2.11.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Version 2.12.0-beta1
  - [Enhancement to how QuickPulseTelemetryModule shares its ServiceEndpoint with QuickPulseTelemetryProcessor.](https://github.com/microsoft/ApplicationInsights-dotnet-server/pull/1266)
+ - [Adding a flag to DependencyTrackingTelemetryModule to enable/disable collection of SQL Command text](https://github.com/microsoft/ApplicationInsights-dotnet-server/pull/1285)
 
 ## Version 2.11.0
  - Updated Base SDK to 2.11.0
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Version 2.12.0-beta1
  - [Enhancement to how QuickPulseTelemetryModule shares its ServiceEndpoint with QuickPulseTelemetryProcessor.](https://github.com/microsoft/ApplicationInsights-dotnet-server/pull/1266)
+ - [Adding a flag to DependencyTrackingTelemetryModule to enable/disable collection of SQL Command text](https://github.com/microsoft/ApplicationInsights-dotnet-server/pull/1285)
```
cijothomas: we need to have this flag in Framework listenever as well (assuming Microsoft.Data. package approves your PR to add sql text)

https://github.com/microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared/Implementation/FrameworkSqlEventListener.cs

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
```diff
@@ -68,6 +68,11 @@ public class DependencyTrackingTelemetryModule : ITelemetryModule, IDisposable
         /// </summary>
         public bool EnableRequestIdHeaderInjectionInW3CMode { get; set; } = true;
 
+        /// <summary>
+        /// Gets or sets a value indicating whether to track the SQL command text in SQL dependencies.
+        /// </summary>
+        public bool EnableSqlCommandTextInstrumentation { get; set; } = true;
```
cijothomas: Need better messaging and possibly link to doc. Else customers might that this flag alone will give full sql query.
https://docs.microsoft.com/en-us/azure/azure-monitor/app/asp-net-dependencies#advanced-sql-tracking-to-get-full-sql-query

cijothomas: I think we should default to False. In .NET Core we are true by default (with no option to turn it off). This is a bug and sql query collection should be always opt-in, to address privacy concerns
cc: @Dmitry-Matveev

CHANGELOG.md
stebet: Will do. The Microsoft.Data.SqlClient repo has merged the PR as of two days ago :)

Src/DependencyCollector/Shared/DependencyTrackingTelemetryModule.cs
stebet: Yeah I'll update the description and try to make it more clear. I'll also default it to false, but people that are maybe counting on it today would need to be made aware that this default behavios is changing then. Perhaps a post in the applicationinsights-announcements repo would be good to explain the reasoning.

cijothomas: Agree its a breaking change. But I could think of the original behavior as a "bug" and we are now fixing it.
Yes I can surely do a post in announcements repo to make .net core customers aware of the need to opt-in. Will do this in changelog.md for aspnetcore repo as well.

cijothomas:
@Dmitry-Matveev @SergeyKanzhelev I am fine with this PR changes. I could use one of your review as well especially on the part where we are changing SQL Text to be opt-in for .net core customers now.

lmolkova:
LGTM.

regarding defaults, I suggest
- keep collecting SQL command test for existing scenarios: (for System.Data) by default
- stop collecting command text for Microsoft.Data by default

Dmitry-Matveev:
Looks good. Regarding defaults, can you please check with Dave? It may very well be a privacy requirement to have it opt-in only anyway (reqs recently were updated for several privacy considerations), in this case, yes, we're fixing a bug.

stebet:
Distinguishing between the libraries in the collection logic would make it a bit more complicated as well, instead of just treating both libraries the same. Will look into it though if you want to split the default logic between the old and new library.

TimothyMothra:
Adding a reminder here, we need to inform the support teams about this change.
Although it's a bug, we're changing the default behavior and this may negatively impact customers.

cijothomas:
@stebet nothing needed from you. We are discussing internally about the default setting. Will be back with finding here. (about whether we want sqltext ON by default or OFF by default)

Issue #1287, opened 2019-10-07
Support reading metadata when collecting EventCounter data


Expecho:
Regarding the (new) Microsoft.ApplicationInsights.EventCounterCollector:

When I define a custom `EventCounter` I can add metadata to it:

```csharp
methodDurationCounter = new EventCounter(nameof(methodDurationCounter), this);
methodDurationCounter.AddMetadata("Environment", "Production");
methodDurationCounter.AddMetadata("Customer", "Ibis");
```

This metadata is added to the payload. I would be very happy when this data is included as custom properties of the MetricTelemetry that is send. See the [source](https://github.com/microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/EventCounterCollector/EventCounterCollector/EventCounterListener.cs#L225).

I would be more than willing to provide a PR for this.

cijothomas:
AddMetadata is available only in .NET Core 3.0 right? But it shouldn't affect listener.

Yes we are happy to accept a PR to make this enhancement.

Expecho:
> AddMetadata is available only in .NET Core 3.0 right? But it shouldn't affect listener.
> Yes we are happy to accept a PR to make this enhancement.

Yes, .Net Core 3 only. But we can check for payload availability so we can make it backwards compatible.

I'll probably start working on it in the next few days.

PR #1289, opened 2019-10-11
Add metadata support

Expecho:
Fix Issue #1287.
Add support for reading Metadata from EventCounter payload.

Some thoughts:

- I could write a unit test for it, but then the test project in `EventCounterCollector.Tests` needs to be .Net Core 3.0 instead of the current .Net Core 2.1
- The metadata is stored in a simple string separated with `,`.  One could potentially add metadata with the characters `,` or `:` in either the name or the value messing things up. I find this unlikely but I do want to mention this.
- We could just store the whole string in a telemetryitem property called "Metadata" but it makes it hard to use when consuming this metrics in Azure Monitor for example.

I tried to mimic the coding style of the rest of the file. If you want me to extract it to a separate method let me know.

### DIFF
--- a/Src/EventCounterCollector/EventCounterCollector/EventCounterListener.cs
+++ b/Src/EventCounterCollector/EventCounterCollector/EventCounterListener.cs
@@ -207,7 +207,10 @@
                             foreach (var keyValuePairString in keyValuePairStrings)
                             {
                                 var keyValuePair = keyValuePairString.Split(':');
-                                metricTelemetry.Properties.Add(keyValuePair[0], keyValuePair[1]);
+                                if (!metricTelemetry.Properties.ContainsKey(keyValuePair[0]))
+                                {
+                                    metricTelemetry.Properties.Add(keyValuePair[0], keyValuePair[1]);
+                                }
                             }
                         }
                     }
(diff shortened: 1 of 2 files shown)


Src/EventCounterCollector/EventCounterCollector/EventCounterListener.cs
```diff
@@ -198,6 +198,19 @@ private void ExtractAndPostMetric(string eventSourceName, IDictionary<string, ob
...
+                    else if (key.Equals("Metadata", StringComparison.OrdinalIgnoreCase))
+                    {
+                        var metadata = payload.Value.ToString();
+                        if (!string.IsNullOrEmpty(metadata))
+                        {
+                            var keyValuePairStrings = metadata.Split(',');
+                            foreach (var keyValuePairString in keyValuePairStrings)
+                            {
+                                var keyValuePair = keyValuePairString.Split(':');
+                                metricTelemetry.Properties.Add(keyValuePair[0], keyValuePair[1]);
```
cijothomas: can you check and not add property, if property already have a key with same name?

cijothomas:
Looks good! Can you add a one line description to the changelog.md as well? I will merge this for 2.12.0-beta2.
https://github.com/microsoft/ApplicationInsights-dotnet-server/blob/develop/CHANGELOG.md

cijothomas:
Manually kicked off CI build (its not triggered automatically for PRs from fork.). All good.

Issue #1207, opened 2019-05-24
Use WAS-DEFAULT-HOSTNAME HttpHeader to get slot name


debugthings:
## Issue
Currently we use the [WEBSITE_HOSTNAME](https://github.com/microsoft/ApplicationInsights-dotnet-server/blob/c47719cdc1d204ed99b19bb033ef9dc18118103d/Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs#L18) to determine the App Service slot the code is deployed to.

This environment variable is passed in upon process creation by the App Service sandbox and is not updated. This requires a reboot of the App Service to get the correct value.

## Result
The result of this is telemetry will have an incorrect `cloud_RoleName` until this happens which can skew customer metrics.

## Proposal
The proposed fix is to read the `WAS-DEFAULT-HOSTNAME` http header instead which is sent by the App Service front end (ARR) servers and will always have the correct app service host name.

cijothomas:
issue: https://github.com/microsoft/ApplicationInsights-dotnet-server/issues/648

debugthings:
Other references: #872, #871, #314

lmolkova:
We need to think about worker roles and telemetry reported in background threads.

cijothomas:
1. Yes we need a way to solve this for worker roles. But it can be a separate enhancement than this.
2. If we use the same TI as before, but modify it to periodically update the slot name from Request headers, then it should automatically benefit telemetry reported from background thread as well. (Assuming that this enhancement will be implemented NOT by reading incoming headers in every request, but once every min or so)

aarondandy:
How would this work for something using IHostedService? Would telemetry from it just use the cloud role name from the last HTTP request? I know it won't be a common case but it could be a strange corner case and possibly complicated further with percentage based traffic allocations.

cijothomas:
Update:
We got confirmation from AppService team that was-default-hostname is correct way to get the role name we want, and its guaranteed to be correct.

There is no update yet on how to get this on webjobs or background jobs, so they'll continue to rely on the env variable.

PR #1294, opened 2019-10-16
Add new RoleName initializer for Azure Web App to accurately reflect …

cijothomas:
…slot swaps

Fix Issue #1207  .

### DIFF
--- a/Src/Web/Web.Shared.Net/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
+++ b/Src/Web/Web.Shared.Net/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
@@ -1,0 +1,174 @@
+﻿namespace Microsoft.ApplicationInsights.Web
+{
+    using System;
+    using System.Web;
+    using Microsoft.ApplicationInsights.Channel;
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+    using Microsoft.ApplicationInsights.Web.Implementation;
+
+    /// <summary>
+    /// A telemetry initializer that will gather Azure Web App Role Environment context information to
+    /// populate TelemetryContext.Cloud.RoleName
+    /// This uses the http header "WAS-DEFAULT-HOSTNAME" to update role name, if available.
+    /// Otherwise role name is populated from "WEBSITE_HOSTNAME" environment variable.
+    /// </summary>
+    /// <remarks>
+    /// The RoleName is expected to contain the host name + slot name, but will be same across all instances of
+    /// a single App Service.
+    /// Populating RoleName from HOSTNAME environment variable will cause RoleName to be incorrect when a slot swap occurs in AppService.
+    /// The most accurate way to determine the RoleName is to rely on the header WAS-DEFAULT-HOSTNAME, as its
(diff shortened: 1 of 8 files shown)


Src/Web/Web.Shared.Net/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
```diff
@@ -0,0 +1,173 @@
...
+                            }
+                        }
+                        else
+                        {
+                            roleName = request.Context.Cloud.RoleName;
+                        }
+                    }
+                    else
+                    {
+                        if (!this.TryGetRoleNameFromHeader(context, out roleName))
```
SergeyKanzhelev: can we cache it in statitic variable and refresh only 1 time a minute? It is a very tiny fix with the only drawback of being delayed for 1 minute after swap

Src/Web/Web.Shared.Net/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
```diff
@@ -0,0 +1,173 @@
...
+                        {
+                            roleName = this.roleName;
+                        }
```
SergeyKanzhelev: else... we need to indicate that this is not available so for this request we will not attempt it any longer

Src/Web/Web.Shared.Net/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
cijothomas: I gave it a lot of though and decided against it. (Dmitry also suggested same)
As mentioned in the comment for this method, it is intentionally read every request and not periodically.  FunctionsHost is also taking the same approach of reading it in every request.
Even few requests (sampling multiplier may make this worse) with incorrect RoleName is bad, especially considering the support cost. (the symptom is just missing telemetry, so customers keep raising support tickets - it may not be obvious to all customers that the root cause is tied to the way app service slot swap works, and that app insights takes some time to detect this.)

Src/Web/Web.Shared.Net/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
```diff
@@ -0,0 +1,173 @@
...
+                    // RoleName is already populated.
+                    return;
+                }
+
+                string roleName = string.Empty;
+                var context = this.ResolvePlatformContext();
+
+                if (context != null)
+                {
+                    var request = this.GetRequestFromContext(context);
```
lmolkova: HttpContext is not guaranteed to be available here (only for Request you can rely it's here - other telemtery items reported in async/after async calls will not get proper role name

Src/Web/Web.Shared.Net/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
cijothomas: Added IsAzureWebApp() at the beginning so this should not matter now. We exit quickly if not running in Azure Web App.

Src/Web/Web.Shared.Net/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
cijothomas: modified cached value to be updated so it always carries "last known rolename".

PR #1298, opened 2019-10-21
Update to Synthetic User-Agent Telemetry Initializer to allow the con…

delradie:
…figuration to optionally treat empty user agent strings as synthetic, as needed to handle FrontDoor health probes

Fix Issue #1297 
As no solution to the blank useragent string has been forthcoming from the FrontDoor team I have updated the SyntheticUserAgentTelemetryInitializer to optionally treat empty user agents as synthetic

- [ X] I ran Unit Tests locally.

### DIFF
--- a/Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
+++ b/Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
@@ -15,7 +15,8 @@
         private const string SyntheticSourceNameKey = "Microsoft.ApplicationInsights.RequestTelemetry.SyntheticSource";
         private const string SyntheticSourceName = "Bot";
         private string filters = string.Empty;
-        private string[] filterPatterns;        
+        private Boolean markEmptyUserAgentAsSynthetic = true;
+        private string[] filterPatterns;
 
         /// <summary>
         /// Gets or sets the configured patterns for matching synthetic traffic filters through user agent string.
@@ -41,6 +42,21 @@
         }
 
         /// <summary>
+        /// Gets or sets whether an empty user agent string should be treated as synthetic
+        /// </summary>
+        public Boolean MarkEmptyUserAgentAsSynthetic
+        {
+            get
+            {
+                return markEmptyUserAgentAsSynthetic;
+            }
+            set
+            {
+                markEmptyUserAgentAsSynthetic = value;
+            }
+        }
+
+        /// <summary>
         /// Implements initialization logic.
         /// </summary>
         /// <param name="platformContext">Http context.</param>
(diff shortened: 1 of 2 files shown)


delradie:
@delradie

Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
```diff
@@ -15,7 +15,8 @@ public class SyntheticUserAgentTelemetryInitializer : WebTelemetryInitializerBas
         private const string SyntheticSourceNameKey = "Microsoft.ApplicationInsights.RequestTelemetry.SyntheticSource";
         private const string SyntheticSourceName = "Bot";
         private string filters = string.Empty;
-        private string[] filterPatterns;        
+        private Boolean markEmptyUserAgentAsSynthetic = true;
```
cijothomas: this will be a change is current behavior which is not desired.

cijothomas (changes requested):
Can customers solve this problem with their own telemetry initializer, instead of make it part of the SDK itself ?
The problem being solved is not an SDK issue - the external probe should be sending user-agent is the actual problem.
Working around it inside this SDK doesn't sound right approach.

Src/Web/Web.Shared.Net/SyntheticUserAgentTelemetryInitializer.cs
delradie: > Can customers solve this problem with their own telemetry initializer, instead of make it part of the SDK itself ?
> The problem being solved is not an SDK issue - the external probe should be sending user-agent is the actual problem.
> Working around it inside this SDK doesn't sound right approach.

Thanks for the feedback

This was the approach I did try initially, but a lot of the utility code required to implement this is internal to the project, meaning that an attempt to write this as a standalone initialiser ended up with me cloning a significant portion of the Web shared project. Added to this, it is effectively a special case of the user agent TI which cannot be catered for with the current "contains" comparison.

I actually agree completely about this being a problem with the probe, but no attempts to get the Azure Front-Door team to add a UA to their health probes have been successful as yet. With the situation as it stands, our AI telemetry is being rendered very difficult to work with due to the flood of probes (several per second). I would imagine any Front-Door users who use App Insights are going to experience the same hindrance (there are actually a fair few conversations around on exactly this)

delradie: @cijothomas  I've just pushed an update that changes the default to preserve current behaviour as default, but allows configuration to be used to turn the empty user agent handling on.

cijothomas: I didn't get why this is difficult to implement in own project?  It'd be just few lines of code to author a new initializer right?

From what I see:
The issue is with Azure Front-Door not sending UA. This must be fixed in Front-Door side.
Until the fix is available, users should write own telemetry initializer to change default behavior.

delradie: OK, will abandon this and push a separate nuget up.

If you have any internal route to ask the FD team to do this it would be much appreciated as have already sent messages through every avenue I'm aware of and no response has been forthcoming as yet.

Thanks in advance

cijothomas: I have asked for follow up with Front Door team.

